PR #36, opened 2018-06-07
Zend screen

tarrenj:
Cherry picks in the new logo from `zen/master`.

Creates a new commit to remove the ZCash security message in `zend` output.


tarrenj:
Created at `master` rather then `development`.  Closing...

PR #37, opened 2018-06-07
Zend screen

tarrenj:
Cherry picks in the new logo from zen/master.

Creates a new commit to remove the ZCash security message in zend output.


pierstab (approved):

LukasBures (commented):
snimek obrazovky 2018-06-08 v 0 28 02

Maybe we could change this text too. 

Shikata ga nai. - could mean "cant help you" (or something like that)

tarrenj:
@pierstab Can you confirm that the copyright year doesn't need to be changed through out  the comments please?

tarrenj:
@LukasBures Yeah, we can change that to whatever we want, just need to figure out what that is.

Also, a user has asked for the current TX pool status to be displayed too, thoughts on that?

LukasBures:
https://github.com/ZencashOfficial/zencash/pull/37/files#diff-08d3e61a9518c8b4a92c2459a76baee6R3

Here could be year changed too.

tarrenj:
Not an IP lawyer, but I don't think we're expected to constantly update every single notice in every single file, right?

LukasBures (approved):

pierstab (approved):

PR #39, opened 2018-06-10
Add mempool TX count metric to splash screen (#143)

tarrenj:
```
           Block height | 321051
            Connections | 8 (TLS: 8)
  Network solution rate | 47399489 Sol/s
       Mempool TX count | 28 TX
```

### DIFF
--- a/src/metrics.cpp
+++ b/src/metrics.cpp
@@ -204,7 +204,7 @@ int printStats(bool mining)
 {
     // Number of lines that are always displayed
 // ZEN_MOD_START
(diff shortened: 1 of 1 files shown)


LukasBures:
I would like to add myself like reviewer but I dont have that wheel, so I cant.

But anyway: 
**Tested ACK**

cronicc:
utACK

cronicc (approved):

pierstab (approved):

PR #43, opened 2018-06-11
Update README.md

tarrenj:
Adds dependency to build instructions


cronicc (changes requested):
Update the Centos, Windows and ARM sections as well.

Edit: And change version in readme to 2.0.14-rc1

cronicc (approved):

PR #54, opened 2018-06-12
Reintroduce check for missing config, closes #44

### LINKED ISSUE #44: Zend runs without config file
Previously, if a user would start `zend` without a config file, it would copy the config file and exit. Now `zend` will run fine without a config file:
https://github.com/ZencashOfficial/zen/blob/master/src/bitcoind.cpp#L104

ZCash 1.0.14 still uses the same warning from when we originally forked:
https://github.com/zcash/zcash/blob/v1.0.14/src/bitcoind.cpp#L104

cronicc:


pierstab (approved):

tarrenj:
UtACK, Testing now....

U. TestedACK.  Merging.

PR #59, opened 2018-06-15
Remove ZCash from error message

PeaStew:
appeasr when trying to run another zend


cronicc:
@PeaStew please combine all of the PRs changing Zcash->Zen into a single PR.

PR #113, opened 2018-09-20
Delaypenalty

tarrenj:
Penalty system for delayed block submission (see http://bit.ly/2MuN5n0 )


src/checkpoints.cpp
```diff
@@ -7,7 +7,7 @@
 #include "chainparams.h"
 #include "main.h"
 #include "uint256.h"
-
+#include "util.h"
```
rzabarazesh, line 10: seems like this is unused?

src/zen/delay.cpp
```diff
@@ -0,0 +1,41 @@
+#include "zen/delay.h"
+
+int GetBlockDelay(const CBlockIndex& newBlock, const CBlockIndex& prevBlock, const int activeChainHeight, const bool isStartupSyncing)
```
rzabarazesh, line 3: Might be a good idea to add a link to the whitepaper here.

rzabarazesh (approved):

pierstab (approved):

albertog78 (approved):

cronicc (approved):

PR #114, opened 2018-09-21
fix one op_cbah condition in the solver, added new line char to some …

pierstab:
fix one op_cbah condition in the solver, added new line char to some debug messages


albertog78 (approved):

cronicc (approved):
Tested ACK

Issue #128, opened 2018-09-27
zcash references in configure.ac


tarrenj:
https://github.com/ZencashOfficial/zen/blob/63e1f4ce04c20788750a03d51e49c00c67b956f7/configure.ac#L19-L21

Should be updated to `zen`.

PR #133, opened 2018-10-02
replace zcash references in configure.ac

greerso:
replace zcash references in configure.ac per issue #128


tarrenj:
Thanks for this PR!

So it appears that this issue doesn't actually cause any real issues, and resolving it would only complicate merging from upstream.  There are a lot of warnings when building Horizen, and to be honest our build system could use some love (differences between building for Apple and Linux).  I'm going to reject this PR, and close the original issue, and create a new one to track possible updates to our build system.

PR #138, opened 2018-10-27
add new node

reqlez:


tarrenj:
We currently don't have any community nodes. @cronicc @albertog78 what are your thoughts on adding community Tor nodes? It seems like a good idea to me...

cronicc (approved):

cronicc:
This was included in https://github.com/ZencashOfficial/zen/commit/ce51e7c403fe7cf0c84c7c823b7c9bdbcfed4923

PR #198, opened 2019-10-15
First release of modifications for undo

alsala:


qa/rpc-tests/sc_02.py
```diff
@@ -0,0 +1,333 @@
...
+        print
+
+
+    def mark_logs(self, msg):
+        print msg
+        self.nodes[0].dbg_log(msg)
+        self.nodes[1].dbg_log(msg)
+        self.nodes[2].dbg_log(msg)
+
+    def run_test(self):
```
i-Alex, line 110: Why not to add the general description of the test?
As I understood, this test verifies that scinfo is updated correctly for both of chain increasing and rollbacks/forks and take into consideration only active chain.

src/main.cpp
```diff
@@ -1058,7 +1056,7 @@ bool CheckTransaction(const CTransaction& tx, CValidationState &state,
         }
     }
 
-    if (!scMgr.checkTransaction(tx, state) )
+    if (!Sidechain::ScMgr::checkTxSemanticValidity(tx, state) )
```
i-Alex, line 1059: why static?

src/main.cpp
```diff
@@ -2228,6 +2239,16 @@ bool DisconnectBlock(CBlock& block, CValidationState& state, CBlockIndex* pindex
                 if (!ApplyTxInUndo(undo, view, out))
                     fClean = false;
             }
+
+            if (scView)
+            {
+                LogPrint("sc", "%s():%d - undo sc creation tx if any\n", __func__, __LINE__);
+                if (!scView->UndoScCreation(txundo) )
```
i-Alex, line 2247: As I understood, the order of `scView->DecrementScBalance` above and `scView->UndoScCreation` is important.
Moreover `scView->UndoScCreation(txundo)` doesn't mean that `txundo` contains sc creation (or even multiple sc creation) ccout.
Why not to change `scView->UndoScCreation` to accept `blockundo` and put this operation just after `scView->DecrementScBalance` in line 2193

src/main.cpp
```diff
@@ -6951,13 +6995,31 @@ static int getInitCbhSafeDepth()
     return Params().CbhSafeDepth();
 }
 
+static int getInitScCoinsMaturity()
+{
+    if ( (Params().NetworkIDString() == "regtest") || (Params().NetworkIDString() == "testnet") )
```
i-Alex: Why do we need to allow change coins maturity parameter via rpc in a testnet?
Seems, that regtest is enough.

src/rpc/blockchain.cpp
```diff
@@ -994,7 +995,7 @@ UniValue getscinfo(const UniValue& params, bool fHelp)
         scId.SetHex(inputString);
  
         UniValue sc(UniValue::VOBJ);
-        if (!ScMgr::fillJSON(scId, sc) )
+        if (!AddScInfoToJSON(scId, sc) )
```
i-Alex, line 1015: in line 977 runtime error with help message returned, but no return data structure specified in it.

src/sc/sidechain.h
```diff
@@ -55,49 +57,65 @@ class ScInfo
...
 
-    // add/remove/find obj in sc map and DB
-    bool addSidechain(const uint256& scId, const ScInfo& info);
-    void removeSidechain(const uint256& scId);
-    bool addScBackwardTx(const uint256& scId, const uint256& hash);
-    bool removeScBackwardTx(const uint256& scId, const uint256& hash);
-    bool containsScBackwardTx(const uint256& scId, const uint256& txHash);
-
     bool checkSidechainCreation(const CTransaction& tx, CValidationState& state);
     bool hasSCCreationConflictsInMempool(const CTxMemPool& pool, const CTransaction& tx);
```
i-Alex: It's present 3 different naming style for sidechains related logic:
1) `Sc`, for example `ScInfo`
2) `SC`, for example `hasSCCreationConflictsInMempool`
3) `Sidechain`, for example checkSidechainCreation
why not to use a single naming?

qa/rpc-tests/sc_02.py
alsala: Ok, added test description

src/main.cpp
alsala: This is static since only semantic validity is controlled, in other words no use is made of side chain state (which btw is managed by the singleton instance)

src/main.cpp
alsala: Balance undo info are stored in a map of the blockundo obj, the same info could have been spread in txundo objects but more space would have been necessary.
SC creation undo info are stored in txundo obj instead, and you are right: they must be applied after having decremented the balance. They are processed in scView->UndoScCreation during a loop which is doing an analogius task restoring tx inputs in the legacy code. In this way just one loop over txundos is used.

src/main.cpp
alsala: You are right, it should probably not be allowed, or it is useless anyway.

src/sc/sidechain.h
alsala: Ok, using Sc more uniformly

src/main.cpp
i-Alex: ok. I got it, thanks

src/main.cpp
i-Alex: ok

src/rpc/blockchain.cpp
alsala: Ok, will add it

PR #200, opened 2019-10-29
Specialized thread and queue for mempool transaction verification

andreysobol:


src/main.cpp
```diff
@@ -4867,6 +4873,157 @@ void static ProcessGetData(CNode* pfrom)
...
 
+bool checkVerificationQueue(){
+    return verificationQueue.isEmpty();
+}
+
+void txVerificationThread(){
+    int i = 0;
+    while(!ShutdownRequested())
+    {
+        if (!checkVerificationQueue() && i < 100)
```
alsala: Probably is more clear to test directly "verificationQueue.isEmpty()" and remove this method

alsala: The value 100 is hardcoded, it would be better to have a #define or a static const int defined somewhere. Moreover, could it be useful to have it set via an options in the zend?

src/main.cpp
```diff
@@ -4867,6 +4873,157 @@ void static ProcessGetData(CNode* pfrom)
...
+        }
+        else
+        {
+            i = 0;
+            std::this_thread::sleep_for (std::chrono::seconds(1));
+        }
+    }
+}
+
+void checkOneTx(CNode* pfrom, CTransaction& tx){
```
alsala: tx could be const ref

src/txverificationqueue.cpp
```diff
@@ -0,0 +1,52 @@
...
+//#include "txverificationqueue.h"
+#include "main.h"
+//#include "coin.h"
+
+void CTxVerificationQueueEntry::setTX(const CTransaction& newTx)
+{
+    tx = newTx;
+};
+
+CTransaction& CTxVerificationQueueEntry::getTX()
```
alsala: accessor method could return a const ref and could be const themselves:
const CTransaction& CTxVerificationQueueEntry::getTX() const
NodeId CTxVerificationQueueEntry::getNodeId() const

src/txverificationqueue.h
```diff
@@ -0,0 +1,26 @@
+class CTxVerificationQueueEntry
+{
+private:
+    CTransaction tx;
```
alsala, line 4: some data member is not actually used and can be removed as well as mutator related methods

src/txverificationqueue.h
```diff
@@ -0,0 +1,26 @@
...
+    void setNodeId(NodeId newNodeId);
+    NodeId getNodeId();
+};
+
+std::shared_ptr<CTxVerificationQueueEntry> createCTxVerificationQueueEntry(const CTransaction& tx, NodeId nodeId);
+
+class CTxVerificationQueue
+{
+public:
+    std::vector<std::shared_ptr<CTxVerificationQueueEntry>> vectorTX;;
```
alsala: a deque could be use instead of a vector in order to achieve FIFO behaviour.
Also, as proposed above, use of heap memory and smart pointer could be avoided.

std::deque<CTxVerificationQueueEntry> vectorTX;

src/main.cpp
andreysobol: done https://github.com/andreysobol/zen/commit/0358cb2dc407f760210bb2157cb71ce1d5a2b9d8

andreysobol: done
https://github.com/andreysobol/zen/commit/c92a94ec1b880d3724696b56617a12f002cce7c0

src/txverificationqueue.h
andreysobol: done
https://github.com/andreysobol/zen/commit/d9ed3419195d843844ba6be99eff75ac0ddc4aec

src/main.cpp
andreysobol: done
https://github.com/andreysobol/zen/commit/0493d81b4ddfad1832426ac1dcfe2630ee138fcb

src/txverificationqueue.cpp
andreysobol: done
https://github.com/andreysobol/zen/commit/0493d81b4ddfad1832426ac1dcfe2630ee138fcb

src/txverificationqueue.h
andreysobol: done
https://github.com/andreysobol/zen/commit/db174642ad56ef42a0c0b29aee97c4ee5e4151f9

src/main.cpp
```diff
@@ -5605,7 +5602,7 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
         CTransaction tx;
         vRecv >> tx;
 
-        std::shared_ptr<CTxVerificationQueueEntry> ctvqe = createCTxVerificationQueueEntry(tx, pfrom->GetId());
+        CTxVerificationQueueEntry* ctvqe = createCTxVerificationQueueEntry(tx, pfrom->GetId());
```
alsala: There is no need to use ptrs allocated on the heap in the queue, objects could be used instead, for example:

        CTxVerificationQueueEntry ctxvqe;
        ctxvqe.setTX(tx);
        ctxvqe.setNodeId(pfrom->GetId());

This also implies changing the container type (queue<obj> instead of queue<ptr>) and the various signatures  of methods handling entries.

src/txverificationqueue.cpp
```diff
@@ -22,31 +22,32 @@ NodeId CTxVerificationQueueEntry::getNodeId()
...
+    CTxVerificationQueueEntry* ctxvqe = new CTxVerificationQueueEntry();
 
     ctxvqe->setTX(tx);
     ctxvqe->setNodeId(nodeId);
 
     return ctxvqe;
 };
 
-void CTxVerificationQueue::append(std::shared_ptr<CTxVerificationQueueEntry> ctxvqe)
+void CTxVerificationQueue::append(CTxVerificationQueueEntry* ctxvqe)
```
alsala: If objects instead of ptrs are used in the shared queue, this signature should be changed accordingly:

void CTxVerificationQueue::append(const CTxVerificationQueueEntry& ctxvqe)

src/txverificationqueue.cpp
```diff
@@ -22,31 +22,32 @@ NodeId CTxVerificationQueueEntry::getNodeId()
...
-    std::shared_ptr<CTxVerificationQueueEntry> ctxvqe = vectorTX.back();
-    vectorTX.pop_back();
+    CTxVerificationQueueEntry* ctxvqe = dequeTX.front();
     CNode* node = FindNode(ctxvqe->getNodeId());
     if (node != NULL){
         checkOneTx(node, ctxvqe->getTX());
     }
+    dequeTX.pop_front();
+    delete ctxvqe;
 }
```
alsala: If obj are used instead of heap allocated ptrs, no memory management should be used, they are done with after popping_front

src/txverificationqueue.h
```diff
@@ -2,25 +2,21 @@ class CTxVerificationQueueEntry
...
 
-std::shared_ptr<CTxVerificationQueueEntry> createCTxVerificationQueueEntry(const CTransaction& tx, NodeId nodeId);
+CTxVerificationQueueEntry* createCTxVerificationQueueEntry(const CTransaction& tx, NodeId nodeId);
 
 class CTxVerificationQueue
 {
 public:
-    std::vector<std::shared_ptr<CTxVerificationQueueEntry>> vectorTX;;
-    void append(std::shared_ptr<CTxVerificationQueueEntry> ctxvqe);
+    std::deque<CTxVerificationQueueEntry*> dequeTX;
```
alsala: Use can be made of obj instead of ptr in the shared queue

src/main.cpp
andreysobol: please checkout this commit

https://github.com/andreysobol/zen/commit/13eac27e36ebabd9a275328c5f7df03d858b3976

I remove pointer. Can you please doublecheck scenario of "segmentation fault"?

src/main.cpp
alsala: In order to have the LOCK protecting the whole access to the queue pls consider this modification:

        if (i < txVerificationLimit)
        {
            LOCK(cs_verificationQueue);
            if (!(verificationQueue.isEmpty()) )
            {
                verificationQueue.verifyOne();
                i++;
            }
        }
        else
        {
            i = 0;
            std::this_thread::sleep_for (std::chrono::seconds(1));
        }

src/txverificationqueue.cpp
```diff
@@ -22,30 +22,24 @@ NodeId CTxVerificationQueueEntry::getNodeId()
...
-void CTxVerificationQueue::append(CTxVerificationQueueEntry* ctxvqe)
-{
     dequeTX.push_back(ctxvqe);
 }
 
 void CTxVerificationQueue::verifyOne()
 {
-    CTxVerificationQueueEntry* ctxvqe = dequeTX.front();
-    CNode* node = FindNode(ctxvqe->getNodeId());
+    CTxVerificationQueueEntry ctxvqe = dequeTX.front();
```
alsala: consider using a reference in order to save an obj copy:

const CTxVerificationQueueEntry& ctxvqe  = dequeTX.front();

And also a const reference, for that getTX() and getNodeId() should be also 'const' methods (i.e. const CTransaction& getTX() const )

src/main.cpp
andreysobol: Done but in little bit different way

https://github.com/andreysobol/zen/commit/2133ee092b4b0dbac7338c0a40b5be2593ddbbe6

```
    int i = 0;
    bool emptyQueue = false;
    while(!ShutdownRequested())
    {
        if ((!emptyQueue) && i < txVerificationLimit)
        {
            LOCK(cs_verificationQueue);
            if(!(verificationQueue.isEmpty())){
                verificationQueue.verifyOne();
                i++;
            }
            else{
                emptyQueue = true;
            }
        }
        else
        {
            i = 0;
            emptyQueue = false;
            std::this_thread::sleep_for (std::chrono::seconds(1));
        }
    }
```

src/txverificationqueue.cpp
andreysobol: done
https://github.com/andreysobol/zen/commit/f67b6c986b6af43aaca0cc21d78e971d6087573a

PR #205, opened 2019-11-14
Rebase code with latest version of sc_development branch; ready for c…

alsala:
…ode review


src/undo.h
```diff
@@ -53,18 +54,39 @@ class CTxInUndo
...
+typedef Sidechain::ScImmatureAmount CScImmatureAmountUndo;
+
 /** Undo information for a CTransaction */
 class CTxUndo
 {
 public:
     // undo information for all txins
     std::vector<CTxInUndo> vprevout;
+    // undo information for cross chain creation
+    std::vector<CScCreationOutUndo> vsc_ccout;
```
albertog78: Remove sc creation in DisconnectBlock as it is already done for removing TX vouts

src/undo.h
```diff
@@ -73,13 +95,15 @@ class CBlockUndo
 {
 public:
     std::vector<CTxUndo> vtxundo; // for all but the coinbase
+    std::map<uint256, std::vector<CScImmatureAmountUndo>> msc_iaundo;
```
albertog78: change std::vector<CScImmatureAmountUndo> to a map of balances indexed by height to save storage space

src/sc/sidechain.h
```diff
@@ -52,12 +57,37 @@ class ScInfo
...
 typedef boost::unordered_map<uint256, ScInfo, ObjectHasher> ScInfoMap;
 
+class ScCoinsViewCache
+{
+    ScInfoMap mUpdate;
+    std::set<uint256> sErase;
+    std::set<uint256> sDirty;
+
+public:
+    bool UpdateScCoins(const CTransaction& tx, const CBlock&, int nHeight, CTxUndo& txundo);
```
albertog78: rename to UpdateScInfo

src/sc/sidechain.h
```diff
@@ -52,12 +57,37 @@ class ScInfo
...
+    bool UndoScCreation(const CTxUndo& undo);
+    bool IncrementScBalance(int nHeight, CBlockUndo& blockundo);
```
albertog78: rename to ApplyMatureBalances

src/sc/sidechain.h
```diff
@@ -52,12 +57,37 @@ class ScInfo
...
+    bool DecrementScBalance(int nHeight, const CBlockUndo& blockundo);
```
albertog78: rename to RestoreImmatureBalances

src/sc/sidechain.h
```diff
@@ -52,12 +57,37 @@ class ScInfo
...
 
+class ScCoinsViewCache
+{
+    ScInfoMap mUpdate;
+    std::set<uint256> sErase;
+    std::set<uint256> sDirty;
+
+public:
+    bool UpdateScCoins(const CTransaction& tx, const CBlock&, int nHeight, CTxUndo& txundo);
+    bool UndoScCreation(const CTxUndo& undo);
```
albertog78: rename to RevertTxOutputs and update logic by processing directly tx (called by DisconnectBlock) instead of using a specific CTxUndo

albertog78 (changes requested):

src/sc/sidechain.h
```diff
@@ -40,6 +41,9 @@ class ScInfo
     // creation data
     ScCreationParameters creationData;
 
+    // immature amounts
+    std::map<int, CAmount> mImmatureAmounts;
```
i-Alex, line 45: It is better to add the description of the meaning of the key and value, like
key is `maturity height`, value is ...

src/sc/sidechain.cpp
```diff
@@ -632,3 +448,338 @@ void ScMgr::dump_info()
...
+        while (it_ia_map != info.mImmatureAmounts.end() )
+        {
+            int maturityHeight = it_ia_map->first;
+            CAmount a = it_ia_map->second;
+
+            LogPrint("sc", "%s():%d - adding immature balance entry into blockundo (h=%d, amount=%s) for scid=%s\n",
+                __func__, __LINE__, maturityHeight, FormatMoney(a), scIdString);
+
+            // store immature balances into the blockundo obj
+            blockundo.msc_iaundo[scId][maturityHeight] += a;
```
i-Alex: why `+= a` and not ` = a`? Seems that for `blockundo` specific sidechain on specific maturity height amount should be set only once, because now we accumulate  `mImmatureAmounts` by maturity height.

src/undo.h
```diff
@@ -73,13 +73,15 @@ class CBlockUndo
 {
 public:
     std::vector<CTxUndo> vtxundo; // for all but the coinbase
+    std::map<uint256, std::map<int, CAmount> > msc_iaundo;
```
i-Alex, line 76: As I understood, `msc_iaundo` is always a snapshot of `ScInfo.mImmatureAmounts`. So it's easy to do rollbacks.
What about storing just changes between 2 states (views), like what to add and what to remove? This will lead to changes in `ApplyMatureBalances` and `RestoreImmatureBalances`, but should occupy (in general) less memory than a full snapshot.

src/sc/sidechain.h
alsala: Ok, will add it

src/sc/sidechain.cpp
alsala: Ok, correct, will apply the modification

src/undo.h
alsala: That would occupy slightly less memory in the 'undo' db but on the other hand would imply storing 2 states (current and previous) both in memory and in 'sc' db.

i-Alex (approved):

PR #207, opened 2019-11-29
Sc undo code unit test

abi87:


src/sc/sidechain.cpp
```diff
@@ -247,68 +247,104 @@ bool ScMgr::hasScCreationConflictsInMempool(const CTxMemPool& pool, const CTrans
...
 void ScMgr::eraseFromDb(const uint256& scId)
 {
+	if (chosenDbCreationPolicy == dbCreationPolicy::mock)
+	{
+		return; //nothing to erase from db
+	}
+
+	if (chosenDbCreationPolicy != dbCreationPolicy::create)
+	{
+		error("%s():%d - error specifying db creation policy", __func__, __LINE__);
```
alsala: similarly to the code used elsewhere:
return error("<msg>");
Also in other lines below.

src/gtest/test_sideChain.cpp
```diff
@@ -0,0 +1,1196 @@
...
+    int scCreationHeight = 71;
+    aTransaction = createSideChainTxWith(newScId, initialAmount);
+    coinViewCache.UpdateScInfo(aTransaction, aBlock, scCreationHeight);
+
+    //let balance mature
+    int maturityHeight = scCreationHeight + Params().ScCoinsMaturity();
+    coinViewCache.ApplyMatureBalances(maturityHeight, aBlockUndo);
+    CAmount scBalance = coinViewCache.getScInfoMap().at(newScId).balance;
+
+    int amountToUndo = 17;
```
alsala: For consistency CAmount should be used instead of int (also other occurrencies below)

src/gtest/test_sideChain.cpp
```diff
@@ -0,0 +1,1196 @@
...
+
+    //checks
+    EXPECT_FALSE(res);
+}
+
+TEST_F(SideChainTestSuite, ScCreationTxCannotBeRevertedIfScIsNotPreviouslyCreated) {
+    uint256 unexistingScId = uint256S("a1b2");
+
+    //create Sc transaction to be reverted
+    CAmount fwdAmount = 999;
```
alsala: Unused fwdAmount

src/gtest/test_sideChain.cpp
```diff
@@ -0,0 +1,1196 @@
...
+    //check
+    EXPECT_TRUE(res);
+    EXPECT_TRUE(txState.IsValid());
+}
+
+TEST_F(SideChainTestSuite, EmptyTxsAreAllowedInNonEmptyMemPool) {
+    aTransaction = createEmptyScTx();
+
+    CAmount txFee;
+    double txPriority;
```
alsala: txPriority is formally an input, it should be initialized

src/gtest/test_sideChain.cpp
```diff
@@ -0,0 +1,1196 @@
...
+    int scCreationHeight = 10;
+    aTransaction = createSideChainTxWith(newScId, initialFwdTxAmount);
+    coinViewCache.UpdateScInfo(aTransaction, aBlock, scCreationHeight);
+
+    //prerequisites
+    ASSERT_TRUE(sideChainManager.sidechainExists(newScId,&coinViewCache))
+        << "Test requires a tx to be ready to be persisted";
+
+    //test
+    bool res = coinViewCache.Flush();
```
alsala: Could it be useful asserting also res? Otherwise it is not used

src/gtest/test_sideChain.cpp
```diff
@@ -0,0 +1,1196 @@
...
+    ASSERT_FALSE(aTransaction.ccIsNull())<<"Test requires null tx";
+    ASSERT_TRUE(txState.IsValid())<<"Test require transition state to be valid a-priori";
+
+    //test
+    bool res = sideChainManager.checkTxSemanticValidity(aTransaction, txState);
+
+    //checks
+    EXPECT_FALSE(res);
+    EXPECT_FALSE(txState.IsValid());
+    EXPECT_TRUE(txState.GetRejectCode() == REJECT_INVALID)
```
alsala: It could be useful also testing the specific error tag returned, but before doing that, code base should be improved with more specific tags.
Same comment applies for other lines below.

src/gtest/test_sideChain.cpp
```diff
@@ -0,0 +1,1196 @@
...
+    //Checks
+    EXPECT_FALSE(bRet) << "Db double initialization should be forbidden";
+}
+
+///////////////////////////////////////////////////////////////////////////////
+////////////////////////// Test Fixture definitions ///////////////////////////
+///////////////////////////////////////////////////////////////////////////////
+void SideChainTestSuite::resetBaseParams() {
+    //force reset of pCurrentBaseParams
+    CBaseChainParams* nakedCurrentBaseParams = &const_cast<CBaseChainParams &>(BaseParams());
```
alsala: What is the need of nakedCurrentBaseParams (and nakedCurrentParams below)?

src/gtest/test_sideChain.cpp
abi87: Agree. Fixed changing type to CAmount.

src/gtest/test_sideChain.cpp
abi87: Agree. Left there following code refactoring. Removed

src/gtest/test_sideChain.cpp
abi87: Agree. Initialized to 0.0 as not really used in test. Following txPriority too

src/gtest/test_sideChain.cpp
abi87: It's a brute force way to get access to a static pointers pCurrentBaseParams/pCurrentParams defined in chainparamsbase.cpp/chainparams.cpp, whose visibility is limited to those files.

My purpose was initially to make sure I reset those parameters to null following my tests. However:
1. I do not know if the state of those parameters is really null. This is because these tests lives in the same binary as others, which may have set it already.
2. The SelectBaseParams()/SelectParams() functions already take care of setting the parameters to the right one at the beginning of the test, regardless the state of parameters before it

I have removed the reset functions and verified they do not affect test correctness

src/gtest/test_sideChain.cpp
abi87: Agree. As discussed off-line, I am at your convenience to extend tests once we introduce specific tags in codebase

src/sc/sidechain.cpp
abi87: agree. change introduced

src/gtest/test_sideChain.cpp
abi87: Agree. Introduced check on res.

alsala (approved):

src/sc/sidechain.h
```diff
@@ -90,19 +101,57 @@ class ScCoinsViewCache
...
 {
-  private:
+public:
+    enum persistencePolicy {
+        STUB = 0, //utility for UTs
+        PERSIST,
+    };
+
+private:
+    class persistenceLayer {
```
alsala: Pls use an uppercase initial in the name of the class. The same applies also for the other ones below.

src/gtest/test_sidechain.cpp
```diff
@@ -0,0 +1,858 @@
...
+    bool res = coinViewCache.Flush();
+
+    //checks
+    EXPECT_TRUE(res);
+    EXPECT_TRUE(sidechainManager.sidechainExists(scId));
+}
+
+TEST_F(SidechainTestSuite, FlushPersistsForwardTransfers) {
+    uint256 scId = uint256S("a1b2");
+    CAmount initialFwdTxAmount = 1;
```
alsala: initialFwdTxAmount var is unused

alsala (approved):

src/gtest/test_sidechain.cpp
abi87: Agree. Removed.

src/sc/sidechain.h
abi87: Agree. Change performed

PR #211, opened 2019-12-11
Modified and reintroduced test_foundersreward

MarcoOl94:
Modified and reintroduced test_foundersreward


src/gtest/test_foundersreward.cpp
```diff
@@ -22,9 +24,9 @@
 // The output can be copied into chainparams.cpp.
 // The temporary wallet file can be renamed as wallet.dat and used for testing with zcashd.
 //
-#if 0
+
 TEST(founders_reward_test, create_testnet_2of3multisig) {
-    ECC_Start();
+    //ECC_Start(); this is called on the main class of gtest
```
abi87: On line 34 bool fFirstRun is created but not initialized; then in on line 36 fFirstRun  is read inside CWallet::LoadWallet.
This is wrong, as we should not read from uninitialized variables. Probably in most most of cases, *nix systems evaluate uninitialized bools to true. @MarcoOl94  please set the right initial value

src/gtest/test_foundersreward.cpp
```diff
@@ -81,21 +83,24 @@ TEST(founders_reward_test, create_testnet_2of3multisig) {
...
+    int maxHeight = Params().GetConsensus().GetLastCommunityRewardBlockHeight();
     printf("maxHeight = %d\n",maxHeight);
     std::set<std::string> addresses;
     for (int i = 1; i <= maxHeight; i++) {
-        addresses.insert(Params().GetFoundersRewardAddressAtHeight(i));
+    	if(Params().GetCommunityFundAddressAtHeight(i, Fork::CommunityFundType::FOUNDATION).size()>0){
+    		auto result=addresses.insert(Params().GetCommunityFundAddressAtHeight(i, Fork::CommunityFundType::FOUNDATION));
+
+    	}
+
```
abi87: Please conform the rest of the file and substitute tab with 4 spaces.
Also please eliminate newline on line 101

src/gtest/test_foundersreward.cpp
```diff
@@ -107,42 +112,53 @@ void checkNumberOfUniqueAddresses(int nUnique) {
 
 
 TEST(founders_reward_test, general) {
+
```
abi87: Please remove empty line

src/gtest/test_foundersreward.cpp
abi87: Please conform the rest of the file and substitute tab with 4 spaces.
Please add space before curly bracket on line 99
Please eliminate newline on line 101

src/gtest/test_foundersreward.cpp
abi87: On line 34 bool fFirstRun is created but not initialized; then in on line 36 fFirstRun is read inside CWallet::LoadWallet.
This is wrong, as we should not read from uninitialized variables. Probably in most most of cases, *nix systems evaluate uninitialized bools to true. @MarcoOl94 please set the right initial value

src/gtest/test_foundersreward.cpp
```diff
@@ -107,42 +112,53 @@ void checkNumberOfUniqueAddresses(int nUnique) {
...
     SelectParams(CBaseChainParams::TESTNET);
```
abi87: This global parameter is already set in previous test and AFAIK, we do not randomize execution order of UT. You should be able to remove this without affecting test correctness

src/chainparams.cpp
```diff
@@ -370,13 +370,15 @@ bool SelectParamsFromCommandLine()
 // Block height must be >0 and <=last CF reward block height (note that after hfCommunityFundHeight hard fork CF reward is permanent)
 // Index variable i ranges from 0 - (vCommunityFundAddress.size()-1)
 std::string CChainParams::GetCommunityFundAddressAtHeight(int nHeight , Fork::CommunityFundType cfType) const {
-
+	assert(nHeight > 0);
+	assert(nHeight<=consensus.GetLastCommunityRewardBlockHeight());
```
abi87: Please change tab with 4 spaces to conform remaining of the file

src/chainparams.cpp
```diff
@@ -370,13 +370,15 @@ bool SelectParamsFromCommandLine()
...
     return ForkManager::getInstance().getCommunityFundAddress(nHeight,consensus.GetLastCommunityRewardBlockHeight(), cfType);
 }
 
 // The community fund address is expected to be a multisig (P2SH) address
 CScript CChainParams::GetCommunityFundScriptAtHeight(int nHeight, Fork::CommunityFundType cfType) const {
     assert(nHeight > 0);
+	assert(nHeight<=consensus.GetLastCommunityRewardBlockHeight());
```
abi87: Please change tab with 4 spaces to conform remaining of the file

src/gtest/test_foundersreward.cpp
```diff
@@ -107,42 +112,53 @@ void checkNumberOfUniqueAddresses(int nUnique) {
...
+    ASSERT_DEATH(params.GetCommunityFundScriptAtHeight(0,Fork::CommunityFundType::FOUNDATION), "nHeight > 0");
+    ASSERT_DEATH(params.GetCommunityFundScriptAtHeight(maxHeight+1,Fork::CommunityFundType::FOUNDATION), "nHeight<=consensus.GetLastCommunityRewardBlockHeight()");
+    ASSERT_DEATH(params.GetCommunityFundAddressAtHeight(0,Fork::CommunityFundType::FOUNDATION), "nHeight > 0");
+    ASSERT_DEATH(params.GetCommunityFundAddressAtHeight(maxHeight+1,Fork::CommunityFundType::FOUNDATION), "nHeight<=consensus.GetLastCommunityRewardBlockHeight()");
 }
 
 
-#define NUM_MAINNET_FOUNDER_ADDRESSES 48
+//#define NUM_MAINNET_FOUNDER_ADDRESSES 48
+#define NUM_MAINNET_FOUNDER_ADDRESSES 7
```
abi87: If I check right, an int variable local to mainnet test below should be enough to make the test work.
If so, please avoid defines. They are basically type-less (it's just text substitution by preprocessor AFAIK) and they are global

src/gtest/test_foundersreward.cpp
```diff
@@ -107,42 +112,53 @@ void checkNumberOfUniqueAddresses(int nUnique) {
...
 
 TEST(founders_reward_test, mainnet) {
     SelectParams(CBaseChainParams::MAIN);
     checkNumberOfUniqueAddresses(NUM_MAINNET_FOUNDER_ADDRESSES);
 }
 
 
-#define NUM_TESTNET_FOUNDER_ADDRESSES 48
+//#define NUM_TESTNET_FOUNDER_ADDRESSES 48
+#define NUM_TESTNET_FOUNDER_ADDRESSES 4
```
abi87: If I check right, an int variable local to testnet test below should be enough to make the test work.
If so, please avoid defines. They are basically type-less (it's just text substitution by preprocessor AFAIK) and they are global

src/gtest/test_foundersreward.cpp
```diff
@@ -175,18 +191,20 @@ TEST(founders_reward_test, slow_start_subsidy) {
     ASSERT_TRUE(totalSubsidy == MAX_MONEY/10.0);
 }
 
-
+//This test has not more sense because GetNumFoundersRewardAddresses(), GetNumFoundersRewardAddresses2(),vCommunityFundAddress and vCommunityFundAddress2 doesn't exist anymore.
+/*
```
abi87: If I understand right, this tests verifies that founders get rewards and currently we have community rather than founders.
Instead of keeping a commented-out test, shouldn't you write a new one, checking where the rewards given to the community go?

src/gtest/test_foundersreward.cpp
abi87: why not just removing the whole line? Do we care knowing that there used to be a ECC_Start call here?

src/gtest/test_foundersreward.cpp
```diff
@@ -81,21 +83,24 @@ TEST(founders_reward_test, create_testnet_2of3multisig) {
 
     pWallet->Flush(true);
 
-    ECC_Stop();
+    //ECC_Stop(); this is called on the main class of gtest
```
abi87: why not just removing the whole line? Do we care knowing that there used to be a ECC_Stop call here?

src/gtest/test_foundersreward.cpp
```diff
@@ -107,42 +112,53 @@ void checkNumberOfUniqueAddresses(int nUnique) {
...
-    EXPECT_EQ(params.GetFoundersRewardAddressAtHeight(53126), "zrBAG3pXCTDq14nivNK9mW8SfwMNcdmMQpb");
-    EXPECT_EQ(params.GetFoundersRewardScriptAtHeight(53127), ParseHex("a914581dd4277287b64d523f5cd70ccd69f9db384d5387"));
-    EXPECT_EQ(params.GetFoundersRewardAddressAtHeight(53127), "zrBAG3pXCTDq14nivNK9mW8SfwMNcdmMQpb");
 
+    /*
+    CBitcoinAddress add ("zrRBQ5heytPMN5nY3ssPf3cG4jocXeD8fm1");
+        CScriptID scriptID = boost::get<CScriptID>(add.Get()); // Get() returns a boost variant
+        CScript script = CScript() << OP_HASH160 << ToByteVector(scriptID) << OP_EQUAL;
+        std::cout<<"Script: "<<script.ToString()<<std::endl;
+        std::cout<<HexStr(script)<<std::endl;*/
```
abi87: I do not understand why this is here. Could you comment on the purpose of this snippet or remove it?

src/gtest/test_foundersreward.cpp
```diff
@@ -107,42 +112,53 @@ void checkNumberOfUniqueAddresses(int nUnique) {
...
+
+
+
+
```
abi87: please remove useless newlines

src/chainparams.cpp
abi87: Please use 4 spaces instead of tab

src/chainparams.cpp
abi87: please use 4 spaces instead of tab

abi87 (changes requested):

abi87 (approved):

src/chainparams.cpp
```diff
@@ -370,13 +370,15 @@ bool SelectParamsFromCommandLine()
 // Block height must be >0 and <=last CF reward block height (note that after hfCommunityFundHeight hard fork CF reward is permanent)
 // Index variable i ranges from 0 - (vCommunityFundAddress.size()-1)
 std::string CChainParams::GetCommunityFundAddressAtHeight(int nHeight , Fork::CommunityFundType cfType) const {
-
+    assert(nHeight > 0);
+    assert(nHeight<=consensus.GetLastCommunityRewardBlockHeight());
```
alsala: This assert should be removed, since height upper limit is a param of getCommunityFundAddress() that is actually used only by fork1 overriding (and an assert is already there)

alsala (approved):

src/chainparams.cpp
```diff
@@ -370,13 +370,14 @@ bool SelectParamsFromCommandLine()
...
 std::string CChainParams::GetCommunityFundAddressAtHeight(int nHeight , Fork::CommunityFundType cfType) const {
-
+    assert(nHeight > 0);
     return ForkManager::getInstance().getCommunityFundAddress(nHeight,consensus.GetLastCommunityRewardBlockHeight(), cfType);
 }
 
 // The community fund address is expected to be a multisig (P2SH) address
 CScript CChainParams::GetCommunityFundScriptAtHeight(int nHeight, Fork::CommunityFundType cfType) const {
     assert(nHeight > 0);
+    assert(nHeight<=consensus.GetLastCommunityRewardBlockHeight());
```
albertog78: remove this assert, the GetLastCommunityRewardBlockHeight method that is unused and maxHeight parameter in getCommunityFundAddress method

src/gtest/test_foundersreward.cpp
```diff
@@ -13,6 +13,8 @@
 #include <vector>
 #include <boost/filesystem.hpp>
 #include "util.h"
+#include <boost/assign/list_of.hpp>
+
 
 // To run tests:
```
albertog78: this test seems already covered by test_checkblock. Verify if test_checkblock check all the relevant situations

albertog78 (changes requested):

src/gtest/test_checkblock.cpp
```diff
@@ -446,6 +446,33 @@ TEST(ContextualCheckBlock, CoinbaseCommunityReward) {
     block.vtx[0] = CTransaction(mtx);;
     EXPECT_TRUE(ContextualCheckBlock(block, state, &indexPrev));
 
+    //Exceed the LastCommunityRewardBlockHeight
+    CChainParams params = Params();
+    int exceedHeight=params.GetConsensus()._deprecatedGetLastCommunityRewardBlockHeight()+1;
```
abi87: You can avoid defining params variable and call Params().GetConsensus()... in line 451

alsala (approved):

albertog78 (approved):

Issue #173, opened 2019-06-18
Insufficient test coverage of RPC command outputs


cronicc:
RPC command output is not sufficiently checked by our current RPC tests, otherwise we'd have detected https://github.com/ZencashOfficial/zen/issues/172 earlier.

A RPC test should be added validating each RPC command against expected outputs.

PR #212, opened 2019-12-11
Introduced vjoinsplit in the decoderawtransaction's output

MarcoOl94:
Introduced vjoinsplit in the decoderawtransaction's output

### DIFF
--- a/src/rpc/rawtransaction.cpp
+++ b/src/rpc/rawtransaction.cpp
@@ -237,6 +237,9 @@ void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry)
     }
     entry.push_back(Pair("vout", vout));
 
+    UniValue vjoinsplit = TxJoinSplitToJSON(tx);
+    entry.push_back(Pair("vjoinsplit", vjoinsplit));
+
     if (!hashBlock.IsNull()) {
         entry.push_back(Pair("blockhash", hashBlock.GetHex()));
         BlockMap::iterator mi = mapBlockIndex.find(hashBlock);


src/rpc/rawtransaction.cpp
```diff
@@ -237,6 +237,9 @@ void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry)
     }
     entry.push_back(Pair("vout", vout));
 
+    UniValue vjoinsplit = TxJoinSplitToJSON(tx);
+    entry.push_back(Pair("vjoinsplit", vjoinsplit));
+
```
abi87: I understand the change and it looks fine to me. However I do not see any extra unit test/qa tests covering the change you made. I saw cronic opened issue 173 to ask for extra tests (it is marked as enhancement).
@MarcoOl94, @alsala  should we add the unit tests or qa tests covering this change?

abi87, line 242: I understand the change and it looks fine to me. However I do not see any extra unit test/qa tests covering the change you made. I saw cronic opened issue 173 to ask for extra tests (it is marked as enhancement).
@MarcoOl94, @alsala should we add the unit tests or qa tests covering this change?

abi87 (changes requested):
Once we agree on whether unit/qa tests should go with the change, I can approve the change

src/rpc/rawtransaction.cpp
cronicc: Issue #173 is about adding unit tests to verify that the output of all RPC calls has the expected format and includes all values that should be there, not just decoderawtransaction.
That clearly is a larger task and deserves it's own PR.

For this PR no extra tests are needed IMO.
LGTM

abi87: Thanks for clarifying how the task splitting was done @cronicc .
Approving the change

abi87 (approved):

alsala (approved):

albertog78 (approved):

PR #218, opened 2020-01-15
Sc python tests

MarcoOl94:
refactored sc_X tests and modified sc_invalidate to recreate the scenario of issue 215 (actually this test fail)

### DIFF
--- a/qa/rpc-tests/sc_create.py
+++ b/qa/rpc-tests/sc_create.py
@@ -6,17 +6,16 @@
 from test_framework.test_framework import BitcoinTestFramework
 from test_framework.authproxy import JSONRPCException
 from test_framework.util import assert_equal, initialize_chain_clean, \
-    start_nodes, sync_blocks, sync_mempools, connect_nodes_bi, p2p_port
+    start_nodes, sync_blocks, sync_mempools, connect_nodes_bi
 import os
 from decimal import Decimal
-import time
 
 NUMB_OF_NODES = 3
 DEBUG_MODE = 0
 SC_COINS_MAT = 2
 
 
-class headers(BitcoinTestFramework):
+class SCCreateTest(BitcoinTestFramework):
     alert_filename = None
 
     def setup_chain(self, split=False):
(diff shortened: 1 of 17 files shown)


qa/rpc-tests/sc_create.py
```diff
@@ -44,29 +43,6 @@ def setup_network(self, split=False):
...
-
-    def join_network(self):
-        # Join the (previously split) network pieces together: 0-1-2
-        assert self.is_network_split
-        connect_nodes_bi(self.nodes, 1, 2)
-        connect_nodes_bi(self.nodes, 2, 1)
-        time.sleep(2)
-        self.is_network_split = False
-
     def dump_sc_info_record(self, info, i):
```
alsala: util routines might be moved in the py framework, since they are redefined similarly also by other tests

qa/rpc-tests/sc_split.py
```diff
@@ -197,60 +178,66 @@ def run_test(self):
...
 
+        # Node 1 try to reuse inputs of FT
+        mark_logs("\nNode 1 try to reuse inputs of FT...",self.nodes,DEBUG_MODE)
+
+        try:
+            self.nodes[1].sendtoaddress("zthXuPst7DVeePf2ZQvodgyMfQCrYf9oVx4", 3.5)
+        except JSONRPCException, e:
+            errorString = e.error['message']
+            mark_logs (errorString,self.nodes,DEBUG_MODE)
+        assert_equal("Insufficient funds" in errorString, True)
```
abi87: With this lines you want to prove that coins used for the forward transfer are not available any more even if the forward transfer has not been confirmed.
Instead of actually trying to spend those UTXOs I would check the balance in node and verify that is is less than 3.5

qa/rpc-tests/sc_split.py
```diff
@@ -106,89 +78,98 @@ def run_test(self):
...
 
         # Check the mempools of every nodes
-        print ("\nChecking mempools...")
+        mark_logs ("\nChecking mempools...",self.nodes,DEBUG_MODE)
         for i in range(0, NUMB_OF_NODES):
             txmem = self.nodes[i].getrawmempool()
             if i == 2:
                 assert_equal(len(txmem), 0)
             else:
                 assert_equal(sorted(txes), sorted(txmem))
```
abi87: I would move this block up, immediately after network split and before any other sync operation, since you state in comment that the purpose of these lines is to check network split went allright.
As an alternative, you may even remove these lines, since test should be able to use network split function with confidence it works, not retesting it.

qa/rpc-tests/sc_invalidate.py
```diff
@@ -67,380 +68,153 @@ def join_network(self):
         time.sleep(2)
         self.is_network_split = False
 
-    def dump_ordered_tips(self, tip_list):
```
abi87, line 70: You probably move disconnect_nodes function to test framework utils, as you did for dump_sc_info_record and dump_ordered_tips. In fact disconnect_nodes I wonder why it was not already there.
If you move disconnect_nodes to framework utils, please make sure to remove the redefinition for all the other tests.

abi87 (changes requested):

abi87:
in ab/issue_215 branch I have implemented the fixes for some of the scenario you create in the tests [namely allowing fwd tx to sc in mempool].
I look forward pulling them into my branch as soon as they pass review process.
In fact we may even merge them to ab/issue_215 and them create a pull request containing tests and fixes, so to have all tests working in development_beta

abi87:
Changes are good for me. As mentioned above, I am pulling the extra tests you wrote into ab/issue_215
In near future we may add further ones highlighting sc behaviour when two sc creation txs race through the network.

abi87 (approved):

PR #220, opened 2020-01-30
Ab/issue 215

abi87:
I have been able to completely merge  the sidechain view/db into the chainstate views/db.
I have updated and expanded UTs in order to prove the merge is done right.
I have solved issues 215 and 216 and proved it with some unit tests.
I am waiting for python tests by @MarcoOl94 to pass review in order to merge them in


src/init.cpp
```diff
@@ -1462,13 +1461,6 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
                     break;
                 }
 
-                // open sidechain db, read data and populate memory objects
```
alsala, line 1465: The sc db size 'nSideChainDBCache' is not used anymore (see above). Moreover the nCoinDBCache should be probably increased accordingly once the sc db contents are completely defined.

src/coins.h
```diff
@@ -321,8 +339,9 @@ struct CNullifiersCacheEntry
     CNullifiersCacheEntry() : entered(false), flags(0) {}
 };
 
-typedef boost::unordered_map<uint256, CCoinsCacheEntry, CCoinsKeyHasher> CCoinsMap;
-typedef boost::unordered_map<uint256, CAnchorsCacheEntry, CCoinsKeyHasher> CAnchorsMap;
+typedef boost::unordered_map<uint256, CCoinsCacheEntry, CCoinsKeyHasher>      CCoinsMap;
+typedef boost::unordered_map<uint256, CSidechainsCacheEntry, ObjectHasher>    CSidechainsMap;
```
alsala: Maybe the same type of hasher can be used for all maps

src/main.cpp
```diff
@@ -1328,13 +1326,13 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
         return false;
     }
 
-    // perform some check related to sidechains state, e.g. creation of an existing scid, fw to
-    // a not existing one and so on
-    if (!scMgr.IsTxApplicableToState(tx) )
+    // If this tx creates a sc, no other tx must be doing the same in the mempool
+    if (!Sidechain::existsInMempool(pool, tx, state) )
```
alsala: This call should be moved in the block below where the pool.cs is locked

src/main.cpp
```diff
@@ -2511,12 +2502,17 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
             return state.DoS(100, error("ConnectBlock(): too many sigops"),
                              REJECT_INVALID, "bad-blk-sigops");
 
+
```
alsala: These two calls could be moved in the block below since both of them return true in case of coin base tx

src/sc/sidechain.cpp
```diff
@@ -30,7 +23,7 @@ std::string ScInfo::ToString() const
     if (mImmatureAmounts.size() )
     {
         str += "immature amounts:\n";
-        BOOST_FOREACH(const auto& entry, mImmatureAmounts)
+        for (auto& entry : mImmatureAmounts)
```
alsala: entry could be made const

src/coins.h
```diff
@@ -390,14 +414,18 @@ class CCoinsViewBacked : public CCoinsView
     bool GetNullifier(const uint256 &nullifier) const;
     bool GetCoins(const uint256 &txid, CCoins &coins) const;
     bool HaveCoins(const uint256 &txid) const;
+    bool HaveScInfo(const uint256& scId) const;
```
alsala: Consider adding c++11 style override specifier to virtual methods

src/sc/sidechain.cpp
abi87: Fixed, thanks

src/main.cpp
abi87: On the other hand the IsCoinBase case is explicitly handled into the functions [I added it for HaveDependencies method], hence we could as well leave them outside the block. This way the IsCoinBase check is done once (while moving into the loop it'll be done twice). 
Also I find it somewhat more readable outside the loop. Don't you too?

src/main.cpp
abi87: My medium term plan is to make Sidechain::existsInMempool a method of pool as the exists() one for ordinary transactions. This is why I chose to move the two methods close together.
I reckon a lock protection is needed; I'd rather fix the concurrencty issue calling LOCK(pool.cs) within Sidechain::existsInMempool. Any objection to it? thanks

src/coins.h
abi87: I changed hasher to CCoinsKeyHasher and verified code correctness is not affected.

src/init.cpp
abi87: I agree, the whole sizing policy for the dbs may be affected by the decision of storing the sidechains into the chainstatedb. I'd move the investigation for the optimal sizing once we consolidate the sidechain informations we are going to persist and make sure the correctness checks out. Thanks

src/coins.h
abi87: Agreed and coded

src/main.cpp
alsala: Agreed

src/main.cpp
alsala: Agreed

alsala (approved):

src/coins.cpp
```diff
@@ -353,20 +382,425 @@ bool CCoinsViewCache::BatchWrite(CCoinsMap &mapCoins,
...
+    LogPrint("sc", "%s():%d - blockHeight=%d, msc_iaundo size=%d\n", __func__, __LINE__, blockHeight,  blockundo.msc_iaundo.size() );
+
+    std::set<uint256> allKnowScIds;
+    queryScIds(allKnowScIds);
+
+    for(auto it_set = allKnowScIds.begin(); it_set != allKnowScIds.end(); ++it_set)
+    {
+        const uint256& scId = *it_set;
+        ScInfo info;
+        assert(GetScInfo(scId, info));
```
abi87, line 669: ApplyMatureBalances is currently called at every block connection in order to mature the balance of sidechains before flushing it to the main chainstateDB cache, pcoinsTip.
So all the sidechains are loaded into the pcoinsTip, regardless whether the current block has any tx referring to any sidechain.
I wonder whether this could make the cache system less effective as the cache could be filled earlier thus increasing the frequency of flush to db (given current cache sizes).
We discussed with @alsala  about possible improvements.
One idea which has come up is to avoid maturing all sidechains at once "on write", i.e. before a flush.
Instead we could mature "on read", i.e. mature a specific sidechain balance on the moment it is requested from the chainstateDb, just before it is passed up to the cache system.
In this way we won't fill the cache with sidechains needing to be matured and still we will make sure that any processing of sidechains will still be on a mature one.

src/coins.h
```diff
@@ -528,7 +572,13 @@ class CCoinsViewCache : public CCoinsViewBacked
 private:
     CCoinsMap::iterator FetchCoins(const uint256 &txid);
     CCoinsMap::const_iterator FetchCoins(const uint256 &txid) const;
+    CSidechainsMap::const_iterator FetchSidechains(const uint256& scId) const;
+    bool hasScCreationOutput(const CTransaction& tx, const uint256& scId);
+    static int getInitScCoinsMaturity();
+    int getScCoinsMaturity();
```
abi87, line 578: These methods probably belong to sidechain.h or transaction.h files.
With @alsala  we decided to postpone these changes since it has proven strangely difficult to move them out, and since it could be smarted to reshuffle code following merging with certificate related code changes

PR #221, opened 2020-02-05
As/sc cert wip rebased

alsala:


src/primitives/certificate.h
```diff
@@ -0,0 +1,240 @@
...
+
+    friend bool operator==(const CScCertificate& a, const CScCertificate& b)
+    {
+        return a.hash == b.hash;
+    }
+
+    friend bool operator!=(const CScCertificate& a, const CScCertificate& b)
+    {
+        return a.hash != b.hash;
+    }
```
abi87, line 41: You can express operator != in function of operator==. Something like
    friend bool operator!=(const CScCertificate& a, const CScCertificate& b)
    {
        return !(a==b);
    }

src/primitives/certificate.h
```diff
@@ -0,0 +1,240 @@
...
+    bool IsNull() const override {
+        return (
+            scId == uint256() &&
+            epochNumber == 0 &&
+            endEpochBlockHash == uint256() &&
+            totalAmount == 0 &&
+            vout.empty() &&
+            vbt_ccout.empty() &&
+            nonce == uint256() );
+    }
```
abi87, line 73: You can use isNull() methods for uint256, instead of comparing with default constructed temporary object

src/main.cpp
```diff
@@ -2454,8 +2671,17 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
     BOOST_FOREACH(const CTransaction& tx, block.vtx) {
         const CCoins* coins = view.AccessCoins(tx.GetHash());
+#else
+    std::vector<const CTransactionBase*> vTxBase;
+    block.GetTxAndCertsVector(vTxBase);
+
+    for (const CTransactionBase* obj: vTxBase)
+    {
+        const CCoins* coins = view.AccessCoins(obj->GetHash());
+#endif
```
abi87: I am not sure you need to call AccessCoins on certificates. If I understand right, they should not spend any coin

src/sc/sidechain.cpp
```diff
@@ -606,36 +947,179 @@ bool ScMgr::erase(const uint256& scId)
...
+        // should not happen
+        LogPrint("sc", "%s():%d - scId[%s] not found\n",
+            __func__, __LINE__, scId.ToString() );
+        return false;
+    }
+
+    int endEpochHeight = info.creationBlockHeight + (epochNumber * info.creationData.withdrawalEpochLength);
+    pblockindex = chainActive[endEpochHeight];
+
+    if (!pblockindex)
```
abi87, line 1046: Here I'd rather use chainActive.count(endEpochHeight) first to check if block at height mentioned in cert exists and only later on I'd pick it up with chainActive[endEpochHeight], similarly to what you have done on line 1018-1026

src/sc/sidechain.cpp
```diff
@@ -606,36 +947,179 @@ bool ScMgr::erase(const uint256& scId)
...
+    }
+
+    const uint256& hash = pblockindex->GetBlockHash();
+
+    bool ret = (hash == endEpochBlockHash);
+    if (!ret)
+    {
+        LogPrint("sc", "%s():%d - bock hash mismatch: endEpochBlockHash[%s] / calculated[%s]\n", 
+            __func__, __LINE__, endEpochBlockHash.ToString(), hash.ToString());
+    }
```
abi87, line 1060: For coherence with the "early return" style you used about, here I'd go for
`if ( hash != endEpochBlockHash ) { LogPrint(...);return false} return true`
avoiding the use of ret

src/sc/sidechain.cpp
```diff
@@ -107,6 +108,192 @@ bool ScCoinsView::checkTxSemanticValidity(const CTransaction& tx, CValidationSta
...
+
+    // a certificate can not be received after a fixed amount of blocks (for the time being it is epoch length / 5) from the end of epoch (TODO)
+    int maxHeight = ScMgr::getCertificateMaxIncomingHeight(cert.scId, cert.epochNumber);
+    if (maxHeight < 0)
+    {
+        LogPrintf("ERROR: certificate %s, can not calculate max recv height\n", certHash.ToString());
+        return state.Invalid(error("can not calculate max recv height for cert"),
+             REJECT_INVALID, "sidechain-certificate-error");
+    }
```
abi87, line 176: I believe you can drop the check` if (maxHeight < 0) `or turn it into assert.
Reason is that isLegatEpoch already drops certificates whose epochNumber is negative. Moreover creationBlockHeight and withdrawalEpochLength stored in scInfo referenced by certificates cannot be non-negative

src/wallet/rpcwallet.cpp
```diff
@@ -3999,6 +4122,265 @@ UniValue sc_sendmany(const UniValue& params, bool fHelp)
...
+        throw JSONRPCError(RPC_INVALID_PARAMETER, string("invalid epoch data"));
+    }
+
+    // a certificate can not be received after a fixed amount of blocks (for the time being it is epoch length / 5) from the end of epoch (TODO)
+    int maxHeight = ScMgr::getCertificateMaxIncomingHeight(scId, epochNumber);
+    if (maxHeight < 0)
+    {
+        LogPrintf("ERROR: Invalid computed height value\n");
+        throw JSONRPCError(RPC_INVALID_PARAMETER, string("invalid cert data"));
+    }
```
abi87, line 4200: I believe you can drop the checkif (maxHeight < 0)or turn it into assert.
Reason is that isLegatEpoch already drops certificates whose epochNumber is negative. Moreover creationBlockHeight and withdrawalEpochLength stored in scInfo referenced by certificates cannot be non-negative

src/sc/sidechain.h
```diff
@@ -102,11 +117,16 @@ class ScCoinsViewCache : public ScCoinsView
...
+    CAmount getSidechainBalance(const uint256& scId) const;
+    CAmount getSidechainBalanceImmature(const uint256& scId) const;
     bool UpdateScInfo(const CTransaction& tx, const CBlock&, int nHeight);
+    bool UpdateScInfo(const CScCertificate& cert, CBlockUndo& bu);
 
     bool RevertTxOutputs(const CTransaction& tx, int nHeight);
+    bool RevertCertOutputs(const CScCertificate& cert, int nHeight);
     bool ApplyMatureBalances(int nHeight, CBlockUndo& blockundo);
     bool RestoreImmatureBalances(int nHeight, const CBlockUndo& blockundo);
+    bool checkCertificateInMemPool(const CTxMemPool& pool, const CScCertificate& cert);
```
abi87, line 129: This is a declaration without the corresponding definition so, as is, can be eliminated.
However I am strongly in favor of disentangling the (more extensive) checks of the mempool invariants vs the (less extensive) checks to be performed upon evaluating if a certificate can make it to mempool or not

src/sc/sidechain.cpp
```diff
@@ -107,6 +108,192 @@ bool ScCoinsView::checkTxSemanticValidity(const CTransaction& tx, CValidationSta
...
+        }
+
+        if (!creationIsInMempool)
+        {
+            // scId creation not found, should never happen
+            LogPrintf("%s():%d - ERROR: cert[%s]: scId[%s] not found in db and no tx in mempool is creating it\n",
+                __func__, __LINE__, certHash.ToString(), scId.ToString() );
+            return state.Invalid(error("certificate with invalid scId considering mempool"),
+                 REJECT_INVALID, "sidechain-certificate");
+        }
```
abi87, line 243: I believe that this whole branch is currently dead.
Reason is that call to `ScMgr::isLegalEpoch` on line 161 returns false and exit from function if sidechain is not found in db. I tried running the tests with the whole branch set to return false and they still pass.
Either `ScMgr::isLegalEpoch` should visit mempool too or this branch can be removed

src/sc/sidechain.cpp
```diff
@@ -107,6 +108,192 @@ bool ScCoinsView::checkTxSemanticValidity(const CTransaction& tx, CValidationSta
...
+    }
+ 
+    if (certAmount > scBalance)
+    {
+        LogPrint("sc", "%s():%d - invalid cert[%s]: scId[%s] has not balance enough: %s\n",
+            __func__, __LINE__, certHash.ToString(), scId.ToString(), FormatMoney(scBalance) );
+        return state.Invalid(error("certificate with no sc balance enough considering mempool"),
+             REJECT_INVALID, "sidechain-certificate");
+    }
```
abi87, line 293: I believe this would code portion refers to our previous understanding of certificates handling.
As far as I understand now a certificate should be allowed to mempool only if its sidechain is already confirmed.
As per the balance check I think the check in `IsCertApplicableToState` is the only relevant one.
I'd remove the whole section

src/txmempool.cpp
```diff
@@ -352,6 +714,29 @@ void CTxMemPool::check(const CCoinsViewCache *pcoins) const
...
+        CValidationState state;
+#if 1
+        // TODO cert: check this 
+        if (!Sidechain::ScCoinsView::IsCertAllowedInMempool(*this, cert, state))
+        {
+            LogPrint("sc", "%s():%d - cert [%s] has conflicts in mempool\n", __func__, __LINE__, cert.GetHash().ToString());
+            assert(false);
+        }
+#endif
+        assert(cert.ContextualCheckInputs(state, mempoolDuplicate, false, chainActive, 0, false, Params().GetConsensus(), NULL));
```
abi87, line 733: `ContextualCheckInputs` is always true for certs currently. I'd drop this assert

src/main.cpp
```diff
@@ -2565,13 +2763,62 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
+            return state.DoS(100, error("ConnectBlock(): fee out of range"),
+                         REJECT_INVALID, "bad-txns-fee-outofrange");
+        }
+        nFees += nFee;
+
+        if (!cert.IsApplicableToState() ) {
+            LogPrint("sc", "%s():%d - ERROR: tx=%s\n", __func__, __LINE__, cert.GetHash().ToString() );
+            return state.DoS(100, error("ConnectBlock(): invalid sc certificate [%s]", cert.GetHash().ToString()),
+                             REJECT_INVALID, "bad-sc-tx");
+        }
```
abi87, line 2802: Currently, upon connect block, we do not check that the certificates in block reference a legal end of epoch block.
Reason is that such checks are done in `IsCertAllowedInMempool` not in `IsApplicableToState`
Move checks in `IsApplicableToState` so to roll them upon block connect

src/wallet/wallet.h
```diff
@@ -294,6 +292,36 @@ class CMerkleTx : public CTransaction
     // memory only
     mutable bool fMerkleVerified;
 
+    void Init()
+    {
+        hashBlock = uint256();
+        nIndex = -1;
+        fMerkleVerified = false;
+    }
```
abi87, line 300: I'd explicitly initialize all of these members in MerkleAbstractBase  ctor and rename `Init` function to `Reset`, should you need to reuse it outside the ctor

src/wallet/wallet.h
```diff
@@ -431,6 +538,45 @@ class CWalletTx : public CMerkleTx
...
 
+    virtual std::shared_ptr<CWalletObjBase> MakeWalletMapObject() const = 0;
+    static std::shared_ptr<CWalletObjBase> MakeWalletObjectBase(const CTransactionBase& obj, const CWallet* pwallet);
+};
+
+/** 
+ * A transaction with a bunch of additional info that only the owner cares about.
+ * It includes any unrecorded transactions needed to link it back to the block chain.
+ */
+class CWalletTx : public CMerkleTx, virtual public CWalletObjBase
```
abi87, line 549: I believe you can drop the virtual here. Both `CMerkleTx `and `CWalletObjBase `inherits from `MerkleAbstractBase`, but they already do it virtually hence the diamond issue is solved already

src/wallet/wallet.h
```diff
@@ -538,7 +674,91 @@ class CWalletTx : public CMerkleTx
             listScSent.push_back(output);
         }
     }
+    std::shared_ptr<CWalletObjBase> MakeWalletMapObject() const override;
+};
+
+class CWalletCert : public CMerkleCert, virtual public CWalletObjBase
```
abi87, line 680: Similarly to `CWalletTx`, you can drop the virtual here

src/txmempool.cpp
```diff
@@ -273,12 +561,35 @@ void CTxMemPool::removeForBlock(const std::vector<CTransaction>& vtx, unsigned i
...
+void CTxMemPool::removeForBlock(const std::vector<CScCertificate>& vcert, unsigned int nBlockHeight, bool fCurrentEstimate)
+{
+    LOCK(cs);
+    std::vector<CCertificateMemPoolEntry> entries;
+    for (const auto& obj : vcert)
+    {
+        uint256 hash = obj.GetHash();
+        if (mapCertificate.count(hash))
+            entries.push_back(mapCertificate[hash]);
+    }
```
abi87, line 575: In RemoveForBlock for transactions, entries is filled and used to update policy estimates. No policies are specified for certificates currently. Handle policies or drop entries

abi87 (changes requested):

src/txmempool.cpp
```diff
@@ -273,12 +561,35 @@ void CTxMemPool::removeForBlock(const std::vector<CTransaction>& vtx, unsigned i
...
+    for (const auto& obj : vcert)
+    {
+        std::list<std::shared_ptr<CTransactionBase>> removed;
+        remove(obj, removed, true);
+        ClearPrioritisation(obj.GetHash());
+    }
```
abi87, line 581: As discussed AFK, there is an unhandled conflict here.
Say a valid certificate for a given epoch is in mempool, possibly with some txs spending it, and a different certificate for the very same epoch arrives in block. The block gets connected then the `removeForBlock` is called. I believe that `removeForBlock` should handle the aforementioned conflicts, as it is done for txs

abi87 (changes requested):

src/sc/sidechain.cpp
```diff
@@ -375,51 +687,78 @@ bool ScCoinsViewCache::RevertTxOutputs(const CTransaction& tx, int nHeight)
...
+        const std::string& scIdString = scId.ToString();
 
-        auto it_ia_map = info.mImmatureAmounts.begin();
+        ScInfo targetScInfo;
+        assert(getScInfo(scId, targetScInfo));
+
+        auto it_ia_map = targetScInfo.mImmatureAmounts.begin();
 
-        while (it_ia_map != info.mImmatureAmounts.end() )
+        while (it_ia_map != targetScInfo.mImmatureAmounts.end() )
```
abi87, line 730: I do not understand why we use the loop at all here.
I believe the whole point of this function is to move immature amount @ height  blockHeight and return error if  blockHeight  is not the first element of the mImmatureAmounts map.
In essence we should never go beyond the first element of the map, so why looping?
Just
``` return false if targetScInfo.mImmatureAmounts.begin()->first == blockHeight
```
and move immature amounts of targetScInfo.mImmatureAmounts.begin()

abi87 (changes requested):

src/sc/sidechain.cpp
alsala: Ok, since the map is ordered by height, it is enough to have the check just on the first element, others will necessarily have an height grater than the one of the first element.

src/sc/sidechain.cpp
```diff
@@ -606,36 +947,179 @@ bool ScMgr::erase(const uint256& scId)
...
+        LogPrint("cert", "%s():%d - scId[%s] not found\n", __func__, __LINE__, scId.ToString() );
+        return false;
+    }
+
+    if (info.lastReceivedCertificateEpoch == epochNumber)
+    {
+        LogPrint("sc", "%s():%d - a cert for scId[%s] and epoch[%d] has already been received\n",
+            __func__, __LINE__, scId.ToString(), epochNumber);
+        return true;
+    }
```
abi87, line 1002: In case certificate is found "in db" its hash is not filled, unlike what happens in mempool.
Fix the incosistency

abi87 (changes requested):

src/txmempool.cpp
```diff
@@ -99,29 +135,70 @@ bool CTxMemPool::addUnchecked(const uint256& hash, const CTxMemPoolEntry &entry,
...
 
+bool CTxMemPool::addUnchecked(const uint256& hash, const CCertificateMemPoolEntry &entry, bool fCurrentEstimate)
+{
+    LOCK(cs);
+    mapCertificate[hash] = entry;
+    // no mapNextTx or mapNullifiers handling is necessary for certs since they have no inputs or joinsplits 
+
+    const CScCertificate& cert = mapCertificate[hash].GetCertificate();
+
+    mapSidechains[cert.scId].CcTransfersSet.insert(hash);
```
abi87, line 174: We discussed about this choice AFK. I'll note down one issue with the idea of keeping both certificates and txs hashes in the same set, which emerged during the merge.
At some point we may want to return the ScInfo for a sidechain including the balance potential changes due to fwds and certs in the mempool. It would be easier to do so if certificates were not in the same set as txs. Consider that, afaik, there could be at most one certificate in mempool for a given sc. Once the merge is done, I'd really like to store this one certificate in a different variable that CcTransfersSet

PR #228, opened 2020-02-10
Sc rpc cmd, new version of command for creating a sidechain 

alsala:


src/wallet/rpcwallet.cpp
```diff
@@ -760,6 +764,171 @@ UniValue sc_create(const UniValue& params, bool fHelp)
...
+        else
+        {
+            nFee = AmountFromValue(val);
+        }
+    }
+    if (nFee < 0)
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, fee must not be negative");
+    if (nFee > nAmount)
+        throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("Fee %s is greater than outputs %s",
+            FormatMoney(nFee), FormatMoney(nAmount)));
```
abi87: To be consistent with readability in lines 876-879
    if (!MoneyRange(nFee))
        throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, fee out of range");
    if (nFee > nAmount)
        throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("Fee %s is greater than outputs %s",
            FormatMoney(nFee), FormatMoney(nAmount)));

src/wallet/rpcwallet.cpp
```diff
@@ -760,6 +764,171 @@ UniValue sc_create(const UniValue& params, bool fHelp)
...
+    // ---------------------------------------------------------
+    CAmount nAmount = 0;
+    if (setKeyArgs.count("amount"))
+    {
+        UniValue av = find_value(inputObject, "amount");
+        nAmount = AmountFromValue( av );
+        if (nAmount <= 0)
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, amount must be positive");
+        if (!MoneyRange(nAmount))
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, amount out of range");
```
abi87: As a matter of readability, here I'd go for
        if (!MoneyRange(nAmount))
            throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, amount out of range");
        if (nAmount == 0)
            throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, amount cannot be zero");

src/wallet/rpcwallet.cpp
```diff
@@ -760,6 +764,171 @@ UniValue sc_create(const UniValue& params, bool fHelp)
...
+    // sanity check, report error if unknown/duplicate key-value pairs
+    for (const string& s : inputObject.getKeys())
+    {
+        if (!validKeyArgs.count(s))
+            throw JSONRPCError(RPC_INVALID_PARAMETER, string("Invalid parameter, unknown key: ") + s);
+
+        if (setKeyArgs.count(s))
+            throw JSONRPCError(RPC_INVALID_PARAMETER, string("Duplicate key in input: ") + s);
+
+        setKeyArgs.insert(s);
```
abi87: You may even merge these two checks into one, like if (!setKeyArgs.insert(s).second) as insert already tells if element you are trying to insert is in the set. 
Not sure it's more readable but you get all in one operation

src/sc/sidechainrpc.h
```diff
@@ -4,8 +4,11 @@
 #include "amount.h"
 
 #include "sc/sidechaintypes.h"
+#include "base58.h"
 
 //------------------------------------------------------------------------------------
+#define SC_RPC_OPERATION_DEFAULT_MINERS_FEE    10000
+#define SC_RPC_OPERATION_DEFAULT_EPOCH_LENGTH    100
```
abi87: As a matter of style, I'd go for static const declarations here instead of defines. Something like
static const CAmount SC_RPC_OPERATION_DEFAULT_MINERS_FEE(10000);
static const int SC_RPC_OPERATION_DEFAULT_EPOCH_LENGTH(100);

src/wallet/rpcwallet.cpp
alsala: ok

src/wallet/rpcwallet.cpp
alsala: ok

src/wallet/rpcwallet.cpp
alsala: ok

src/sc/sidechainrpc.h
alsala: ok

src/sc/sidechainrpc.cpp
```diff
@@ -219,6 +227,298 @@ void AddScInfoToJSON(UniValue& result)
...
+
+    // sort in ascending order, so smaller utxos appear first
+    std::sort(vInputUtxo.begin(), vInputUtxo.end(), [](CmdUTXO i, CmdUTXO j) -> bool {
+        return ( std::get<2>(i) < std::get<2>(j));
+    });
+
+    CAmount totUTXOAmount = 0;
+    for (const auto & t : vInputUtxo) {
+        totUTXOAmount += std::get<2>(t);
+    }
```
abi87: I think you can avoid this loop and perform the check totUTXOAmount < targetAmount immediately after the loop on lines 314 -327.
So you would optimistically start collecting the UTXOs you need stopping when you reach amount + target fees. You can catch the case where you have insufficient found testing for  _totalInputAmount  < targetAmount

src/sc/sidechainrpc.cpp
```diff
@@ -219,6 +227,298 @@ void AddScInfoToJSON(UniValue& result)
...
+    {
+        CReserveKey keyChange(pwalletMain);
+        CPubKey vchPubKey;
+
+        if (!keyChange.GetReservedKey(vchPubKey))
+        {
+            throw JSONRPCError(RPC_WALLET_KEYPOOL_RAN_OUT, "Could not generate a taddr to use as a change address"); // should never fail, as we just unlocked
+        }
+
+        // TODO: maybe we should send change to sender address if any?
```
abi87: I understand this is still under discussion and it tangles with Sphere functioning. I'd vote for sending the change back to sender address

src/wallet/rpcwallet.cpp
```diff
@@ -760,6 +764,171 @@ UniValue sc_create(const UniValue& params, bool fHelp)
...
+    if (setKeyArgs.count("scid"))
+    {
+        string inputString = find_value(inputObject, "scid").get_str();
+        if (inputString.length() == 0 || inputString.find_first_not_of("0123456789abcdefABCDEF", 0) != std::string::npos)
+            throw JSONRPCError(RPC_TYPE_ERROR, "Invalid scid format: not an hex");
+        scId.SetHex(inputString);
+    }
+    else
+    {
+        scId = GetRandHash();
```
abi87, line 829: If I understand right there is no guarantee here that random scId is not taken already.
It should not happen in gazillion times and user can always retry, but maybe it worth making her succeeding on her first attempt? By doing something like
do { scId = GetRandHash()  } while ( !ScMgr::instance().sidechainExists(scId) );

src/sc/sidechainrpc.h
```diff
@@ -77,6 +80,45 @@ void fundCcRecipients(const CTransaction& tx, std::vector<CcRecipientVariant>& v
 void AddScInfoToJSON(UniValue& result);
 bool AddScInfoToJSON(const uint256& scId, UniValue& sc);
 void AddScInfoToJSON(const uint256& scId, const ScInfo& info, UniValue& sc);
+
+// Input UTXO is a tuple (triple) of txid, vout, amount)
+typedef std::tuple<uint256, int, CAmount> CmdUTXO;
```
abi87: I have two objections to this typedef:
1- It is only relevant within scRpcCreationCmd now, so I would move it within scRpcCreationCmd  declaration, even with private access
2- It's not really a command, rather a data type. How about renaming it to something like selectedUTXO?

qa/rpc-tests/sc_create_2.py
```diff
@@ -0,0 +1,291 @@
...
+    def __init__(self, items):
+        # need to have something in the dictionary
+        self['something'] = 'something'
+        self.items = items
+
+    def __getitem__(self, key):
+        return self.last_val
+
+    def __iter__(self):
+        subiter = iter(self.items)
```
abi87: This iterator looks unused and can be removed

qa/rpc-tests/sc_create_2.py
```diff
@@ -0,0 +1,291 @@
...
+
+        mark_logs("Node 1 generates 2 block",self.nodes,DEBUG_MODE)
+        self.nodes[1].generate(2)
+        self.sync_all()
+
+        mark_logs("Node 0 generates 220 block",self.nodes,DEBUG_MODE)
+        self.nodes[0].generate(220)
+        self.sync_all()
+
+        tx = []
```
abi87, line 75: This is not needed as tx is rebound on line 97 without reading empty list first

abi87: Same applies to errorString

qa/rpc-tests/sc_create_2.py
```diff
@@ -0,0 +1,291 @@
...
+            mark_logs(errorString,self.nodes,DEBUG_MODE)
+            assert_true("withdrawal" in errorString)
+
+        # Node 1 create the SC using a valid input and a fromaddress key/value
+        #------------------------------------------------------------------------------------------
+        mark_logs("\nNode 1 create SC with valid input and a fromaddress key/value", self.nodes,DEBUG_MODE)
+
+        sc_id = "95f5de0829fd3c5e45b63f67beab8c4cb8c1c359ef6e2787aaaf5442d6f4779f"
+        wel = 5
+        fromaddr = []
```
abi87, line 202: Useless, thou shalt not do it, C++ developer!

qa/rpc-tests/sc_create_2.py
```diff
@@ -0,0 +1,291 @@
...
+import pprint
+from decimal import Decimal
+import json
+
+NUMB_OF_NODES = 2
+DEBUG_MODE = 0
+SC_COINS_MAT = 2
+
+
+class FakeDict(dict):
```
abi87, line 21: Nice trick. You get a kind of iterable only dictionary, in the sense that you cannot access directly its keys, but in iterating it you can simulate duplicated keys

qa/rpc-tests/sc_create_2.py
```diff
@@ -0,0 +1,291 @@
...
+                                 extra_args=[["-sccoinsmaturity=%d" % SC_COINS_MAT, '-logtimemicros=1', '-debug=sc',
+                                              '-debug=py', '-debug=mempool', '-debug=net',
+                                              '-debug=bench']] * NUMB_OF_NODES)
+
+        if not split:
+            # 1 and 2 are joint only if split==false
+            if NUMB_OF_NODES > 2:
+                connect_nodes_bi(self.nodes, 1, 2)
+            sync_blocks(self.nodes[1:NUMB_OF_NODES])
+            sync_mempools(self.nodes[1:NUMB_OF_NODES])
```
abi87: I think this is irrelevant to the test, clear it out

qa/rpc-tests/sc_create_2.py
```diff
@@ -0,0 +1,291 @@
...
+
+    def setup_chain(self, split=False):
+        print("Initializing test directory " + self.options.tmpdir)
+        initialize_chain_clean(self.options.tmpdir, NUMB_OF_NODES)
+        self.alert_filename = os.path.join(self.options.tmpdir, "alert.txt")
+        with open(self.alert_filename, 'w'):
+            pass  # Just open then close to create zero-length file
+
+    def setup_network(self, split=False):
+        self.nodes = []
```
abi87: useless, remove

src/sc/sidechainrpc.cpp
```diff
@@ -3,10 +3,18 @@
...
 
+#include <wallet/wallet.h>
 
 extern UniValue ValueFromAmount(const CAmount& amount);
 extern CAmount AmountFromValue(const UniValue& value);
+extern CWallet* pwalletMain;
+extern CFeeRate minRelayTxFee;
+extern std::string EncodeHexTx(const CTransaction& tx);
+extern UniValue signrawtransaction(const UniValue& params, bool fHelp);
+extern UniValue sendrawtransaction(const UniValue& params, bool fHelp);
```
abi87: You should be able to change these externs with the following includes
#include <core_io.h>
#include <rpc/server.h>
#include <main.h>
#include <init.h>

src/wallet/rpcwallet.cpp
```diff
@@ -49,6 +49,10 @@ extern UniValue TxJoinSplitToJSON(const CTransaction& tx);
 int64_t nWalletUnlockTime;
```
abi87, line 49: The file contains a mix of tabs and spaces. Please take the time to format it to spaces

abi87 (changes requested):
Request changes though previous comments

src/sc/sidechainrpc.cpp
alsala: Ok, you are right

src/sc/sidechainrpc.cpp
alsala: Yes, this is an open point but I agree, a better default approach is to use the sender address for the change.

src/wallet/rpcwallet.cpp
alsala: A clash is highly unlikely but theoretically might happen. In that case the command will fail anyway because the clash will be handled when the generated tx is checked.

src/sc/sidechainrpc.h
alsala: Ok to both

qa/rpc-tests/sc_create_2.py
alsala: ok

qa/rpc-tests/sc_create_2.py
alsala: :) ok

src/wallet/rpcwallet.cpp
alsala: ok

src/sc/sidechainrpc.cpp
```diff
@@ -8,13 +8,13 @@
 
 #include <wallet/wallet.h>
 
+#include <core_io.h>
+#include <rpc/server.h>
+#include <main.h>
+#include <init.h>
+
 extern UniValue ValueFromAmount(const CAmount& amount);
 extern CAmount AmountFromValue(const UniValue& value);
```
abi87, line 17: Sorry I left out these lines in previous comment. You can remove them out as these declarations are in `rpc/server.h` already

abi87 (changes requested):

PR #236, opened 2020-03-30
Further cleanup by getters

abi87:
Following introduction of getters for vin/vout/vjoinsplit in tx, a whole bunch of transaction/certs methods introduced for polymorphism can be aggregated into a single method processing a TransactionBase. I have done that for some of them, in order to eliminate some code duplication and improve readability.

I have also made some minor changes in Accept*ToMemoryPool implementation so to highlight similarities between transactions and certificates processing

### DIFF
--- a/src/coins.cpp
+++ b/src/coins.cpp
@@ -395,7 +395,7 @@ bool CCoinsViewCache::BatchWrite(CCoinsMap &mapCoins,
                 ); //A fresh entry should not exist in localCache or be already erased
                 cacheSidechains[entryToWrite.first] = entryToWrite.second;
                 break;
-            case CSidechainsCacheEntry::Flags::DIRTY:               //A dirty entry may or may not exist in localCache
+            case CSidechainsCacheEntry::Flags::DIRTY: //A dirty entry may or may not exist in localCache
                     cacheSidechains[entryToWrite.first] = entryToWrite.second;
                 break;
             case CSidechainsCacheEntry::Flags::ERASED:
--- a/src/coins.cpp
+++ b/src/coins.cpp
@@ -1064,8 +1064,8 @@ CAmount CCoinsViewCache::GetValueIn(const CTransactionBase& txBase) const
         return 0;
 
     CAmount nResult = 0;
-    for (const CTxIn& vin : txBase.GetVin())
-        nResult += GetOutputFor(vin).nValue;
+    for (const CTxIn& in : txBase.GetVin())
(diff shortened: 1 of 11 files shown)


src/coins.cpp
```diff
@@ -1058,16 +1058,16 @@ const CTxOut &CCoinsViewCache::GetOutputFor(const CTxIn& input) const
...
+CAmount CCoinsViewCache::GetValueIn(const CTransactionBase& txBase) const
 {
-    if (tx.IsCoinBase())
+    if (txBase.IsCoinBase())
         return 0;
 
     CAmount nResult = 0;
-    for (unsigned int i = 0; i < tx.GetVin().size(); i++)
-        nResult += GetOutputFor(tx.GetVin()[i]).nValue;
+    for (const CTxIn& vin : txBase.GetVin())
```
alsala: 'vin' shout be probably renamed in 'in' or 'txin'

src/primitives/certificate.cpp
```diff
@@ -64,6 +64,16 @@ bool CScCertificate::CheckVersionBasic(CValidationState &state) const
     return true;
 }
 
+bool CScCertificate::CheckVersionIsStandard(std::string& reason, const int nHeight) const {
```
alsala: Here is the right place for adding certificate version checking, that was missing even in original code

src/primitives/certificate.h
```diff
@@ -74,6 +74,7 @@ class CScCertificate : virtual public CTransactionBase
 
     //CHECK FUNCTIONS
     bool CheckVersionBasic        (CValidationState &state) const override;
+    bool CheckVersionIsStandard   (std::string& reason, const int nHeight) const override;
```
alsala: probably here 'const' is useless for int var

alsala (approved):

src/primitives/certificate.cpp
abi87: Agree. I think also CheckVersionBasic may be filled with a gross check about min max version as with transactions. Also a ContextualVersionCheck should be added for certs, which now is missing.
I would like to fix the version checks in one session rather than here, however. Here I'd content myself with opening up the places for checks

src/primitives/certificate.h
abi87: Done

src/coins.cpp
abi87: Done

PR #237, opened 2020-03-30
Sc bwt vout

alsala:
Certificate serialized version specifies backward transfers vector as a separate field from regular vouts vectors. Introduced in CTxOut an "in memory only" field (not to be serialized) indicating if it is a backward transfer to be able to distinguish the origin of each CTxOut in vout (backward transfer/regular output).


qa/rpc-tests/sc_cert_and_fwt_in_mempool.py
```diff
@@ -0,0 +1,122 @@
...
+        block = self.nodes[1].getblock(bl_hash, True)
+        vtx   = block['tx']
+        vcert = block['cert']
+        assert_equal(True, fwt in vtx)
+        assert_equal(True, cert in vcert)
+
+
+
+if __name__ == '__main__':
+    sc_cr_fw_cert().main()
```
abi87: I understand the usefulness of this test in fixing the issue related to the faulty clean-up of certs and txs. However I do not think we should include it as a separate test in the regression suite. The reason is that we have a couple of very specific unit tests in gtest/test_sidechain_to_mempool.cpp which address this issue.
If we want to keep a python level test of the removal done right, I'd suggest to add it to one of the py tests which we already have.

src/primitives/certificate.h
```diff
@@ -117,6 +167,9 @@ class CScCertificate : virtual public CTransactionBase
     unsigned int GetLegacySigOpCount() const override;
 
     bool IsCoinCertified() const override { return true; }
+protected:
+    static void UnserializeTranslate(const BackwardTransferData& btdata, CTxOut& txout) ;
+    static void SerializeTranslate(const CTxOut& txout, BackwardTransferData& btdata) ;
```
abi87: Why not introducing a CTxOut constructor from a BackwardTransferData and, viceversa a BackwardTransferData from CTxOut?
This way we would avoid these functions' awkward names. Awkward since, while they are currently used for serialization/deserialization purposes, their implementations have nothing to do really with these kind of operations. They just build one data type out of the other

src/primitives/transaction.h
```diff
@@ -361,12 +361,16 @@ class CTxOut
     CAmount nValue;
     CScript scriptPubKey;
 
+    /* mem only */
+    bool backwardTransfer;
```
abi87: I'd rename it to something like isFromBackwardTransfer for clarity

src/primitives/transaction.h
```diff
@@ -361,12 +361,16 @@ class CTxOut
...
     CTxOut()
     {
         SetNull();
     }
 
-    CTxOut(const CAmount& nValueIn, CScript scriptPubKeyIn);
+    CTxOut(const CAmount& nValueIn, CScript scriptPubKeyIn, bool backwardTransferIn = false) :
+        nValue(nValueIn), scriptPubKey(scriptPubKeyIn), backwardTransfer(backwardTransferIn) {}
 
     ADD_SERIALIZE_METHODS;
```
abi87, line 380: Placing here a comment related to inline `void SerializationOp`.
Serialization op will create an object out of the stream. What will the bool backwardTransfer be initialized to?
*nix system probably will set it to false, but this is not guaranteed by C++ standand. Maybe I am wrong but I think we should set its value explicitly for read operations as it does not have a default ctor

src/rpc/rawtransaction.cpp
```diff
@@ -187,24 +186,52 @@ void CertToJSON(const CScCertificate& cert, const uint256 hashBlock, UniValue& e
         UniValue o(UniValue::VOBJ);
         ScriptPubKeyToJSON(txout.scriptPubKey, o, true);
         out.push_back(Pair("scriptPubKey", o));
+        if (txout.backwardTransfer)
```
abi87: I am not sure I understand the logic here. Shouldn't we push the scriptPubKey for  non backwardTransfer only and push the phk for backwardTransfer?
In other words don't you have and empty scriptPubKey here for  backwardTransfers? If so, is it acceptable?

src/rpc/rawtransaction.cpp
```diff
@@ -187,24 +186,52 @@ void CertToJSON(const CScCertificate& cert, const uint256 hashBlock, UniValue& e
...
+            if (chainActive.Contains(pindex)) {
+                entry.push_back(Pair("confirmations", 1 + chainActive.Height() - pindex->nHeight));
+                entry.push_back(Pair("blocktime", pindex->GetBlockTime()));
+            }
+            else
+            {
+                entry.push_back(Pair("confirmations", 0));
+            }
+        }
+    }
```
abi87, line 234: Not sure why this appears as new code. It appears to be in the destination branch already and not modified by us

src/sc/sidechain.cpp
```diff
@@ -117,18 +117,18 @@ bool Sidechain::checkCertSemanticValidity(const CScCertificate& cert, CValidatio
         return state.DoS(100, error("version too low"), REJECT_INVALID, "bad-cert-version-too-low");
     }
 
-    if (!MoneyRange(cert.totalAmount) || !MoneyRange(cert.GetValueOut()))
+    if (!MoneyRange(cert.totalAmount) || !MoneyRange(cert.GetValueOfBackwardTransfers()))
```
abi87: Not strictly related to this feature, but it may sufficient to check totalAmount only and skip vouts or the subset of vouts related to backward transfers.
The reason is that CheckCertificate already check range for whole outputs.

abi87 (changes requested):

qa/rpc-tests/sc_cert_and_fwt_in_mempool.py
alsala: Ok, removed

src/primitives/certificate.h
alsala: OK

src/primitives/transaction.h
alsala: Ok

src/primitives/transaction.h
alsala: It should be already initialized by default ctor via SetNull(), but I am adding it explicitly when reading stream

src/rpc/rawtransaction.cpp
alsala: It just add a json entry with the public key hash value in case out is related to a  backward transfer

src/rpc/rawtransaction.cpp
alsala: Probably due to a 'git rebase' command

src/sc/sidechain.cpp
alsala: Ok

PR #240, opened 2020-04-06
Sc tx mining dependencies

alsala:
Handling of dependency of sc creation and fw transfers txes in mining


src/miner.cpp
```diff
@@ -105,14 +105,27 @@ void UpdateTime(CBlockHeader* pblock, const Consensus::Params& consensusParams,
...
+    // TODO add handling of dependancies when vin will be included in certificates
+
     for (auto it_cert = mempool.mapCertificate.begin();
          it_cert != mempool.mapCertificate.end(); ++it_cert)
     {
         const CScCertificate& cert = it_cert->second.GetCertificate();
 
         const uint256& hash = cert.GetHash();
+
+        // a certificate has no dependancies, the creation of the sidechain it refers to has happened
```
abi87: The comment is true, however the `view` below does not encompass the mempool. Check CreateNewBlock function defines the view as CCoinsViewCache view(pcoinsTip).
If you want to check (or asserts) that sc creation is not in mempool, lookup for `cert.scId` in `mempool.mapTx` or, prolly faster, check that `mempool.mapSidechains.at(cert.scId).scCreationTxHash.IsNull()`

src/miner.cpp
```diff
@@ -105,14 +105,27 @@ void UpdateTime(CBlockHeader* pblock, const Consensus::Params& consensusParams,
...
+        // before the first epoch has even started, therefore the creating tx can not be in mempool
+        if (!view.HaveSidechain(cert.scId) )
+        {
+            // This should never happen 
+            LogPrintf("%s():%d - ERROR: mempool certificate missing sidechain\n", __func__, __LINE__);
+            if (fDebug) assert("mempool certificate missing sidechain" == 0);
+            continue;
+        }
+
         double dPriority = it_cert->second.GetPriority(nHeight);
```
abi87, line 132: I believe `GetPriority` is amenable to be refactored to avoid polymorphism. In other words, `CCoinsViewCache::GetPriority` could accept a CTransactionBase as input and if it is a certificate the max priority can be assigned to it

src/miner.cpp
```diff
@@ -155,19 +205,22 @@ void GetBlockTxPriorityData(const CBlock *pblock, int nHeight, int64_t nMedianTi
         unsigned int nTxSize = ::GetSerializeSize(tx, SER_NETWORK, PROTOCOL_VERSION);
         uint256 hash = tx.GetHash();
```
abi87: hash could be redefined as const reference. Also mempool.ApplyDeltas(...) and mempool.ClearPrioritisation(...) could accept as inputs passed by const ref instead of by value

src/miner.cpp
```diff
@@ -128,16 +141,53 @@ void GetBlockCertPriorityData(const CBlock *pblock, int nHeight, std::vector<TxP
...
+bool HandleScDependancy(const CTransaction& tx, const CCoinsViewCache& view, COrphan*& porphan, list<COrphan>& vOrphan, map<uint256, vector<COrphan*> >& mapDependers)
+{
+    // detect dependancies from the sidechain point of view 
+    for (const auto& ft: tx.vft_ccout)
+    {
+        if (mempool.hasSidechainCreationTx(ft.scId))
+        {
+            const uint256& scCreationHash = mempool.mapSidechains.at(ft.scId).scCreationTxHash; 
+            assert(!scCreationHash.IsNull());
+            assert(mempool.exists(scCreationHash));
```
abi87, line 156: Here you should safely assume that mempool internal data structure are well built and coherent and that they can be used without extra checks.
Probably the asserts can be removed at some point.
On the other hand I wonder if you should protect with a lock the access to mempool.mapSidechains

src/miner.cpp
```diff
@@ -155,19 +205,22 @@ void GetBlockTxPriorityData(const CBlock *pblock, int nHeight, int64_t nMedianTi
...
+            // and surely can not be mined in the same block, even in the proper dependancy order
             if (mempool.mapCertificate.count(txin.prevout.hash))
             {
-                LogPrint("cert", "%s():%d - skipping txin of tx[%s] since s out of cert %s\n",
+                LogPrint("cert", "%s():%d - skipping tx[%s] since txin is out of cert[%s] in mempool\n",
                     __func__, __LINE__, tx.GetHash().ToString(), txin.prevout.hash.ToString() ); 
-                dependsOnCertificate = true;
+                dependsOnCertificateInMempool = true;
                 break;
             }
```
abi87, line 228: Here we are explicitly handling the case of a zero-confirmation spending of a certificate in mempool.
I wonder what happen in light of the coming introduction of maturity for certificate spending.
I would remove this handling altogether as it won't happen and I do not think this is the right place to check mempool coherence.
Prolly merging with cert maturity branch won't be a simple automatic merge

src/miner.cpp
```diff
@@ -128,16 +141,53 @@ void GetBlockCertPriorityData(const CBlock *pblock, int nHeight, std::vector<TxP
...
+            // This should never happen; all sc fw transactions in the memory
+            // pool should connect to either sidechain in the chain or sidechain created by
+            // other transactions in the memory pool.
+            LogPrintf("ERROR: mempool transaction missing sidechain\n");
+            if (fDebug) assert("mempool transaction missing sidechain" == 0);
+            if (porphan)
+                vOrphan.pop_back();
+            return false;
+        }
+    }
```
abi87, line 183: I am not sure we should check coherence of dependency of transactions and certificate in mempool/mainchain.
Why don't we limit ourselves to handle them and leave coherence check to mempool.check?

abi87 (changes requested):

src/miner.cpp
alsala: Ok

src/miner.cpp
alsala: Ok, added

src/miner.cpp
alsala: It makes sense, but it would imply making virtual another method called by it, therefore for the time being I would not do it.

src/miner.cpp
alsala: The lock is taken by the caller, but I agree it is useful to add a check that the lock is already taken when excuting that code

src/miner.cpp
alsala: I agree this will be reviewed after the merge with code related to cert maturity

src/miner.cpp
alsala: This should mimic the behaviour of legacy code

PR #241, opened 2020-04-06
Rpc cmd sbh

alsala:
rpc cmds for sphereByHorizen wallet

### DIFF
--- a/qa/rpc-tests/getunconfirmedtxdata.py
+++ b/qa/rpc-tests/getunconfirmedtxdata.py
@@ -35,7 +35,8 @@ class getunconfirmedtxdata(BitcoinTestFramework):
         self.nodes = []
 
         self.nodes = start_nodes(NUMB_OF_NODES, self.options.tmpdir, extra_args=
-            [['-debug=py', '-debug=sc', '-debug=mempool', '-debug=net', '-debug=cert', '-logtimemicros=1', '-spendzeroconfchange=%d'%bool_flag]] * NUMB_OF_NODES)
+            [['-debug=py', '-debug=sc', '-debug=mempool', '-debug=net', '-debug=cert',
+            '-logtimemicros=1', '-rescan', '-spendzeroconfchange=%d'%bool_flag]] * NUMB_OF_NODES)
 
         connect_nodes_bi(self.nodes, 0, 1)
         sync_blocks(self.nodes[1:NUMB_OF_NODES])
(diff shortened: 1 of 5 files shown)


qa/rpc-tests/getunconfirmedtxdata.py
```diff
@@ -31,11 +31,12 @@ def setup_chain(self, split=False):
...
             pass  # Just open then close to create zero-length file
 
-    def setup_network(self, split=False):
+    def setup_network(self, split=False, bool_flag=True):
         self.nodes = []
 
         self.nodes = start_nodes(NUMB_OF_NODES, self.options.tmpdir, extra_args=
-            [['-debug=py', '-debug=sc', '-debug=mempool', '-debug=net', '-debug=cert', '-logtimemicros=1']] * NUMB_OF_NODES)
+            [['-debug=py', '-debug=sc', '-debug=mempool', '-debug=net', '-debug=cert',
+            '-logtimemicros=1', '-rescan', '-spendzeroconfchange=%d'%bool_flag]] * NUMB_OF_NODES)
```
abi87: Not sure why you need the --rescan option here. Consider removing it

src/wallet/rpcwallet.cpp
```diff
@@ -2454,13 +2457,27 @@ UniValue getunconfirmedtxdata(const UniValue &params, bool fHelp)
...
+        else
+        {
+            zconfchangeusage = CWallet::ZCC_FALSE;
+        }
+    }
 
     int n = 0;
-    CAmount amount = pwalletMain->GetUnconfirmedData(script, n);
+    CAmount unconfInput = 0;
+    CAmount unconfOutput;
```
abi87: initialize `unconfOutput` as well

src/wallet/wallet.cpp
```diff
@@ -2851,31 +2878,56 @@ CAmount CWallet::GetUnconfirmedBalance() const
...
     numbOfUnconfirmedTx = 0;
+
+    MapTxWithInputs txOrdered = OrderedTxWithInputsMap(address);
+
+    CBitcoinAddress taddr = CBitcoinAddress(address);
+    if (!taddr.IsValid())
+    {
+        // taddr should be checked by the caller
+        return;
+    }
```
abi87: I would switch the `taddr.IsValid()` with construction of `txOrdered `, so to avoid calling `OrderedTxWithInputsMap` for invalid addresses

src/wallet/wallet.cpp
```diff
@@ -2736,6 +2736,33 @@ bool CWalletTx::IsTrusted() const
...
+        const CWalletObjBase* parent = pwallet->GetWalletTx(txin.prevout.hash);
+        if (parent == NULL)
+            return false;
+        const CTxOut& parentOut = parent->GetVout()[txin.prevout.n];
+        if (pwallet->IsMine(parentOut) != ISMINE_SPENDABLE)
+            return false;
+    }
+    return true;
+}
+
```
abi87: `bool CWalletTx::IsTrusted(bool spendZeroConfChangeIn) const ` is the copy of `bool CWalletTx::IsTrusted() const` with the exception of `spendZeroConfChangeIn `.
Can you write a single function with bool input parameter set to bSpendZeroConfChange?

src/wallet/wallet.cpp
```diff
@@ -4998,6 +5074,11 @@ void CWalletCert::GetAmounts(std::list<COutputEntry>& listReceived, std::list<CO
     }
 }
 
+bool CWalletCert::IsTrusted(bool /*unused*/) const 
+{
+    return IsTrusted();
+}
+
```
abi87: By conflating the two IsTrusted functions into one, we can avoid this pass-through

src/wallet/wallet.h
```diff
@@ -1299,7 +1303,14 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
     std::vector<uint256> ResendWalletTransactionsBefore(int64_t nTime);
     CAmount GetBalance() const;
     CAmount GetUnconfirmedBalance() const;
-    CAmount GetUnconfirmedData(const CScript& scriptToMatch, int& numbOfUnconfirmedTx) const;
+
+    typedef enum {
+        ZCC_FALSE = 0,
+        ZCC_TRUE,
+        ZCC_UNDEF
+    } eZeroConfChangeUsage;
```
abi87: I should be safer to use an enum class instead of a simple enum. For instance, enum classes forbids implicit constructions from int

abi87 (changes requested):

PR #242, opened 2020-04-08
Remove code related to version bit handling in block versions

alsala:
Remove code related to version bit handling in block versions


src/main.h
```diff
@@ -566,12 +559,9 @@ int GetSpendHeight(const CCoinsViewCache& inputs);
  */
 bool IsCommunityFund(const CCoins *coins, int nIn);
 
-extern VersionBitsCache versionbitscache;
-
 /**
 * Determine what nVersion a new block should use.
 */
-int32_t ComputeBlockVersion(const CBlockIndex* pindexPrev, const Consensus::Params& params);
 int32_t ComputeBlockVersion(int nHeight);
```
abi87, line 565: Slightly unrelated maybe, but `int32_t ComputeBlockVersion(int nHeight)` is now called only once by `CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn,  unsigned int nBlockMaxComplexitySize)`. Both of them locks `cs_main`, hence I'd remove this function too

src/primitives/block.h
```diff
@@ -26,7 +26,7 @@ class CBlockHeader
 public:
     // header
     static const size_t HEADER_SIZE=4+32+32+32+4+4+32; // excluding Equihash solution
-    static const int32_t SC_CERT_BLOCK_VERSION = BLOCK_VERSION_3; // defined in consensus.h
+    static const int32_t SC_CERT_BLOCK_VERSION = BLOCK_VERSION_SC_SUPPORT; // defined in consensus.h
```
abi87, line 29: I am not sure why we need this double define anymore? AFAIK `BLOCK_VERSION_SC_SUPPORT `is used only to init `SC_CERT_BLOCK_VERSION `.
Should we keep one only?

src/consensus/consensus.h
```diff
@@ -10,21 +10,21 @@
  *--------------------------
  * this is the most recent block version before sidechain fork. Blocks mined up to such fork
  * will still have this version */
-static const int32_t BLOCK_VERSION_0x2000000  = 0x20000000;
+static const int32_t BLOCK_VERSION_BEFORE_SC  = 0x20000000;
 
 /** old legacy fork */
-static const int32_t BLOCK_VERSION_4          = 0x4;
+static const int32_t BLOCK_VERSION_OLD        = 0x4;
```
abi87, line 16: I'd go here for something like ORIGINAL_BLOCK_VERSION, since it's ZClassic original fork before the chainsplit

src/consensus/consensus.h
```diff
@@ -10,21 +10,21 @@
  *--------------------------
  * this is the most recent block version before sidechain fork. Blocks mined up to such fork
  * will still have this version */
-static const int32_t BLOCK_VERSION_0x2000000  = 0x20000000;
+static const int32_t BLOCK_VERSION_BEFORE_SC  = 0x20000000;
```
abi87, line 13: Here I'd go for SHIELDED_TX_BLOCK_VERSION since this block version is related to the introduction of shielded txs

src/consensus/consensus.h
```diff
@@ -10,21 +10,21 @@
...
+static const int32_t BLOCK_VERSION_OLD        = 0x4;
 
 /** new block version valid from sidechain fork on 
  * We choose this value since:
  * 1. it is a number < 4, this is to enforce that before the sidechain fork and deprecation
  *    date, any block with that value is rejected by both legacy and update node
  * 2. it is the larger number < 4, this is to update the MIN_BLOCK_VERSION macro below, which is
  *    tested in the code for checking block validity  */
-static const int32_t BLOCK_VERSION_3          = 0x3;
+static const int32_t BLOCK_VERSION_SC_SUPPORT = 0x3;
```
abi87, line 24: Here I'd go for SIDECHAIN_BLOCK_VERSION, to signal its introduction is related to sidechains functionalities

src/rpc/mining.cpp
```diff
@@ -562,8 +562,7 @@ UniValue getblocktemplate(const UniValue& params, bool fHelp)
     if (vNodes.empty())
         throw JSONRPCError(RPC_CLIENT_NOT_CONNECTED, "Horizen is not connected!");
 
-    // from https://github.com/ZencashOfficial/zen/commit/e7a774e9a72fae1228ccbc764d520bd685860822
-    if (IsInitialBlockDownload() && ForkManager::getInstance().isAfterChainsplit(chainActive.Tip()->nHeight-(Params().GetConsensus().nMinerConfirmationWindow * 2)))
+    if (IsInitialBlockDownload())
```
abi87, line 565: I doubt this change is correct. You basically disable any call to getblocktemplate until initial block download is done. Why can't we just disable till chainActive.Tip()->nHeight is after chain split?

qa/rpc-tests/forknotify.py
```diff
@@ -79,30 +78,6 @@ def run_test(self):
             v = self.nodes[0].getblock(str(i))['version']
             assert_equal(v, UP_VERSION)
 
-        # -alertnotify should trigger on the 51'st,
```
abi87, line 82: Lines 46/7: I may be wrong here, but I do not see the option -blockversion among those displayed by zend binary. Am I wrong or this option is somehow outdated/unused?

abi87 (changes requested):

src/rpc/mining.cpp
abi87: Scratch this comment. I rechecked `IsInitialBlockDownload()` and indeed if `IsInitialBlockDownload() == true` than `ForkManager::getInstance().isAfterChainsplit(chainActive.Tip()->nHeight == false`

src/main.h
alsala: Ok

qa/rpc-tests/forknotify.py
alsala: It is a regtest only option, I see it can be displayed via:
  src/zend -regtest --help -help-debug=1

src/consensus/consensus.h
alsala: We will a have a final review of names, leaving this for the time being

src/consensus/consensus.h
alsala: We will a have a final review of names, leaving this for the time being

src/consensus/consensus.h
alsala: Ok, changed to BLOCK_VERSION_ORIGINAL

src/primitives/block.h
alsala: Ok, removed class member

PR #243, opened 2020-04-16
Solving Issue 233 - Maturity on bwt

abi87:
This pull request solves issue233, i.e. forbids spending of outputs coming from a backward transfer until a certificate for sequent epoch is received and accepted in mainchain.
Checks are performed at an output level, so that outputs coming from a change are indeed amenable for immediate redemption


qa/rpc-tests/sc_cert_base.py
```diff
@@ -63,36 +63,40 @@ def run_test(self):
...
         fwd_tx = self.nodes[0].sc_send("abcd", fwt_amount, scid)
-        mark_logs("Node 0 transfers {} coins to SC with tx {}...".format(fwt_amount, fwd_tx), self.nodes, DEBUG_MODE)
+        mark_logs("Node0 transfers {} coins to SC with tx {}...".format(fwt_amount, fwd_tx), self.nodes, DEBUG_MODE)
         self.sync_all()
 
         mark_logs("Node0 confirms fwd transfer generating 1 block", self.nodes, DEBUG_MODE)
         blocks.extend(self.nodes[0].generate(1))
         self.sync_all()
 
+        # TODO: Check node0 balance following fwd transfert
```
alsala: This is marked as TODO and should be added

qa/rpc-tests/sc_cert_base.py
```diff
@@ -203,31 +199,90 @@ def run_test(self):
...
+        assert_equal(retrieved_cert['details'][0]['amount'], amount_cert_1[0]["amount"])  # In cert details you can see the actual amount transferred
+
+        mark_logs("Checking Node1 balance is duly updated,".format(epoch_number), self.nodes, DEBUG_MODE)
+        assert_equal(bal_after_cert_2, bal_before_cert_2  + amount_cert_1[0]["amount"])
+
+        Node2_bal_before_cert_expenditure = self.nodes[2].getbalance("", 0)
+        mark_logs("Checking that Node1 can spend coins received from bwd transfer in previous epoch", self.nodes, DEBUG_MODE)
+        mark_logs("Node 1 sends {} coins to node2...".format(amount_cert_1[0]["amount"] / 2), self.nodes, DEBUG_MODE)
+        tx = self.nodes[1].sendtoaddress(self.nodes[2].getnewaddress(), amount_cert_1[0]["amount"] / 2)
         assert(len(tx) > 0)
```
alsala: The sendtoaddress cmd above could be performed in a try/catch block instead of asserting len

src/coins.cpp
```diff
@@ -1025,6 +1025,35 @@ bool CCoinsViewCache::UpdateScInfo(const CScCertificate& cert, CBlockUndo& block
     return true;
 }
 
+bool CCoinsViewCache::IsOutputMature(const uint256& txHash, unsigned int pos) const
+{
+    if (HaveCoins(txHash) == 0)
```
alsala: HaveCoins returns a boolean, consider replacing it with !HaveCoins(). Moreover this call could be avoided by testing the ptr returned by AccessCoins() below, since both call FetchCoins()

src/coins.cpp
```diff
@@ -1025,6 +1025,35 @@ bool CCoinsViewCache::UpdateScInfo(const CScCertificate& cert, CBlockUndo& block
...
+        return false;
+
+    const CCoins* pRefCoin =  AccessCoins(txHash);
+
+    if (!pRefCoin->IsAvailable(pos))
```
alsala, line 1035: Consider testing ptr before dereferencing it, something like: if (!pRefCoin || !pRefCoin->IsAvailable(pos))

src/main.cpp
```diff
@@ -746,7 +713,12 @@ bool IsStandardTx(const CTransaction& tx, string& reason, const int nHeight)
 
     unsigned int nDataOut = 0;
     txnouttype whichType;
-    BOOST_FOREACH(const CTxOut& txout, tx.GetVout()) {
+    BOOST_FOREACH(const CTxOut& txout, txBase.GetVout()) {
+
+        // if the output comes from a backward transfer (when we are a certificate), skip this check
+        // but go on if the certificate txout is an ordinary one
+        if (txout.isFromBackwardTransfer)
```
alsala: Why are we skipping this check? Also, we are also skipping the IsDust() check below.

src/main.cpp
```diff
@@ -1199,20 +1173,31 @@ bool AcceptCertificateToMemoryPool(CTxMemPool& pool, CValidationState &state, co
...
 
-    // is it already in the memory pool?
     uint256 certHash = cert.GetHash();
-    if (pool.existsCert(certHash))
+
+    // Check if cert is already in mempool or if there are conflicts with in-memory certs
     {
-        LogPrint("mempool", "Dropping certificateId %s : already in mempool\n", certHash.ToString());
-        return false;
+        LOCK(pool.cs);
```
alsala, line 1179: Consider moving these checks in the block below, where the lock on mempool is taken again. The same holds for AcceptToMemPool() below.

src/main.cpp
```diff
@@ -1471,8 +1449,27 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
...
+            for(const CTxIn& in: tx.GetVin()) {
+                const CCoins *coins = view.AccessCoins(in.prevout.hash);
+                assert(coins);
+
+                if (coins->IsFromCert()) { //This is just to avoid following assert for non-cert coins
+                    // HaveInputs above checks for utxos availability. So accessing vout[in.prevout.n] is safe
+                    assert(coins->IsAvailable(in.prevout.n));
+
+
+                    //check on vout is better than coins->IsFromCert() since it'll skip loading scInfo for zero bwt amount certs
```
alsala, line 1457: This comment is not clear to me: how is it better?

src/primitives/certificate.cpp
```diff
@@ -198,24 +207,6 @@ bool CScCertificate::IsApplicableToState(CValidationState& state, int nHeight) c
...
-    
-bool CScCertificate::IsStandard(std::string& reason, int nHeight) const
-{
-    if (!zen::ForkManager::getInstance().areSidechainsSupported(nHeight))
-    {
-        reason = "version";
-        return false;
-    }
-
-    // so far just one version is supported, but in future it might not be the case
```
alsala, line 209: Why this check has been removed?

src/txmempool.cpp
```diff
@@ -362,40 +361,55 @@ void CTxMemPool::remove(const CScCertificate &origCert, std::list<CTransaction>&
...
+        for(const CTxIn& txin: tx.GetVin()) {
             std::map<uint256, CTxMemPoolEntry>::const_iterator it2 = mapTx.find(txin.prevout.hash);
             if (it2 != mapTx.end())
                 continue;
 
-            // if input is a certificate skip as well, even if it can not be coinbase anyway
             std::map<uint256, CCertificateMemPoolEntry>::const_iterator it3 = mapCertificate.find(txin.prevout.hash);
-            if (it3 != mapCertificate.end())
-            {
+            if (it3 != mapCertificate.end()) {
```
alsala, line 377: Should never happen, a certificate in the mempool can not be an input of a tx, consider removing the tx and tracing an error

src/wallet/wallet.cpp
```diff
@@ -1921,14 +1862,30 @@ CAmount CWallet::GetDebit(const CTransactionBase& txBase, const isminefilter& fi
...
     CAmount nCredit = 0;
-    for(const CTxOut& txout: txBase.GetVout()) {
-        nCredit += GetCredit(txout, filter);
+    fCanBeCached = true;
+    for(unsigned int pos = 0; pos < txBase.GetVout().size(); ++pos) {
+        if (txBase.IsCertificate()) { //optimization since non-certs have mature vouts only
+
+            if (!pcoinsTip->IsOutputMature(txBase.GetHash(), pos)) {
+                fCanBeCached = false;
+                if (!keepImmatureVoutsOnly) continue;
```
alsala: This apparently is skipped only if flag is false. If flag is true an immature amount can be part of GetCredit computation below?

src/coins.cpp
abi87: Agree, following bugfixing the HaveCoins + AccessCoins is changed to GetCoins which addresses both your points

src/main.cpp
abi87: We already touched upon this point in the past. I agree that locking once may be more efficient (not sure how to measure it though). However the two locked blocks gives better readability since they correspond to two different operations: first block contains "light" checks which peek into mempool datastructs only, while second blocks contains "less-light" checks picking coins from the caches system. I'd keep them separated for readability, at least currently

qa/rpc-tests/sc_cert_base.py
abi87: Agree. Indeed these extra checks have been precious as they uncovered faulty handling of some corner cases

src/primitives/certificate.cpp
abi87: version check has been moved to CScCertificate::CheckVersionIsStandard.
CheckOutputsAreStandard has been absorbed into IsStandard function

src/wallet/wallet.cpp
abi87: Agree it was a bug. Fixed and extended sc_cert_base.py to prove it

src/txmempool.cpp
abi87: agree i'll keep "dependant" tx marked for removal and improve tracing

src/main.cpp
abi87: Probably poorly put, I am that we need to load in the cache system information about sidechains iif spent input is a certificate with backward transfers. Conversely, we won't load sidechains info for both ordinary tx and for certs with no bwt

qa/rpc-tests/sc_cert_base.py
abi87: agree, done

src/main.cpp
alsala: We should not skip this check, it would be better to explicitly skip the replay-protection related check few lines below,as is it done for coinbase.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -614,11 +614,7 @@ bool AsyncRPCOperation_sendmany::main_impl() {
             int wtxDepth = -1;
             {
                 LOCK2(cs_main, pwalletMain->cs_wallet);
-#if 0
-                const CWalletTx& wtx = pwalletMain->mapWallet[jso.hash];
-#else
-                const CWalletObjBase& wtx = *(pwalletMain->mapWallet[jso.hash]);
-#endif
+                const CWalletObjBase& wtx = *(pwalletMain->getMapWallet().at(jso.hash));
```
alsala, line 617: the at() throws a std exception, the behaviour would be different from previous code

src/wallet/wallet.cpp
```diff
@@ -2497,178 +2403,201 @@ CAmount CWalletObjBase::GetDebit(const isminefilter& filter) const
...
+        switch(this->IsOutputMature(pos)) {
+        case COutputEntry::maturityState::MATURE:
+            return true;
+
+        case COutputEntry::maturityState::IMMATURE:
+            continue;
+
+        case COutputEntry::maturityState::NOT_APPLICABLE:
+        default:
+            return false;
```
alsala, line 2419: Why should maybe continue in this case, instead of returning from method wth false value

src/wallet/wallet.cpp
```diff
@@ -2497,178 +2403,201 @@ CAmount CWalletObjBase::GetDebit(const isminefilter& filter) const
...
-        {
-            const CTxOut &txout = vout[i];
-            nCredit += pwallet->GetCredit(txout, ISMINE_SPENDABLE);
+        COutputEntry::maturityState outputMaturity = this->IsOutputMature(pos);
+        if (outputMaturity == COutputEntry::maturityState::NOT_APPLICABLE) {
+            fAvailableCreditCached = false;
+            return CAmount(0);
+        }
+
+        if (outputMaturity == COutputEntry::maturityState::IMMATURE) {
```
alsala, line 2534: Would it be better to check against != MATURE? In this case we also would skip the NOT_APPLICBLE. The same for other similar checks

src/wallet/wallet.cpp
abi87: `COutputEntry::maturityState::NOT_APPLICABLE` applies to coins not in mempool nor in mainchain. As such it is valid for all outputs

src/wallet/wallet.cpp
abi87: `COutputEntry::maturityState::NOT_APPLICABLE` being valid for the tx as a whole, we can have an early return by distinguishing the cases

src/main.cpp
abi87: Agreed, as you explained AFK I'll skip only the isTransactionTypeAllowedAtHeight check for bwt and run the rest of checks on all outputs

PR #244, opened 2020-04-24
Sc blockundo ver

alsala:
Handling of versioning in the CBlackUndo class

### DIFF
--- a/src/gtest/test_sidechain.cpp
+++ b/src/gtest/test_sidechain.cpp
@@ -9,6 +9,23 @@
 #include <undo.h>
 #include <main.h>
 
+class CBlockUndo_OldVersion
+{
+    public:
+        std::vector<CTxUndo> vtxundo;
+        uint256 old_tree_root;
+
+        ADD_SERIALIZE_METHODS;
+
+        template <typename Stream, typename Operation>
+        inline void SerializationOp(Stream& s, Operation ser_action, int nType, int nVersion) {
+            READWRITE(vtxundo);
+            READWRITE(old_tree_root);
+        }   
+};
+
+
+
 class CInMemorySidechainDb final: public CCoinsView {
 public:
     CInMemorySidechainDb()  = default;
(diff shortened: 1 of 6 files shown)


src/undo.h
```diff
@@ -86,21 +86,87 @@ struct ScUndoData
...
 {
+    /** Magic number read from the value expressing the size of vtxundo vector.
+     *  It is used for distinguish new version of CBlockUndo instance from old ones.
+     *  The maximum number of tx in a block is roughly MAX_BLOCK_SIZE / MIN_TX_SIZE, which is:
+     *   2M / 61bytes =~ 33K = 0x8012
+     * Therefore the magic number must be a number greater than this limit. */
+    static const uint16_t _marker = 0xfec1;
+
+    /** memory only */
+    bool isNewVersion;
```
abi87: Here I'd go for something more telling, like "fIncludeSidechainAttributes"

src/undo.h
```diff
@@ -86,21 +86,87 @@ struct ScUndoData
...
+        str += strprintf("vtxundo.size %u\n", vtxundo.size());
+        str += strprintf("old_tree_root %s\n", old_tree_root.ToString().substr(0,10));
+        str += strprintf("msc_iaundo.size %u\n", msc_iaundo.size());
+        str += strprintf(" ===> tot size %u\n", GetSerializeSize(SER_NETWORK, PROTOCOL_VERSION));
+        return str;
     }
 
-    std::string ToString() const;
-};
+    /** for tresting */
```
abi87: minor typo here

src/gtest/test_sidechain.cpp
```diff
@@ -1020,6 +1037,107 @@ TEST_F(SidechainTestSuite, CSidechainFromMempoolRetrievesUnconfirmedInformation)
     EXPECT_TRUE(retrievedInfo.mImmatureAmounts.at(-1) == fwdAmount);
 }
 
+TEST_F(SidechainTestSuite, CSidechainBlockUndoVersioning) {
+
```
abi87, line 1041: In other UTs involving serialization ops, I have used a `CDataStream` type. Is it reusable here? I'd make these UTs more isolated wrt the environment and prove the

src/undo.h
```diff
@@ -86,21 +86,87 @@ struct ScUndoData
...
+        }
+        else
+        {
+#if 1
+            s.Rewind(GetSizeOfCompactSize(nSize));
+            ::Unserialize(s, (vtxundo), nType, nVersion);
+#else
+            ::AddEntriesInVector(s, vtxundo, nType, nVersion, nSize);
+#endif
+            ::Unserialize(s, (old_tree_root), nType, nVersion);
```
abi87: If I get you right, these two branches are both valid implementations. I'd go for the  `::AddEntriesInVector` one, since it avoids specifying extra rewind operations.
The reason I'd like to avoid the rewind is because it leverages on fseek which, as I understand, have different semantics for binary or text files.

src/gtest/test_sidechain.cpp
```diff
@@ -1020,6 +1037,107 @@ TEST_F(SidechainTestSuite, CSidechainFromMempoolRetrievesUnconfirmedInformation)
...
+    }
+    catch(...)
+    {
+        good_read = false;
+    }
+    EXPECT_TRUE(good_read == false);
+
+    filein.fclose();
+}
+
```
abi87: If we go for the `::AddEntriesInVector ` alternative, I'd restructure this UTs to prove the following points:

1. The new serialization/deserialization pair can successfully write and read in memory CBlocksUndo with and without sidechain related attributes
2. Given an explicit string represented the serialized version of the pre-sidechain CBlockUndo, the new serialization ops handle it allright.
Check out /src/test/coins_tests.cpp for a reference

src/main.cpp
```diff
@@ -2235,6 +2235,9 @@ bool UndoReadFromDisk(CBlockUndo& blockundo, const CDiskBlockPos& pos, const uin
     CHashWriter hasher(SER_GETHASH, PROTOCOL_VERSION);
     hasher << hashBlock;
     hasher << blockundo;
+
+    LogPrint("sc", "%s\n", blockundo.ToString());
+
```
abi87, line 2240: How big is this blockundo? I wonder if this log does affect prod performances in any way

abi87 (changes requested):

src/gtest/test_sidechain.cpp
alsala: Agree, on the other side is also useful testing the specific stream CAutoFile, which is used by the code for reading/writing the undo file

src/undo.h
alsala: ok

src/undo.h
alsala: ok

src/undo.h
alsala: Ok, you are right

src/gtest/test_sidechain.cpp
alsala: I will add that test to existing one

src/main.cpp
alsala: We are printing the log if the category is explicitly set in the -debug option list

PR #245, opened 2020-04-28
Sc msg cert removal

alsala:
Removed MSG_CERT which was added to network protocol for handling certificates and made MSG_TX message able to handle both transactions and certificates.


src/primitives/transaction.h
```diff
@@ -705,7 +705,8 @@ class CTransactionBase
         return !(a==b);
     }
 
-    virtual bool IsScVersion() const = 0;
+    virtual bool IsCertificateVersion() const { return false; }
```
abi87: The method `IsCertificateVersion` is never called on a CTransaction. For ScCertificate it is called only once, in `Sidechain::checkCertSemanticValidity(const CScCertificate& cert, CValidationState& state)`
I would remove this method altogether and explicitly compare the versions in 'Sidechain::checkCertSemanticValidity(const CScCertificate& cert, CValidationState& state)'

src/primitives/transaction.h
```diff
@@ -705,7 +705,8 @@ class CTransactionBase
...
+    virtual bool IsScVersion() const { return false; }
```
abi87: `IsScVersion` is called only on CTransaction, never on CTransactionBase.
I would Remove it for CTransactionBase. Maybe also rename to IsScRelated

abi87 (changes requested):

src/primitives/transaction.h
alsala: Ok

src/primitives/transaction.h
alsala: Ok, but for now I leave the name as it is

PR #246, opened 2020-05-04
Sc cert vin

alsala:
Added vector of inputs to certificate object in order to be able to have a fee charged to the sender of the certificate


src/txmempool.cpp
```diff
@@ -690,29 +751,72 @@ void CTxMemPool::check(const CCoinsViewCache *pcoins) const
...
+                assert(tx2.GetVout().size() > txin.prevout.n && !tx2.GetVout()[txin.prevout.n].IsNull());
+                fDependsWait = true;
+            } else {
+                // maybe our input is a certificate?
+                std::map<uint256, CCertificateMemPoolEntry>::const_iterator itCert = mapCertificate.find(txin.prevout.hash);
+                if (itCert != mapCertificate.end()) {
+                    const CTransactionBase& cert = itCert->second.GetCertificate();
+                    assert(cert.GetVout().size() > txin.prevout.n && !cert.GetVout()[txin.prevout.n].IsNull());
+                    fDependsWaitCert = true;
+                }
```
abi87: We currently do not allow zero-conf spending of certificate outputs, neither bwt .
nor changes.
While I would leave in the code for recursive removal of txs or certs zero-conf spending the cert output, I would enforce the impossibility of this spending in the check.
Hence here I would assert that vin is not a certificate in mempool.
Same for transaction inputs above

src/txmempool.cpp
```diff
@@ -187,39 +198,58 @@ bool CTxMemPool::addUnchecked(const uint256& hash, const CCertificateMemPoolEntr
...
+                for(const auto& sc: origTx.GetVscCcOut()) {
+                    if (mapSidechains.count(sc.scId) == 0)
+                        continue;
+ 
+                    if (removeDependantFwds) {
+                        for(const auto& fwdTxHash : mapSidechains.at(sc.scId).fwdTransfersSet)
+                            objToRemove.push_back(fwdTxHash);
+                    } else
+                        objToRemove.push_back(mapSidechains.at(sc.scId).scCreationTxHash);
+                }
```
abi87: You can probably avoid the duplication as follows
`    // Remove transaction from memory pool
    {
        LOCK(cs);
        std::deque<uint256> objToRemove;
        objToRemove.push_back(origTx.GetHash());

        if(!fRecursive)
            return removeInternal(objToRemove, removedTxs, removedCerts, fRecursive, removeDependantFwds);

        if (!mapCertificate.count(origTx.GetHash()) && !mapTx.count(origTx.GetHash())) {
            // If recursively removing but origTx isn't in the mempool
            // be sure to remove any children that are in the pool. This can
            // happen during chain re-orgs if origCert isn't re-accepted into
            // the mempool for any reason.
            for (unsigned int i = 0; i < origTx.GetVout().size(); i++) {
                std::map<COutPoint, CInPoint>::iterator it = mapNextTx.find(COutPoint(origTx.GetHash(), i));
                if (it == mapNextTx.end())
                    continue;
                LogPrint("sc", "%s():%d - adding tx [%s] to list for removing\n", __func__, __LINE__, it->second.ptx->GetHash().ToString());
                objToRemove.push_back(it->second.ptx->GetHash());
            }

            for(const auto& sc: origTx.GetVscCcOut()) {
                if (mapSidechains.count(sc.scId) == 0)
                    continue;

                if (removeDependantFwds) {
                    for(const auto& fwdTxHash : mapSidechains.at(sc.scId).fwdTransfersSet)
                        objToRemove.push_back(fwdTxHash);
                } else
                    objToRemove.push_back(mapSidechains.at(sc.scId).scCreationTxHash);
            }
        }

        removeInternal(objToRemove, removedTxs, removedCerts, fRecursive, removeDependantFwds);
    }`

src/txmempool.cpp
```diff
@@ -690,29 +751,72 @@ void CTxMemPool::check(const CCoinsViewCache *pcoins) const
...
+            assert(it3->second.ptx->GetHash() == cert.GetHash());
+            assert(it3->second.n == i);
+            i++;
+        }
+
         checkTotal += it->second.GetCertificateSize();
         innerUsage += it->second.DynamicMemoryUsage();
         CValidationState state;
 
         assert(cert.ContextualCheckInputs(state, mempoolDuplicate, false, chainActive, 0, false, Params().GetConsensus(), NULL));
```
abi87: this polymorphic check is hollow, i.e. it return true for certificates and ::ContextualCheckInputs on txs.
You should remove it and change it for ::ContextualCheckInputs since currently it accepts a CTransactionBase object

src/txmempool.cpp
```diff
@@ -318,99 +348,109 @@ void::CTxMemPool::removeInternal(
...
-            LogPrint("cert", "%s():%d - removing cert [%s] from mempool\n", __func__, __LINE__, hash.ToString() );
+            LogPrint("mempool", "%s():%d - removing cert [%s] from mempool\n", __func__, __LINE__, hash.ToString() );
             mapCertificate.erase(hash);
             nCertificatesUpdated++;
         }
     }
 }
 
-void CTxMemPool::remove(const CScCertificate &origCert, std::list<CTransaction>& removedTxs, std::list<CScCertificate>& removedCerts, bool fRecursive, bool removeDependantFwds)
+bool CTxMemPool::addToListForRemovalImmatureExpenditures(
```
abi87: I would inline this

src/txmempool.cpp
```diff
@@ -318,99 +348,109 @@ void::CTxMemPool::removeInternal(
                 }
```
abi87, line 331: I comment here from previous block. Given polymorphism of scOut and fwdOut, you could avoid duplication of the `if(fRecursive)` block which refresh the list of objects to remove and keep different removal of current object, as it is now

src/txmempool.h
```diff
@@ -95,11 +95,11 @@ class CBlockPolicyEstimator;
 class CInPoint
 {
 public:
-    const CTransaction* ptx;
+    std::shared_ptr<const CTransactionBase> ptx;
```
abi87: I believe a nake pointer should be equally fine here

src/main.cpp
```diff
@@ -1188,18 +1174,42 @@ bool AcceptCertificateToMemoryPool(CTxMemPool& pool, CValidationState &state, co
...
 
+        for (const CTxIn & vin : cert.GetVin()) {
+            if (pool.mapNextTx.count(vin.prevout)) {
+                LogPrint("mempool", "%s():%d - Dropping cert %s : it double spends input of [%s] that is in mempool\n",
+                    __func__, __LINE__, certHash.ToString(), vin.prevout.hash.ToString());
+                return state.DoS(0,
+                         error("AcceptToMemoryPool: Dropping cert %s : it double spends input of another tx in mempool", certHash.ToString()),
+                         REJECT_INVALID, "double spend");
+            }
+            if (pool.existsCert(vin.prevout.hash)) {
```
abi87: existsCert acquire the mempool lock, which here is not necessary. You can peek into the mapCertificate directly. Same comment for same check in AcceptTxToMemoryPool

src/wallet/wallet.cpp
```diff
@@ -2445,8 +2493,9 @@ COutputEntry::maturityState CWalletObjBase::IsOutputMature(unsigned int vOutPos)
         if ((nDepth == 0) && vout[vOutPos].isFromBackwardTransfer)
             return COutputEntry::maturityState::IMMATURE;
 
+        // we do not allow even unconfirmed change output in a certificate
         if ((nDepth == 0) && !vout[vOutPos].isFromBackwardTransfer)
-            return COutputEntry::maturityState::MATURE;
+            return COutputEntry::maturityState::IMMATURE;
```
abi87: You can put all together writing:
```
if (nDepth == 0)
   return COutputEntry::maturityState::IMMATURE;
```

src/script/interpreter.cpp
```diff
@@ -1140,6 +1140,97 @@ class CTransactionSignatureSerializer {
...
+            if ((*it).isFromBackwardTransfer)
+            {
+                CBackwardTransferOut btout((*it));
+                vbt_ccout_ser.push_back(btout);
+            }
+            else
+            {
+                vout_ser.push_back(*it);
+            }
+        }
```
abi87, line 1221: This serialization facility seems to be used only for hash calculation. I do not see why we need to go back to vbt vs vout vectors. Can't we serialize a single vout, including the isFromBackwardTransfer flag?

abi87 (changes requested):

src/txmempool.cpp
alsala: Ok, you are right

src/txmempool.cpp
alsala: Ok

src/txmempool.cpp
alsala: Ok

src/txmempool.cpp
alsala: Ok

src/txmempool.cpp
alsala: Here probably is not that clean as the block above, I am leaving the code as it is

src/txmempool.h
alsala: Ok, you are right, it is useless here

src/script/interpreter.cpp
alsala: The serialization is used for the signature, that must be applied on the serialized obj

src/main.cpp
alsala: Ok

PR #247, opened 2020-05-04
Cleanup coins from ceased sidechains

abi87:
I have introduced a sidechainHandler to remove coins from ceased sidechains [of course only those coins immediately preceding the cease epoch]. Also forward transfers to ceased sidechains are forbidden.
I have changed the serialization of BlockUndo in order to record key infos for removed coins.
A nice side-effect is that I have centralized and unit tested the sidechain expiration checks


src/coins.cpp
```diff
@@ -782,6 +829,9 @@ bool CCoinsViewCache::RestoreImmatureBalances(int blockHeight, const CBlockUndo&
 
             cacheSidechains.at(scId).flag = CSidechainsCacheEntry::Flags::DIRTY;
         }
+
+        cacheSidechains.at(scId).scInfo.lastCertificateHash = lastCertHash;
```
alsala: for sake of uniformity this should be removed from here and added to 827 above, together with the setting of targetScInfo.lastEpochReferencedByCertificate, that would also spare a useless DIRTY set

src/gtest/test_sidechain_ceased.cpp
```diff
@@ -0,0 +1,487 @@
...
+};
+
+
+///////////////////////////////////////////////////////////////////////////////
+/////////////////////////// isSidechainCeased /////////////////////////////////
+///////////////////////////////////////////////////////////////////////////////
+TEST_F(CeasedSidechainsTestSuite, UnknownSidechainIsNeitherAliveNorCeased) {
+    uint256 scId = uint256S("aaa");
+    int creationHeight = 1912;
+    CTransaction scCreationTx = txCreationUtils::createNewSidechainTxWith(scId, CAmount(10));
```
alsala: Creating this is probably useless

src/sc/sidechain.h
```diff
@@ -33,6 +55,9 @@ class CSidechain {
     // last epoch for which a certificate have been received
     int lastEpochReferencedByCertificate;
```
alsala, line 56: This should also be added in the output of getscinfo RPC cmd

src/sc/sidechain.h
```diff
@@ -78,6 +109,15 @@ namespace Sidechain {
     bool hasScCreationOutput(const CTransaction& tx, const uint256& scId);
 
     bool checkCertSemanticValidity(const CScCertificate& cert, CValidationState& state);
+
+
+    enum class state {
```
alsala, line 114: This also should be added in the output of getscinfo RPC cmd

src/gtest/test_sidechain_ceased.cpp
abi87: Agree, and it doesn't even help readability. Removed

src/sc/sidechain.h
abi87: Agree, but as discussed AFK,  let's make it subject on subsequent PR

src/sc/sidechain.h
abi87: Agree, but as discussed AFK,  let's make it subject on subsequent PR

PR #254, opened 2020-05-15
Remove get sc id from tx base && splitted UpdateCoins

abi87:


src/miner.cpp
```diff
@@ -605,7 +606,11 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn,  unsigned int nBlo
             if (!tx.ContextualCheckInputs(state, view, true, chainActive, MANDATORY_SCRIPT_VERIFY_FLAGS | SCRIPT_VERIFY_CHECKBLOCKATHEIGHT, true, Params().GetConsensus()))
                 continue;
 
-            UpdateCoins(tx, state, view, nHeight);
+            CTxUndo dummyUndo;
```
alsala, line 609: consider using a try/catch block since a failing dynamic cast on references throws std::bad_cast exception

abi87: As discussed AFK, we'll trap the potential bad_cast and then assert(false), since it really should not happen

PR #255, opened 2020-05-15
Removed getters for cc out from txbase; splitted method for handling …

alsala:
…dependencies in miner


src/miner.cpp
```diff
@@ -168,9 +161,10 @@ void HandleDependancy(const CTransactionBase& txBase, int nHeight, double dPrior
             if (fDebug) assert("mempool transaction unspendable input that is an unconfirmed certificate output" == 0);
 
             LogPrintf("%s():%d - ERROR: mempool [%s] spends output from certificate[%s] in mempool\n",
-                __func__, __LINE__, txBase.GetHash().ToString(), txin.prevout.hash.ToString() );
-            dependsOnCertificateInMempool = true;
-            break;
+                __func__, __LINE__, hash.ToString(), txin.prevout.hash.ToString() );
+            if (porphan)
+                vOrphan.pop_back();
```
abi87: If I remember right, we agreed to move the pop_back in case porphan has been filled outside of this function, to its caller.
Same comment applies to VerifyForwardTransfersDependencies

abi87: Also maybe we can have a single LogPrintf on line 159 and eliminate that on line 163

src/miner.cpp
```diff
@@ -188,15 +182,16 @@ void HandleDependancy(const CTransactionBase& txBase, int nHeight, double dPrior
...
-    {
-        // should never happen because that means inconsistency in mempool, but this tx must not be
-        // added to vecPriority nor in the vOrphan
-        LogPrint("cert", "%s():%d - skipping [%s] for invalid dependancies\n",
-            __func__, __LINE__, txBase.GetHash().ToString() ); 
-        return;
-    }
+bool AddTxToPriorities(const CTransactionBase& txBase, const CCoinsViewCache& view, CAmount& nTotalIn,
+                       int nHeight, double dPriorityIn, CAmount nFeeIn, list<COrphan>& vOrphan, COrphan*& porphan, vector<TxPriority>& vecPriority)
+{
```
abi87: I would like to have in this function a single if(!porphan) instead of two.
It should be more readable

src/txmempool.cpp
```diff
@@ -222,14 +222,26 @@ void CTxMemPool::remove(const CTransactionBase& origTx, std::list<CTransaction>&
...
+                                objToRemove.push_back(fwdTxHash);
+                        } else
+                            objToRemove.push_back(mapSidechains.at(sc.scId).scCreationTxHash);
+                    }
+                }
+                catch(...)
+                {
+                    // should never happen
+                    LogPrintf("%s():%d - could not make a tx from obj[%s]\n", __func__, __LINE__, origTx.GetHash().ToString());
+                }
```
abi87: I'd make the try-catch more local, surrounding the dynamic_cast only. Once this is done, the rest of the code shouldn't throw exceptions

abi87: Also in this way you can specify the expection to be caught, namely std::bad_cast

abi87 (changes requested):

src/txmempool.cpp
alsala: Done using a ptr instead of a reference

src/miner.cpp
alsala: done

src/miner.cpp
alsala: Done

src/wallet/wallet.cpp
```diff
@@ -2492,9 +2492,10 @@ CCoinsViewCache::outputMaturity CWalletObjBase::IsOutputMature(unsigned int vOut
 
     if (IsCertificate())
     {
-        // we do not consider even an unconfirmed change output in a certificate as mature
-        if ((nDepth == 0))
-            return CCoinsViewCache::outputMaturity::IMMATURE;
+        // we consider an unconfirmed change output in a certificate as mature
+        // even if that can only be spent by a certificate (not by a transaction)
+        if ((nDepth == 0) && !GetVout()[vOutPos].isFromBackwardTransfer)
```
abi87: probably best checking for availability of vOutPos entry. Also we should explicit return Immature for backward transfers in mempool, without going on to `pcoinsTip->IsCertOutputMature` call

abi87: Another point. IsOutputMature is widely use for both balance calculations and actual search of unspent utxos. If you choose to declare change output mature since it may be used for financing certificates, you need then to crosscheck, that caller use this utxo to finance only a certificate and not a tx.

PR #267, opened 2020-05-23
computation of serialized size method made virtual in tx base

alsala:


src/primitives/transaction.h
```diff
@@ -667,8 +668,8 @@ class CTransactionBase
     virtual bool CheckVersionIsStandard   (std::string& reason, int nHeight) const = 0;
     virtual bool CheckInputsAvailability  (CValidationState &state) const = 0;
     virtual bool CheckOutputsAvailability (CValidationState &state) const = 0;
-    virtual bool CheckSerializedSize      (CValidationState &state) const = 0;
 
+    bool CheckSerializedSize (CValidationState &state) const;
```
abi87, line 670: If we want a single `CheckSerializedSize` there must be a single max size for both txs and certs. Remove `MAX_CERT_SIZE ` in consensus.h

abi87 (changes requested):
Just remove MAX_CERT_SIZE useless define

src/primitives/transaction.h
abi87: Done

PR #281, opened 2020-06-02
Fixup wallet sync to mainchain

abi87:


src/gtest/test_sidechain_ceased.cpp
```diff
@@ -1002,3 +1002,189 @@ TEST_F(CeasedSidechainsTestSuite, UndoEmptyCertUpdatesToCeasingScs) {
...
+
+    //Generate coin from cert, to check it is fully reconstructed from BlockUndo
+    CTxUndo dummyTxUndo;
+    int certHeight = 1987;
+    EXPECT_FALSE(view->HaveCoins(cert.GetHash()));
+    UpdateCoins(cert, *view, dummyTxUndo, certHeight);
+    CCoins coinFromCert;
+    EXPECT_TRUE(view->GetCoins(cert.GetHash(),coinFromCert));
+
+    //Create Tx spending first change from the certificate
```
alsala: comment is probably a cut/paste and should be updated

src/gtest/test_sidechain_ceased.cpp
```diff
@@ -1002,3 +1002,189 @@ TEST_F(CeasedSidechainsTestSuite, UndoEmptyCertUpdatesToCeasingScs) {
...
```
alsala: Update comment, maybe explain that prevout.n is 0 since we have 1 bwt and no change

src/main.cpp
```diff
@@ -1918,6 +1918,40 @@ void static InvalidBlockFound(CBlockIndex *pindex, const CValidationState &state
...
+ * @param out The out point that corresponds to the tx input.
+ * @return True on success.
+ */
+bool ApplyTxInUndo(const CTxInUndo& undo, CCoinsViewCache& view, const COutPoint& out)
+{
+    bool fClean = true;
+
+    CCoinsModifier coins = view.ModifyCoins(out.hash);
+    if (undo.nHeight != 0)
+    {
```
alsala, line 1934: It is not clear to me why coins->Clear(); has been removed

src/undo.h
```diff
@@ -97,6 +97,8 @@ class CTxUndo
     uint256 refTx;            //hash of coins from ceased sidechains. It's not needed for ordinary coins and certs
     unsigned int firstBwtPos; //position of the first bwt.
 
+    CTxUndo(): vprevout(), refTx(), firstBwtPos(0) {};
```
alsala: Is 0 a ggo default value for firstBwtPos? What if we have no bwt and no change?

src/validationinterface.cpp
```diff
@@ -36,10 +37,12 @@ void UnregisterValidationInterface(CValidationInterface* pwalletIn) {
     g_signals.EraseTransaction.disconnect(boost::bind(&CValidationInterface::EraseFromWallet, pwalletIn, _1));
     g_signals.SyncTransaction.disconnect(boost::bind(&CValidationInterface::SyncTransaction, pwalletIn, _1, _2));
     g_signals.UpdatedBlockTip.disconnect(boost::bind(&CValidationInterface::UpdatedBlockTip, pwalletIn, _1));
+    g_signals.SyncBwtCeasing.disconnect(boost::bind(&CValidationInterface::SyncBwtCeasing, pwalletIn, _1, _2));
```
alsala: For symmetry move this at the beginning; the same for line 45 below

src/wallet/wallet.cpp
```diff
@@ -1572,38 +1572,49 @@ void CWallet::SyncTransaction(const CTransaction& tx, const CBlock* pblock)
...
 }
 
-void CWallet::SyncCertificate(const CScCertificate& cert, const CBlock* pblock)
+void CWallet::SyncCertificate(const CScCertificate& cert, const CBlock* pblock, int bwtMaturityHeight)
 {
     LOCK2(cs_main, cs_wallet);
     if (!AddToWalletIfInvolvingMe(cert, pblock, true))
         return; // Not one of ours
 
     MarkAffectedTransactionsDirty(cert);
```
alsala: Is it correct to mark it dirty here instead of doing at the end of the method?

src/wallet/wallet.cpp
```diff
@@ -1572,38 +1572,49 @@ void CWallet::SyncTransaction(const CTransaction& tx, const CBlock* pblock)
...
+    std::map<uint256, std::shared_ptr<CWalletTransactionBase>>::iterator itCert = mapWallet.find(cert.GetHash());
+    if (itCert != mapWallet.end())
```
alsala: If cert is not found that should be marked as an error

src/wallet/wallet.h
```diff
@@ -651,6 +608,8 @@ class CWalletCert : public CScCertificate, public CWalletTransactionBase
         READWRITE(hashBlock);
         READWRITE(vMerkleBranch);
         READWRITE(nIndex);
+        READWRITE(bwtMaturityDepth);
+        READWRITE(areBwtCeased);
         std::vector<CScCertificate> vUnused; //! Used to be vtxPrev
```
alsala, line 613: Unused fields could be probably removed from this cert obj because it is a brand new data type in wallet db

src/gtest/test_sidechain_ceased.cpp
```diff
@@ -1002,3 +1002,189 @@ TEST_F(CeasedSidechainsTestSuite, UndoEmptyCertUpdatesToCeasingScs) {
...
     EXPECT_TRUE(updatedCeasingScIds.ceasingScs.count(scId) != 0);
 }
+///////////////////////////////////////////////////////////////////////////////
+//////////////////////////////// ApplyTxInUndo ////////////////////////////////
+///////////////////////////////////////////////////////////////////////////////
+TEST_F(CeasedSidechainsTestSuite, SpendChangeOutput_CoinReconstructionFromBlockUndo)
+{
+    //Create sidechain
+    uint256 scId = uint256S("aaa");
+    int dummyHeight = 100;
```
alsala: setting to constant values could be made via 'static const int'

alsala (changes requested):

src/gtest/test_sidechain_ceased.cpp
abi87: Done, indeed the only bwt is spent

src/wallet/wallet.h
abi87: Agree, since CWalletCert is stored under different key than CWalletTx

src/gtest/test_sidechain_ceased.cpp
abi87: agree done

src/wallet/wallet.cpp
abi87: Indeed MarkAffectedTransactionsDirty only reset the flag related to credit/debit calculation and does not affect `bwtMaturityHeight `nor `areBwtCeased`.
However for readability I'll move `MarkAffectedTransactionsDirty` to the bottom of the method as for `CWallet::SyncTransaction` and `CWallet::SyncBwtCeasing`

src/wallet/wallet.cpp
abi87: I think it should never happen, and I should probably assert cert presence and nature (i.e. IsCertificate) or throw. As of now, I'd go for the first

src/validationinterface.cpp
abi87: Done

src/gtest/test_sidechain_ceased.cpp
abi87: Done

src/undo.h
abi87: Current code actually fills the CTxUndo::firstBwtPos only for certificates which have bwt. Certificates with no bwts do not create a CTxUndo.
A way of structuring this code which is more coherent with the rest of the codebase is to type firstBwtPos as int and use -1 as default value

abi87: In case there are no bwt, the CTxUndos for the ceased bwts are not generated at all in HandleCeasingScs nor restored in RevertCeasingScs
However, for clarity sake, I'll set -1 as default value for firstBwtPos (changing it to int) and skip the relevant processing on RevertCeasingScs should firstBwtPos  == -1

src/wallet/wallet.cpp
abi87: Your comment and my previous one are wrong.
The reason is that a certificate will appear in a mapWallet of a specific node only if it is originated from a directed to one of the addresses in the wallet.
There is nothing to be done here if certificate is not found. Simply return

PR #284, opened 2020-06-05
Zendoo-mc-cryptolib integration + MTF and python tests

95DDB:


depends/packages/librustzcash.mk
```diff
@@ -1,10 +1,14 @@
 package=librustzcash
 $(package)_version=0.1
-$(package)_download_path=https://github.com/zcash/$(package)/archive/
-$(package)_file_name=$(package)-$($(package)_git_commit).tar.gz
+#$(package)_download_path=https://github.com/zcash/$(package)/archive/
+#$(package)_file_name=$(package)-$($(package)_git_commit).tar.gz
+$(package)_download_path=https://github.com/ark2038/librustzcash/archive/
```
albertog78, line 5: instead of introducing another package on another repo, apply patch after download @ark2038. If this is not possible quickly, move the project into our repo @cronicc.

albertog78 (changes requested):

depends/packages/librustzcash.mk
cronicc: Tracking dependencies in https://github.com/ZencashOfficial/zen/issues/295

PR #285, opened 2020-06-05
As/pr 09

alsala:
scid can not be specified when creating a Sidechain, it is generated using the hash of the creating tx and the index of the relevant creation crosschain output

### DIFF
--- a/src/primitives/transaction.cpp
+++ b/src/primitives/transaction.cpp
@@ -281,16 +281,12 @@ std::string CTxScCreationOut::ToString() const
 
 void CTxScCreationOut::GenerateScId(const uint256& txHash, unsigned int pos) const
 {
-    // TODO with respect to the comments of CR this could be added as well
-    const uint256& ccoutHash = GetHash();
-
     const uint256& scid = Hash(
             BEGIN(txHash),    END(txHash),
-            BEGIN(ccoutHash), END(ccoutHash),
             BEGIN(pos),       END(pos) );
 
-    LogPrint("sc", "%s():%d - updating scid=%s - tx[%s], ccout[%s], pos[%u]\n",
-        __func__, __LINE__, scid.ToString(), txHash.ToString(), ccoutHash.ToString(), pos);
+    LogPrint("sc", "%s():%d - updating scid=%s - tx[%s], pos[%u]\n",
+        __func__, __LINE__, scid.ToString(), txHash.ToString(), pos);
 
     *const_cast<uint256*>(&generatedScId) = scid;
 }
--- a/src/primitives/transaction.cpp
+++ b/src/primitives/transaction.cpp
@@ -263,9 +263,9 @@ bool CTxCrosschainOut::CheckAmountRange(CAmount& cumulatedAmount) const
 }
 
 CTxScCreationOut::CTxScCreationOut(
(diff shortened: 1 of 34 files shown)


src/primitives/transaction.cpp
```diff
@@ -275,8 +275,28 @@ uint256 CTxScCreationOut::GetHash() const
 
 std::string CTxScCreationOut::ToString() const
 {
-    return strprintf("CTxScCreationOut(scId=%s, withdrawalEpochLength=%d, nValue=%d.%08d, address=%s, customData=[%s]",
-        scId.ToString(), withdrawalEpochLength, nValue / COIN, nValue % COIN, HexStr(address).substr(0, 30), HexStr(customData) );
+    return strprintf("CTxScCreationOut(withdrawalEpochLength=%d, nValue=%d.%08d, address=%s, customData=[%s]",
```
abi87: Why not keeping the generated scId? wouldn't be useful for debugging?

src/primitives/transaction.cpp
```diff
@@ -577,6 +600,24 @@ double CTransactionBase::ComputePriority(double dPriorityInputs, unsigned int nT
     return dPriorityInputs / nTxSize;
 }
 
+const uint256& CTransaction::GetScIdFromScCcOut(int pos) const
+{
+    static const uint256 nullHash;
+    if (pos >= GetVscCcOut().size())
```
abi87: Since pos is an int, should you check that pos >= 0 ?

src/primitives/transaction.h
```diff
@@ -443,24 +443,16 @@ class CTxOut
...
     uint256 address;
 
-    CTxCrosschainOut(const uint256& scIdIn, const CAmount& nValueIn, const uint256& addressIn)
-        : scId(scIdIn), nValue(nValueIn), address(addressIn) { }
+    CTxCrosschainOut(const CAmount& nValueIn, const uint256& addressIn)
+        : nValue(nValueIn), address(addressIn) { }
 
     virtual ~CTxCrosschainOut() {};
 
     CTxCrosschainOut():nValue(-1) {}
```
abi87: Maybe here call explicitly address() in the initialization list

src/primitives/transaction.h
```diff
@@ -527,47 +520,40 @@ class CTxForwardTransferOut : public CTxCrosschainOut
 
 class CTxScCreationOut : public CTxCrosschainOut
 {
+friend class CTransaction;
```
abi87, line 525: It look like you need this friend declaration because you need to call `GenerateScId` when you call `UpdateHash `on `CTransaction`.
Why instead of storing the `generatedScId` into the `CTxScCreationOut ` don't we store it into a vector into CTransaction? After all `GetScId()` looks called only from the corresponding method in CTransation

src/primitives/transaction.cpp
abi87: Done

src/primitives/transaction.cpp
abi87: Done

src/primitives/transaction.h
abi87: Done

src/primitives/transaction.h
abi87: As agree AKF, let's postpone it to the final review to tackle it along with other CTx/CScCert related changes

PR #287, opened 2020-06-06
Beta preview

i-Alex:


src/coins.cpp
```diff
@@ -101,11 +111,31 @@ bool CCoins::IsFromCert() const {
     return (nVersion & 0x7f) == (SC_CERT_VERSION & 0x7f);
 }
 
+int CCoins::GetMaturityHeightForOutput(unsigned int outPos) const
```
albertog78: change to isOutputMature(unsigned int outPos, int spendHeight) return immediately true if coin is from regular tx (non coinbase) or if output is the change of a certificate. In all the other cases (cert bwt/coinbase) check the height

src/main.cpp
```diff
@@ -2150,14 +2095,28 @@ bool CheckTxInputs(const CTransactionBase& txBase, CValidationState& state, cons
         const CCoins *coins = inputs.AccessCoins(prevout.hash);
         assert(coins);
 
-        if (coins->IsCoinBase()) {
-            // Ensure that coinbases are matured
-            if (nSpendHeight - coins->nHeight < COINBASE_MATURITY) {
+        // Ensure that coinbases and certificates outputs are matured
+        int maturityHeight = coins->GetMaturityHeightForOutput(in.prevout.n);
```
albertog78: use the new method isOutputMature don't paying attention to MEMPOOL_HEIGHT etc ..

src/coins.cpp
```diff
@@ -1304,13 +1340,15 @@ bool CCoinsViewCache::HandleCeasingScs(int height, CBlockUndo& blockUndo)
...
-            undo.originScId = coins->originScId;
+            undo.nHeight            = coins->nHeight;
+            undo.fCoinBase          = coins->fCoinBase;
+            undo.nVersion           = coins->nVersion;
+            undo.nBwtMaturityHeight = coins->nBwtMaturityHeight;
 
             coins->Spend(pos);
         }
+
+        SyncBwtCeasing(scInfo.lastCertificateHash, true);
```
albertog78: This would signal also ceasings that will not be committed to the current state! This must be invoked only in block connect / disconnect as for synctransaction. Add an output parameter list that will be filled withe the ceased sidechains. Rename to SyncVoidedCertificate

src/coins.cpp
```diff
@@ -1332,32 +1370,37 @@ bool CCoinsViewCache::RevertCeasingScs(const CTxUndo& ceasedCertUndo)
...
-        coins->vout.at(firstBwtPos + bwtOutPos) = outVec.at(bwtOutPos).txout;
+            if(coins->IsAvailable(firstBwtPos + bwtOutPos))
+                fClean = fClean && error("%s: undo data overwriting existing output", __func__);
+            if (coins->vout.size() < (firstBwtPos + bwtOutPos+1))
+                coins->vout.resize(firstBwtPos + bwtOutPos+1);
+            coins->vout.at(firstBwtPos + bwtOutPos) = outVec.at(bwtOutPos).txout;
+        }
     }
 
+    SyncBwtCeasing(ceasedCertUndo.refTx, false);
```
albertog78: This would signal also ceasing reverts that will not be committed to the current state! This must be invoked only in block connect / disconnect as for synctransaction. Add an output parameter list that will be filled withe the unceased sidechains.

albertog78 (changes requested):

src/coins.cpp
```diff
@@ -1111,130 +1034,218 @@ bool CCoinsViewCache::RevertCertOutputs(const CScCertificate& cert)
...
+    CSidechainEventsMap::const_iterator it = FetchSidechainEvents(height);
+    if (it != cacheSidechainEvents.end() && it->second.flag != CSidechainEventsCacheEntry::Flags::ERASED) {
+        scEvents = it->second.scEvents;
         return true;
     }
     return false;
 }
 
-bool CCoinsViewCache::UpdateCeasingScs(const CTxScCreationOut& scCreationOut)
+bool CCoinsViewCache::ScheduleSidechainEvent(const CTxScCreationOut& scCreationOut, int creationHeight)
```
albertog78, line 1056: schedule events directly in UpdateSCInfo considering that maturity is also handled there

src/coins.cpp
```diff
@@ -1111,130 +1034,218 @@ bool CCoinsViewCache::RevertCertOutputs(const CScCertificate& cert)
...
+    }
+
+    LogPrint("sc", "%s():%d - SIDECHAIN-EVENT: scId[%s]: scCreation next ceasing height [%d]\n",
+            __func__, __LINE__, scCreationOut.GetScId().ToString(), nextCeasingHeight);
 
     return true;
 }
 
-bool CCoinsViewCache::UndoCeasingScs(const CTxScCreationOut& scCreationOut)
+bool CCoinsViewCache::ScheduleSidechainEvent(const CTxForwardTransferOut& forwardOut, int fwdHeight)
```
albertog78, line 1097: schedule events directly in UpdateSCInfo considering that maturity is also handled there

src/coins.cpp
```diff
@@ -1111,130 +1034,217 @@ bool CCoinsViewCache::RevertCertOutputs(const CScCertificate& cert)
...
-    LogPrint("cert", "%s():%d - CEASING HEIGHTS: scId[%s]: undo of creation removes currentCeasingHeight [%d]\n",
-            __func__, __LINE__, scCreationOut.scId.ToString(), currentCeasingHeight);
+    LogPrint("sc", "%s():%d - SIDECHAIN-EVENT: scId[%s]: fwd Transfer next maturing height [%d]\n",
+             __func__, __LINE__, forwardOut.scId.ToString(), maturityHeight);
 
     return true;
 }
 
-bool CCoinsViewCache::UpdateCeasingScs(const CScCertificate& cert)
+bool CCoinsViewCache::ScheduleSidechainEvent(const CScCertificate& cert)
```
albertog78, line 1124: move to UpdateScInfo

src/coins.cpp
```diff
@@ -1111,130 +1034,217 @@ bool CCoinsViewCache::RevertCertOutputs(const CScCertificate& cert)
...
+        cacheSidechainEvents[maturityHeight].scEvents.maturingScs.erase(scCreationOut.GetScId());
+        if (!cacheSidechainEvents[maturityHeight].scEvents.IsNull()) //still other sc ceasing at that height or fwds
+            cacheSidechainEvents[maturityHeight].flag = CSidechainEventsCacheEntry::Flags::DIRTY;
+        else
+            cacheSidechainEvents[maturityHeight].flag = CSidechainEventsCacheEntry::Flags::ERASED;
+
+        LogPrint("sc", "%s():%d - SIDECHAIN-EVENT: scId[%s] deleted maturing height [%d] for creation amount.\n",
+            __func__, __LINE__, scCreationOut.GetScId().ToString(), maturityHeight);
+    } else
+        LogPrint("sc", "%s():%d - SIDECHAIN-EVENT: scId[%s] nothing to do for scCreation amount maturing canceling at height [%d].\n",
```
albertog78, line 1193: here is not returning false because in the same block we could have had a ft that is reverted before sc creation (then removing the entry in this list). This can be managed better if we move the amount of maturing balance from scInfo to the cacheSidechainEvents maturingScs structure. In such a way we would have been able to detect if the mature balance reached zero after having removed the sc creation amount. Moreover in that way we will be able to not keep in two separated structures similar data.

src/sc/sidechain.h
```diff
@@ -63,11 +71,12 @@ class CSidechain {
     // value = the immature amount
     std::map<int, CAmount> mImmatureAmounts;
```
albertog78, line 72: Move immatureAmounts in cacheSidechainEvents scEvents maturingScs

src/coins.cpp
```diff
@@ -1303,117 +1333,159 @@ bool CCoinsViewCache::HandleCeasingScs(int height, CBlockUndo& blockUndo)
...
         }
-    }
 
-    LogPrint("sc", "%s():%d Exiting: CBlockUndo: %s\n",
-        __func__, __LINE__, blockUndo.ToString());
+        assert(pVoidedCertsList != nullptr);
+        pVoidedCertsList->push_back(scInfo.lastCertificateHash);
+    }
 
     return true;
```
albertog78, line 1370: we must mark as erased the scEvents entry inside the cache otherwise this list will grow forever. The entry must be restored in case of disconnect block.

src/coins.cpp
```diff
@@ -1303,117 +1333,159 @@ bool CCoinsViewCache::HandleCeasingScs(int height, CBlockUndo& blockUndo)
...
-        coins->vout.at(firstBwtPos + bwtOutPos) = outVec.at(bwtOutPos).txout;
+        assert(pVoidedCertsList != nullptr);
+        pVoidedCertsList->push_back(voidedCertUndo.voidedCertHash);
+
+        if (!fClean) return false;
     }
 
-    LogPrint("cert", "%s():%d - POST:%s\n", __func__, __LINE__, coins->ToString());
-    return fClean;
+    return true;
```
albertog78, line 1477: we must restore the scEvents entry inside the cache with the reverted ceased sidechains and the reverted matured amounts

src/core_memusage.h
```diff
@@ -25,58 +25,42 @@ static inline size_t RecursiveDynamicUsage(const CTxOut& out) {
     return RecursiveDynamicUsage(out.scriptPubKey);
 }
 
+static inline size_t RecursiveDynamicUsage(const CTxScCreationOut& ccout)
+{
+    return memusage::DynamicUsage(ccout.customData);
```
albertog78, line 30: update with other fields

albertog78 (changes requested):

src/coins.cpp
```diff
@@ -1446,12 +1460,16 @@ bool CCoinsViewCache::RevertSidechainEvents(const CBlockUndo& blockUndo, int hei
...
-        pVoidedCertsList->push_back(voidedCertUndo.voidedCertHash);
+        if(pVoidedCertsList != nullptr);
+            pVoidedCertsList->push_back(coinHash);
 
         if (!fClean) return false;
+
+        recreatedScEvent.ceasingScs.insert(voidedCertUndo.voidedCertScId);
     }
 
+    cacheSidechainEvents[height].scEvents = recreatedScEvent;
```
albertog78: if not null add entry

src/main.h
```diff
@@ -285,11 +286,14 @@ void FlushStateToDisk();
 void PruneAndFlush();
 
 /** (try to) add transaction to memory pool **/
-bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransaction &tx, bool fLimitFree,
-                        bool* pfMissingInputs, bool fRejectAbsurdFee=false);
+bool AcceptTxBaseToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransactionBase &txBase, bool fLimitFree,
```
albertog78, line 289: to avoid possible non clear behaviors don't set a default value to disconnecting

src/primitives/certificate.cpp
```diff
@@ -193,82 +217,57 @@ bool CScCertificate::CheckFinal(int flags) const
...
+    return nValueOut;
 }
 
-
 // Mutable Certificate
 //-------------------------------------
 CMutableScCertificate::CMutableScCertificate() :
     scId(), epochNumber(CScCertificate::EPOCH_NULL), quality(CScCertificate::QUALITY_NULL),
-    endEpochBlockHash(), scProof() { }
+    endEpochBlockHash(), scProof(), nFirstBwtPos(BWT_POS_UNSET) { }
```
albertog78, line 266: each method that modifies vouts (except resizeOut) set nFirstBwtPos to 0 if was set to BWT_POS_UNSET. It would be better to directly initialize nFirstBwtPos to 0 in constructor (where we are sure there are not bwt nor regular outputs). Alternatively following the current approach we have to check for initialization also in resizeOut (currently accidentally working because BWT_POS_UNSET is -1).

src/primitives/certificate.cpp
```diff
@@ -280,6 +279,49 @@ uint256 CMutableScCertificate::GetHash() const
...
+    if (pos < nFirstBwtPos)
+        --(*const_cast<int*>(&nFirstBwtPos));
+
+    vout.erase(vout.begin() + pos);
+}
+void CMutableScCertificate::resizeOut(unsigned int newSize) {
+    if (newSize > nFirstBwtPos)
+        (*const_cast<int*>(&nFirstBwtPos)) = newSize;
+
+    vout.resize(newSize);
```
albertog78, line 304: What is the expected behavior if this method is called when there are already some bwts? 
In current implementation if we have for example already 2 regular outputs and 2 bwt and then call resizeOut(3) it sets the size of vout to 3 (deleting the last bwt) but also updating nFirstBwtPos to 3 keeping bwt at vout[2] that will be interpreted as regular output. A more consistent strategy would be to enforce there are no bwts if resizeOut is called or an alternative strategy would be to move bwts (and obviously update nFirstBwtPos) accordingly to new regular outs size

src/primitives/certificate.cpp
```diff
@@ -15,83 +15,104 @@
...
     return true;
 }
 
-bool CScCertificate::CheckInputsAvailability(CValidationState &state) const
+bool CScCertificate::IsVersionStandard(int nHeight) const
 {
-    // there might be no inputs if 0 fee, therefore this never fails
-    return true;
-}
+    if (!zen::ForkManager::getInstance().areSidechainsSupported(nHeight))
```
albertog78, line 81: [Copied from previous PR]
return always true. No non-standard versions are allowed (IsValidVersion will restrict the only valid version that is standard)

src/primitives/certificate.cpp
```diff
@@ -15,83 +15,104 @@
...
+        if (pos >= nFirstBwtPos && txout.nValue == 0)
+            return state.DoS(100, error("CheckAmounts(): backward transfer has zero amount"),
+                             REJECT_INVALID, "bad-txns-bwd-vout-zero");
+        nCumulatedValueOut += txout.nValue;
+        if (!MoneyRange(nCumulatedValueOut))
+            return state.DoS(100, error("CheckAmounts(): txout total out of range"),
+                             REJECT_INVALID, "bad-txns-txouttotal-toolarge");
+    }
+
+    if (!MoneyRange(GetValueOfBackwardTransfers()))
```
albertog78, line 110: How can be possible that if nCumulatedValueOut is in range (that includes both change and bwts), GetValueOfBackwardTransfers() is not in range?

albertog78 (changes requested):

src/script/interpreter.cpp
```diff
@@ -1199,18 +1199,14 @@ class CScCertificateSignatureSerializer {
         std::vector<CTxOut> vout_ser;
 
         // reading from memory and writing to data stream
-        for (auto it = certTo.GetVout().begin(); it != certTo.GetVout().end(); ++it)
+        for(int pos = 0; pos < certTo.GetVout().size(); ++pos)
```
albertog78: CScCertificateSignatureSerializer does not serialize properly inputs taking in consideration selected nIn. Moreover there is no test verifying the correct behavior of certificate's signature! 
Remove CScCertificateSignatureSerializer class and "adapt" CTransactionSignatureSerializer to manage CTransactionBase and just differentiate the behavior in the Serialize method to keep the previous logic for tx and instead for certs serialize as in CCertficate  (only differences should be LockTime, Joinsplits, Bwts and certs base data like scId, epochNumber, quality, endEpochBlockHash, ..).
Introduce a test passing a Certificate instance to CTransactionSignatureSerializer.

albertog78 (changes requested):

src/test/sighash_tests.cpp
```diff
@@ -121,8 +121,8 @@ void static RandomTransaction(CMutableTransaction &tx, bool fSingle, bool emptyI
         txin.nSequence = (insecure_rand() % 2) ? insecure_rand() : (unsigned int)-1;
     }
     for (int out = 0; out < outs; out++) {
-        tx.vout.push_back(CTxOut());
-        CTxOut &txout = tx.vout.back();
+        tx.addOut(CTxOut());
```
albertog78, line 225: not related to this line:
add certificate signature hash test in sighash_test: create a modified version of SignatureHashOld for certificates and verify if new SignatureHash for certs is producing the same hash

albertog78 (changes requested):

src/txdb.cpp
```diff
@@ -315,17 +315,8 @@ bool CCoinsViewDB::GetStats(CCoinsStats &stats) const {
...
-                    for(const CTxOut& out: coins.vout) {
-                        if (!out.isFromBackwardTransfer)
-                            ++changeOutputCounter;
-                        else
-                            break;
-                    }
-
-                    ss << changeOutputCounter;
+                    ss << coins.nBwtMaturityHeight;
+                    ss << coins.nBwtMaturityHeight;;
```
albertog78, line 319: nFirstBwtPos and remove additional ;

src/txmempool.cpp
```diff
@@ -1004,7 +993,7 @@ bool CCoinsViewMemPool::GetSidechain(const uint256& scId, CSidechain& info) cons
         const uint256& scCreationHash = mempool.mapSidechains.at(scId).scCreationTxHash;
         const CTransaction & scCreationTx = mempool.mapTx.at(scCreationHash).GetTx();
         for (const auto& scCreation : scCreationTx.GetVscCcOut()) {
-            if (scId == scCreation.scId) {
+            if (scId == scCreation.GetScId()) {
```
albertog78, line 996: IMPORTANT: GetSidechain is NOT working properly if in the mempool there are two transactions creating two different scs. In particular if there is TX1 creating SC1 and TX2 creating SC2, GertSidechain(2) will return an empty SCInfo! Move the break at line 1005 inside the if.

albertog78 (changes requested):

src/txmempool.cpp
albertog78: Moreover at line 1023 why not updating also info.lastEpochReferencedByCertificate and info.lastCertificateHash? @abi87

albertog78 (changes requested):

src/txmempool.cpp
albertog78: > IMPORTANT: GetSidechain is NOT working properly if in the mempool there are two transactions creating two different scs. In particular if there is TX1 creating SC1 and TX2 creating SC2, GertSidechain(2) will return an empty SCInfo! Move the break at line 1005 inside the if.

SOLVED

abi87 (approved):

alsala (approved):

PR #291, opened 2020-06-06
Moved certUndo infos CTxUndo

abi87:
There were in scUndoData before.

### DIFF
--- a/src/undo.h
+++ b/src/undo.h
@@ -120,43 +120,71 @@ public:
 class CTxUndo
 {
 public:
-    // undo information for all txins
-    std::vector<CTxInUndo> vprevout;
-    CTxUndo(): vprevout() {};
+    std::vector<CTxInUndo> vprevout; // undo information for all txins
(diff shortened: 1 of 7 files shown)


src/undo.h
```diff
@@ -120,43 +120,71 @@ class CVoidedCertUndo {
...
 
 struct ScUndoData
 {
-    CAmount immAmount;
-    int certEpoch;
-    uint256 lastCertificateHash;
-    
-    ScUndoData(): immAmount(0), certEpoch(CScCertificate::EPOCH_NOT_INITIALIZED),
-                  lastCertificateHash() {}
+    CAmount appliedMaturedAmount;
```
alsala, line 175: This structure now has just a CAmount field, it can be probably removed and CAmount used directly in the map

alsala (approved):

src/undo.h
abi87: I agree this could be changed. I'd carry out this change with the ones needed to aggregated UpdateCeasingScs with ApplyMatureBalance

PR #293, opened 2020-06-08
Added nFirstBwtPos in CCoins

abi87:
The attribute nFirstBwtPos is used only for certificates, while it is set to -1 for coins from ordinary transactions.
Removal of IsFromBackwardTransfer for CTxOut is ongoing and will be subject of different pull request

### DIFF
--- a/src/coins.cpp
+++ b/src/coins.cpp
@@ -137,7 +137,7 @@ int CCoins::GetMaturityHeightForOutput(unsigned int outPos) const
     if(!IsAvailable(outPos))
         return -1; //This may happen in wallet when you check credit on certificate whose bwt has been erased
                    
-    if (vout.at(outPos).isFromBackwardTransfer)
+    if (outPos >= nFirstBwtPos)
         return nBwtMaturityHeight;
     else
         return nHeight;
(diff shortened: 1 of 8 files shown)


src/coins.cpp
```diff
@@ -51,13 +53,21 @@ void CCoins::From(const CScCertificate &cert, int nHeightIn, int bwtMaturityHeig
     nVersion           = cert.nVersion;
     nBwtMaturityHeight = bwtMaturityHeight;
     ClearUnspendable();
+    nFirstBwtPos = vout.size();
+    for(int idx = 0; idx < this->vout.size(); ++idx) {
```
alsala: idx here can be a size_t type

src/coins.h
```diff
@@ -97,6 +97,11 @@ class CCoins
     //! as new tx version will probably only be introduced at certain heights
     int nVersion;
 
+    //| If coins comes from a certificate, nFirstBwtPos represents the position of the first backward transfer;
+    //! All outputs after nFirstBwtPos, including nFirstBwtPos, are backward transfers
+    //! If coins comes from a tx, it is currently set to zero and not used
```
alsala: Fix the comment; moreover a static const in NOT_VALID_POS = -1 could be used

alsala (approved):

src/coins.cpp
abi87: Done

src/coins.h
abi87: Done

PR #299, opened 2020-06-10
As/pr 13

alsala:
Various sparse fixes from code review, mostly concerning txmempool and a few others

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -4670,23 +4670,15 @@ UniValue send_certificate(const UniValue& params, bool fHelp)
         throw JSONRPCError(RPC_INVALID_PARAMETER, string("invalid cert height"));
     }
 
-    // - there must not be another certificate for the same epoch (multiple certificates are not allowed)
-    // This also checks in mempool
+    // there must not be another certificate for the same scId in mempool (multiple certificates are not allowed)
     {
         LOCK(mempool.cs);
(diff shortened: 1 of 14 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -4670,23 +4670,15 @@ UniValue send_certificate(const UniValue& params, bool fHelp)
...
 
-    // - there must not be another certificate for the same epoch (multiple certificates are not allowed)
-    // This also checks in mempool
+    // there must not be another certificate for the same scId in mempool (multiple certificates are not allowed)
     {
         LOCK(mempool.cs);
-        CCoinsViewMemPool viewMemPool(pcoinsTip, mempool);
-
-        if (viewMemPool.HaveCertForEpoch(scId, epochNumber))
+        if ((mempool.mapSidechains.count(scId) != 0) && (!mempool.mapSidechains.at(scId).backwardCertificate.IsNull()))
```
abi87, line 4676: This is not equivalent to what `HaveCertForEpoch` was doing, since it does not check the blockchain (via view). Instead let's pull out the sidechain information and check whether certificate epoch is +1 the one in the infos

abi87 (changes requested):

src/wallet/rpcwallet.cpp
abi87: Indeed you check for blockchain state in previous call to `scView.isEpochDataValid`.
Comment solved

abi87 (approved):

PR #307, opened 2020-06-12
Added ZENDOO_MC env variable

95DDB:


qa/rpc-tests/test_framework/mc_test/mc_test.py
```diff
@@ -10,12 +10,12 @@
...
 
 
-def generate_params(datadir, srcdir, id):
+def generate_params(datadir, id):
     params_dir = _get_params_dir(datadir, id)
     if os.path.isfile(params_dir + "test_mc_pk") and os.path.isfile(params_dir + "test_mc_vk"):
         return
     args = []
-    args.append(os.path.join(srcdir, "zendoo/mcTest"))
+    args.append(os.getenv("ZENDOOMC", "mcTest"))
```
cronicc: Better to use os.path for any paths (handling of whitespace in leading directories, backslash on windows), and shouldn't the default still be `./src/zendoo/mcTest`?
e.g.
```suggestion
    args.append(os.path(os.getenv("ZENDOOMC", srcdir + "/zendoo/mcTest"))
```

qa/rpc-tests/test_framework/mc_test/mc_test.py
```diff
@@ -25,15 +25,15 @@ def generate_params(datadir, srcdir, id):
...
+    datadir, id, epoch_number, end_epoch_block_hash, prev_end_epoch_block_hash,
     quality, constant, pks, amounts):
 
     params_dir = _get_params_dir(datadir, id)
     if not os.path.isfile(params_dir + "test_mc_pk") or not os.path.isfile(params_dir + "test_mc_vk"):
         return
     proof_path = "{}epoch_{}_wcert_proof".format(_get_proofs_dir(datadir, id), epoch_number)
     args = []
-    args.append(os.path.join(srcdir, "zendoo/mcTest"))
+    args.append(os.getenv("ZENDOOMC", "mcTest"))
```
cronicc: Same as above

qa/rpc-tests/test_framework/mc_test/mc_test.py
95DDB: I'm not sure about this, I just followed the same logic that was present in the code base (e.g. in calls to zend and zen-cli in utils.py and authproxy.py in the test framework)

95DDB: Also, the intention was also to eliminate the need to pass "srcdir" for each call of this function

cronicc (approved):

albertog78 (approved):

PR #325, opened 2020-10-01
Mo/getblock verbose

MarcoOl94:


alsala (approved):

src/rpc/blockchain.cpp
```diff
@@ -398,12 +398,13 @@ UniValue getblock(const UniValue& params, bool fHelp)
...
-            "\nIf verbose is false, returns a string that is serialized, hex-encoded data for block 'hash|height'.\n"
-            "If verbose is true, returns an Object with information about block <hash|height>.\n"
+            "\nIf verbosity is 0, returns a string that is serialized, hex-encoded data for the block.\n"
+            "If verbosity is 1, returns an Object with information about the block.\n"
+            "If verbosity is 2, returns an Object with information about the block and information about each transaction. \n"
             "\nArguments:\n"
             "1. \"hash|height\"     (string, required) The block hash or height\n"
-            "2. verbose           (boolean, optional, default=true) true for a json object, false for the hex encoded data\n"
-            "\nResult (for verbose = true):\n"
+             "2. verbosity              (numeric, optional, default=1) 0 for hex encoded data, 1 for a json object, and 2 for json object with transaction data\n"
```
abi87: While here verbosity is reported as numeric, later on you handle the case it is boolean, for backward compatibility I guess. Should we extend the help reporting that older boolean True is mapped to verbosity=1 and False to 0?

abi87 (changes requested):

albertog78 (approved):

PR #331, opened 2020-10-14
As/test halving

alsala:
Added regtest only option for setting halving period; added a new py test for verifying coinbase subsidy amounts and halving height crossing


alsala:
CLosed and reopened for merging indevelopment

PR #336, opened 2020-10-16
As/minimal encoding rp

alsala:
Added checks for minimal encoding height in rp and UT. Python test to be added

### DIFF
--- a/src/script/standard.cpp
+++ b/src/script/standard.cpp
@@ -19,8 +19,6 @@ using namespace zen;
 
 using namespace std;
 
-typedef vector<unsigned char> valtype;
-
 unsigned nMaxDatacarrierBytes = MAX_OP_RETURN_RELAY;
 
 CScriptID::CScriptID(const CScript& in) : uint160(Hash160(in.begin(), in.end())) {}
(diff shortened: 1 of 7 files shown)


src/script/standard.cpp
```diff
@@ -317,8 +320,30 @@ bool Solver(const CScript& scriptPubKey, txnouttype& typeRet, vector<vector<unsi
...
+                            __FILE__, __func__, __LINE__, HexStr(vchBlockHeight.begin(), vchBlockHeight.end()), e.what());
+                        break;
+                    }
+                    catch(...)
+                    {
+                        LogPrint("%s: %s():%d - unexpected exception\n", __FILE__, __func__, __LINE__);
+                        break;
+                    }
+
+                    if (!CheckMinimalPush(vchBlockHeight, hopcode))
```
albertog78, line 340: EvalScript seems to be checking minimal push only if opcode <= OP_PUSHDATA4
https://github.com/HorizenOfficial/zen/blob/2aaf4b3bedb6eedd6e8a7a45dbc15b95bf7172b0/src/script/interpreter.cpp#L298

albertog78 (changes requested):

src/script/standard.cpp
alsala: The opcode value is red in a different context here compared to the other parser, where the same logic is used only when we are checking specifically the RP height and using the associated opcode, stored during a previous parser step.

albertog78 (approved):

cronicc (approved):

PR #348, opened 2020-11-10
Improve private chains tracking

abi87:
Along @cronicc indications we are going to extend getchaintips with a --with-penalties=true/false paramenter to include, for each branch:

1. penalty-at-start of branch. A non-zero value indicates branch starts of penalized
2. penalty-at-tip of branch
3. blocks-to-mainchain, i.e. how many blocks on selected branch must be mined in order for it to become the mainchain

### DIFF
--- a/src/zen/delay.cpp
+++ b/src/zen/delay.cpp
@@ -16,8 +16,12 @@ int64_t GetBlockDelay(const CBlockIndex& newBlock, const CBlockIndex& prevBlock,
     if (prevBlock.nChainDelay > 0)
     {
     	// blocks on tip get penalty reduced by 1;
-    	// others get penalty increased proportionally to distance from tip
-    	return std::max<int>((activeChainHeight - newBlock.nHeight),-1);
+    	// other blocks get penalty increased proportionally to distance from tip
+    	int64_t blockDelay =  std::max<int64_t>((activeChainHeight - newBlock.nHeight),int64_t(-1));
+
+    	LogPrintf("calculated blockDelay %d for newBlockHeight %d (activeChainHeight: %d, prevBlockChainDelay: %d)!\n",
+    			blockDelay, activeChainHeight, newBlock.nHeight, prevBlock.nChainDelay);
+    	return blockDelay;
     }
 
     // Introduce penalty in case we receive a historic block.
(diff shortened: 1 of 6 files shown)


src/zen/delay.cpp
```diff
@@ -2,35 +2,31 @@
...
-            return (activeChainHeight - newBlock.nHeight);
-        // no delay detected.
-        } else {
-            return 0;
-        }
+    if (prevBlock.nChainDelay > 0)
+    {
+    	// blocks on tip get penalty reduced by 1;
+    	// others get penalty increased proportionally to distance from tip
+    	return std::max<int>((activeChainHeight - newBlock.nHeight),-1);
```
alsala: This is more compact than before but we have not the trace anymore; should we keep it?

alsala (approved):

cronicc (approved):

src/zen/delay.cpp
```diff
@@ -2,35 +2,35 @@
...
-            return (activeChainHeight - newBlock.nHeight);
-        // no delay detected.
-        } else {
-            return 0;
-        }
+    if (prevBlock.nChainDelay > 0)
+    {
+    	// blocks on tip get penalty reduced by 1;
+    	// other blocks get penalty increased proportionally to distance from tip
+    	int64_t blockDelay =  std::max<int64_t>((activeChainHeight - newBlock.nHeight),int64_t(-1));
```
albertog78: Is this code behaving the same way if for example activeChainHeight is MIN_INT64  and newBlock.height is MIN_INT64 +1? 
Why this PR needed this change?

abi87: Change is purely to improve readability, I can roll it back should there not be agreement on such improvement.
Moreover this function is called uniquely by `AddToBlockIndex`, with `activeChainHeight ` set to `chainActive.Height()`. `chainActive.Height()` is either positive, or -1 (at start, when no block has been added yet). In both cases, the value far away for int64_t boundaries, which would cause the refactored code to diverge from the original

abi87: Reverted change as agree offline

albertog78 (approved):

cronicc:
To be rebased on top of current development to make a new CI run pass, ready to be merged once CI passes.

cronicc:
Not merging until I had a chance to look at #366

albertog78 (approved):

Issue #347, opened 2020-11-09
Reindex performance issue


albertog78:
Reindex operation seems really slow even compared to full download. We should verify if reindex is also performing the whole consistency checks even if the chain is synced after a checkpoint (during download such checks are not performed if a checkpoint was present). For backward compatibility, an option would be to introduce a new switch (eg. fast_reindex) that skips the "heavy" checks before checkpoints if already synced after that point (use same logic of download phase).

cronicc:
Closed with https://github.com/HorizenOfficial/zen/commit/e109c9ae9395048febb23d41420665371bfe927b

`-reindexfast` added addressing this issue

PR #350, opened 2020-11-23
Improve reindex

abi87:
Following on https://github.com/HorizenOfficial/zen/issues/347 I introduced a `-reindexfast` option which allows to skip some expensive checks for txes in blocks below checkpoints, upon loading files from disk.
The mechanism is exactly that following upon receiving blocks from the network, header first


src/init.cpp
```diff
@@ -366,7 +366,8 @@ std::string HelpMessage(HelpMessageMode mode)
             "Warning: Reverting this setting requires re-downloading the entire blockchain. "
             "(default: 0 = disable pruning blocks, >%u = target size in MiB to use for block files)"), MIN_DISK_SPACE_FOR_BLOCK_FILES / 1024 / 1024));
     strUsage += HelpMessageOpt("-reindex", _("Rebuild block chain index from current blk000??.dat files on startup"));
-#if !defined(WIN32)
+    strUsage += HelpMessageOpt("-reindexfast", _("Rebuild block chain index from current blk000??.dat files on startup, skipping expensive checks for blocks below checkpoints"));
```
alsala: Should they be mutually exclusive? In that case specify it in help too

src/init.cpp
```diff
@@ -1393,6 +1416,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     // ********************************************************* Step 7: load block chain
 
     fReindex = GetBoolArg("-reindex", false);
+    fReindexFast = GetBoolArg("-reindexfast", false);
```
alsala, line 1424: Should they be mutually exclusive? In that case handle the clash if any

src/init.cpp
```diff
@@ -1727,7 +1752,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     if (fPruneMode) {
         LogPrintf("Unsetting NODE_NETWORK on prune mode\n");
         nLocalServices &= ~NODE_NETWORK;
-        if (!fReindex) {
+        if (!fReindex || !fReindexFast) {
```
alsala: Shouldn't it be &&?

src/init.cpp
abi87: Indeed `-reindexfast` absorbs `-reindex`, so if you specify both reindex fast is performed.
However this behaviour should not be allowed, so I am going to forbid specifying them both

src/init.cpp
abi87: agree, introduce check for mutual exclusivity

src/init.cpp
abi87: agree, fixed

src/gtest/test_reindex.cpp
```diff
@@ -0,0 +1,604 @@
...
+private:
+    CBlockHeader createCoinBaseOnlyBlockHeader(const uint256& prevBlockHash);
+
+    boost::filesystem::path  dataDirLocation;
+    const unsigned int       chainStateDbSize;
+    CFakeCoinDB*             pChainStateDb;
+    CCriticalBlock           csMainLock;        /**< Critical section needed when compiled with --enable-debug, which activates ASSERT_HELD */
+};
+
+TEST_F(ReindexTestSuite, LoadingBlocksFromNullFilePtrWillAbort) {
```
lander86: This test is not portable and makes the compilation on windows fail.

alsala (approved):

cronicc:
Please rebase on top of current development to get the CI tests to work.

I've so far completed the following tests successfully:
1. Sync from scratch on mainnet with a wallet containing PHGR, Groth16 and transparent tx history
2. -reindexfast with the same wallet

In both cases the private keys were imported at height 0, all transactions were detected as expected.

Currently running tests with this branch merged into AddressIndexing to make sure that -addressindex -timestampindex -spentindex are working as expected.

cronicc:
Manual test with AddressIndexing on mainnet worked, as well as a [CI run](https://travis-ci.com/github/HorizenOfficial/zen/builds/216646959), separate PR here https://github.com/HorizenOfficial/zen/pull/363

cronicc (approved):

albertog78 (approved):

PR #357, opened 2021-01-26
Improved the processing of blocks when zk-SNARKs verification is not needed (e.g. when using checkpoints).

ptagl:
- Skip cryptographic objects initialization for PHGR transactions when no verification is required.
- Skip verification of GROTH transactions when possible.


ptagl:
The same fix has been implemented on ZCash a few months ago with this [commit](https://github.com/zcash/zcash/commit/7fa2bb6005b5b408fc46a0242ba2cd686210148c).

PR #362, opened 2021-02-09
Modify contrib/linearize to work with zend

cronicc:


lander86 (approved):

elpuji:
I have to do thomething?

PR #367, opened 2021-03-02
Mo/websocket

MarcoOl94:


src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+    enum WsEventType {
+        UPDATE_TIP = 0,
+        UPDATE_MEMPOOL = 1,
+		UPDATE_PEER = 2,
+        EVT_UNDEFINED = 0xff
+    };
+    enum WsRequestType {
+        GET_SINGLE_BLOCK = 0,
+        GET_MULTIPLE_BLOCK_HASHES = 1,
+        GET_NEW_BLOCK_HASHES = 2,
```
alsala, line 122: add a comment for reserving the value, for instance RESERVED=3; here is properly unused for avoiding conflicts when merging with zendoo

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+        rspPayload.push_back(Pair("certificates", certHashes));
+
+        UniValue* rv = wse->getPayload();
+        if (!clientRequestId.empty())
+            rv->push_back(Pair("requestId", clientRequestId));
+        rv->push_back(Pair("responsePayload", rspPayload));
+        wsq.push(wse);
+    }
+
+    void sendMempoolEvent(int size, std::list<uint256>& listHashes, WsEvent::WsEventType eventType)
```
alsala: listHashes can be made const std::list<uint256>&

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+        rspPayload.push_back(Pair("transactions", txHashes));
+        rspPayload.push_back(Pair("certificates", certHashes));
+
+        UniValue* rv = wse->getPayload();
+        rv->push_back(Pair("eventType", eventType));
+        rv->push_back(Pair("eventPayload", rspPayload));
+        wsq.push(wse);
+    }
+
+    void sendPeerEvent(std::list<std::string> peers, WsEvent::WsEventType eventType)
```
alsala: peers should be a reference in order to avoid copying, and also const since it is not going to be modified. The same comment applies for similar lines below.

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+        UniValue* rv = wse->getPayload();
+        if (!clientRequestId.empty())
+            rv->push_back(Pair("requestId", clientRequestId));
+        rv->push_back(Pair("responsePayload", rspPayload));
+        wsq.push(wse);
+
+    }
+
+    void sendNodeConf(int version, int protocolVersion, int timeoffset, int relayfee,
+            std::string proxy, std::string errors,int blocks, int connections, bool testnet, double difficulty, double networkDifficulty, std::list<std::string> peers,
```
alsala: All strings parameters should be const reference to string

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+
+class WsNotificationInterface;
+class WsHandler;
+
+static int getblock(const CBlockIndex *pindex, std::string& blockHexStr);
+static void ws_updatetip(const CBlockIndex *pindex);
+static void ws_updatemempool();
+static void ws_updatepeer();
+
+double GetNetworkDifficulty(const CBlockIndex * pindex);
```
alsala: Add a prepending 'extern' keyword, it is redundant but is good practice to set it explicitly

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+
+
+        UniValue* rv = wse->getPayload();
+        rv->push_back(Pair("eventType", eventType));
+        rv->push_back(Pair("eventPayload", rspPayload));
+        wsq.push(wse);
+    }
+
+    void sendBlock(int height, const std::string& strHash, const std::string& blockHex, const CBlockIndex* pblockindex,
+            WsEvent::WsMsgType msgType, std::string clientRequestId = "")
```
alsala: Not modified in this PR, but clientRequestId can be made const std::string& even if it is a default-value parameter. The same comment applies for other occurrencies in following functions.

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+    void sendBlockEvent(int height, const std::string& strHash, const std::string& blockHex, WsEvent::WsEventType eventType, const CBlockIndex * pindex)
+    {
+        // Send a message to the client:  type = eventType
+        WsEvent* wse = new WsEvent(WsEvent::MSG_EVENT);
+        LogPrint("ws", "%s():%d - allocated %p\n", __func__, __LINE__, wse);
+        UniValue rspPayload(UniValue::VOBJ);
+        rspPayload.push_back(Pair("height", height));
+        rspPayload.push_back(Pair("hash", strHash));
+        rspPayload.push_back(Pair("block", blockHex));
+        rspPayload.push_back(Pair("chainwork",  pindex->nChainWork.GetHex()));
```
alsala: Consider adding an if (pindex) to be on the safe side (unless the comment proposing a modification to the interface of this function is accepted)

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+            WsEvent::WsMsgType msgType, std::string clientRequestId = "")
+    {
+        // Send a message to the client:  type = eventType
+        WsEvent* wse = new WsEvent(msgType);
+        LogPrint("ws", "%s():%d - allocated %p\n", __func__, __LINE__, wse);
+        UniValue rspPayload(UniValue::VOBJ);
+        rspPayload.push_back(Pair("height", height));
+        rspPayload.push_back(Pair("hash", strHash));
+        rspPayload.push_back(Pair("block", blockHex));
+        rspPayload.push_back(Pair("chainwork",  pblockindex->nChainWork.GetHex()));
```
alsala: Consider adding a if (pblockindex), even if the caller is currently checking this

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+        rspPayload.push_back(Pair("hashes", hashes));
+
+        UniValue* rv = wse->getPayload();
+        if (!clientRequestId.empty())
+            rv->push_back(Pair("requestId", clientRequestId));
+        rv->push_back(Pair("responsePayload", rspPayload));
+        wsq.push(wse);
+    }
+
+    void sendMempool(int size, std::list<uint256>& listHashes,
```
alsala: listHashes can be made const

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+    		WsEvent::WsMsgType msgType, std::string clientRequestId = "")
+    {
+        // Send a message to the client:  type = responseType
+        WsEvent* wse = new WsEvent(msgType);
+        LogPrint("ws", "%s():%d - allocated %p\n", __func__, __LINE__, wse);
+        UniValue rspPayload(UniValue::VOBJ);
+        rspPayload.push_back(Pair("size", size));
+
+        UniValue txHashes(UniValue::VARR);
+        UniValue certHashes(UniValue::VARR);
```
alsala: Probably we should remove certificates handling

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+    void sendMempoolEvent(int size, std::list<uint256>& listHashes, WsEvent::WsEventType eventType)
+    {
+        // Send a message to the client:  type = eventType
+        WsEvent* wse = new WsEvent(WsEvent::MSG_EVENT);
+        LogPrint("ws", "%s():%d - allocated %p\n", __func__, __LINE__, wse);
+        UniValue rspPayload(UniValue::VOBJ);
+        rspPayload.push_back(Pair("size", size));
+
+        UniValue txHashes(UniValue::VARR);
+        UniValue certHashes(UniValue::VARR);
```
alsala: Same as above, probably we should remove certificates handling

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+        }
+        rspPayload.push_back(Pair("peers", peerList));
+
+        UniValue* rv = wse->getPayload();
+        rv->push_back(Pair("eventType", eventType));
+        rv->push_back(Pair("eventPayload", rspPayload));
+        wsq.push(wse);
+    }
+
+    void sendTxs(std::list<std::string> listHex,
```
alsala: Same as above, const std::list&

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+        MSG_EVENT = 0,
+        MSG_REQUEST = 1,
+        MSG_RESPONSE = 2,
+        MSG_ERROR = 3,
+        MSG_UNDEFINED = 0xff
+    };
+
+    explicit WsEvent(WsMsgType xn): type(xn), payload(UniValue::VOBJ)
+    {
+        payload.push_back(Pair("msgType", type));
```
alsala, line 141: In zend all push_back(Pair("key", value)) calls have been replaced with pushKV("key", value). See https://github.com/HorizenOfficial/zen/commit/557c0513b8752bc1e0dbb377098f9cc60ffa7443#diff-63c7710d72381cffe692175dd70e47f1891bf75c3e09249ed826f9df3917d02d

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+    {
+        // Send a message to the client:  type = responseType
+        WsEvent* wse = new WsEvent(msgType);
+        LogPrint("ws", "%s():%d - allocated %p\n", __func__, __LINE__, wse);
+        UniValue rspPayload(UniValue::VOBJ);
+
+        UniValue txHex(UniValue::VARR);
+
+        for(std::list<std::string>::const_iterator it = listHex.begin(); it != listHex.end(); ++it) {
+        	UniValue value;
```
alsala: UniValue(const std::string&) ctor calls setStr, therefore consider rewriting this block as simply txHex.push_back(UniValue(*it)); or even txHex.push_back(*it); which implicitly converts to UniValue.

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+
+        UniValue* rv = wse->getPayload();
+        if (!clientRequestId.empty())
+            rv->push_back(Pair("requestId", clientRequestId));
+        rv->push_back(Pair("responsePayload", rspPayload));
+        wsq.push(wse);
+    }
+
+
+    int getHashByHeight(std::string height, std::string& strHash)
```
alsala: height is a const std::string&

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+        LOCK(mempool.cs);
+
+        std::list<std::string> txs;
+        for(const UniValue& hash : hashes.getValues()) {
+            if (hash.isObject())
+            {
+                LogPrint("ws", "%s():%d - invalid obj\n", __func__, __LINE__);
+                return INVALID_PARAMETER;
+            }
+            uint256 txid(uint256S(hash.get_str()));
```
alsala: get_str() throws an exception on failure, we had better testing the UniValue with isStr() or enclosing it is a try/catch and handling errors.

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+        return OK;
+
+    }
+
+    int sendMempoolTxs(const UniValue& hashes, const std::string& clientRequestId) {
+        LOCK(mempool.cs);
+
+        std::list<std::string> txs;
+        for(const UniValue& hash : hashes.getValues()) {
+            if (hash.isObject())
```
alsala: This check should be replaced with !isStr(). Consider that get_str() throws an exception if UniVal is not a string.

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+            auto entry = mempool.mapTx.find(txid);
+            if (entry != mempool.mapTx.end()) {
+                txs.push_back(EncodeHexTx(
+                		static_cast<CTransaction>(entry->second.GetTx())));
+            } else {
+                LogPrint("ws", "%s():%d - mempool transaction not found!\n", txid.ToString(), __LINE__);
+            }
+
+        }
+        sendTxs(txs, WsEvent::MSG_RESPONSE, clientRequestId);
```
alsala: In case we have no txs to send at this point, for whatever reason, have we to send the message anyway?

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+
+    int sendRawMempool(const std::string& clientRequestId) {
+        LOCK(mempool.cs);
+
+        std::list<uint256> hashes;
+        BOOST_FOREACH(const PAIRTYPE(uint256, CTxMemPoolEntry)& entry, mempool.mapTx)
+        {
+           hashes.push_back(entry.first);
+        }
+        sendMempool(hashes.size(), hashes, WsEvent::MSG_RESPONSE, clientRequestId);
```
alsala, line 598: In case we have no txs in mempool.mapTx, have we to send the message anyway?

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+        proxyType p;
+        GetProxy(NET_IPV4, p);
+        std::string proxy = p.IsValid() ? p.proxy.ToStringIPPort() : std::string();
+        std::string errors = GetWarnings("statusbar");
+        int blocks = (int)chainActive.Height();
+        int connections = (int)vNodes.size();
+        bool testnet = Params().TestnetToBeDeprecatedFieldRPC();
+        double difficulty = (double)GetDifficulty();
+        double networkDifficulty = (double)GetNetworkDifficulty();
+        LOCK(cs_vNodes);
```
alsala: LOCK(cs_vNodes) should be enclosed in a scope, to limit the size of the critical section to the minimum.
Moreover, also:
    int connections = (int)vNodes.size();
should be in the same critical section.

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+            LogPrint("ws", "%s():%d - invalid len %d (max is %d)\n", __func__, __LINE__, len, MAX_BLOCKS_REQUEST);
+            return INVALID_PARAMETER;
+        }
+
+        std::list<CBlockIndex*> listBlock;
+        CBlockIndex* pblockindexStart = NULL;
+        int lastH = 0;
+        for (const UniValue& o : hashes.getValues())
+        {
+            if (o.isObject())
```
alsala: This check should be replaced with !isStr(). Consider that get_str() below throws an exception if UniVal is not a string.

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+    int sendEstimateFee(std::string strBlock, const std::string& clientRequestId) {
+        int nBlocks = -1;
+        try {
+            nBlocks = std::stoi(strBlock);
+        } catch (const std::exception &e) {
+            LogPrint("ws", "%s():%d - %s\n", __func__, __LINE__, e.what());
+            return INVALID_PARAMETER;
+        }
+
+        if (nBlocks < 1) {
```
alsala, line 781: Why are we overwriting negative values with 1?

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+                    return MISSING_REQID;
+                }
+                const UniValue& reqPayload = find_value(request, "requestPayload");
+                if (reqPayload.isNull()) {
+                    LogPrint("ws", "%s():%d - requestPayload null: msg[%s]\n", __func__, __LINE__, msg);
+                    return INVALID_JSON_FORMAT;
+                }
+                const UniValue&  hashArray = find_value(reqPayload, "hash");
+                if (hashArray.isNull()) {
+                    LogPrint("ws", "%s():%d - locatorHash empty: msg[%s]\n", __func__, __LINE__, msg);
```
alsala: In the trace 'locatorHash' seems not the correct name

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+                }
+                const UniValue& reqPayload = find_value(request, "requestPayload");
+                if (reqPayload.isNull()) {
+                    LogPrint("ws", "%s():%d - requestPayload null: msg[%s]\n", __func__, __LINE__, msg);
+                    return INVALID_JSON_FORMAT;
+                }
+
+                const std::string& strBlock = findFieldValue("block", reqPayload);
+                if (strBlock.empty()) {
+                    LogPrint("ws", "%s():%d - limit empty: msg[%s]\n", __func__, __LINE__, msg);
```
alsala: In the trace 'limit' is not the correct name

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+        else
+        {
+            LogPrint("ws", "%s():%d - there are no connected ws clients\n", __func__, __LINE__);
+        }
+    }
+}
+
+static void ws_updatemempool()
+{
+    LOCK(mempool.cs);
```
alsala: The scope of the LOCK should be reduced just to include the loop on mempool.mapTx

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+		else
+		{
+			LogPrint("ws", "%s():%d - there are no connected ws clients\n", __func__, __LINE__);
+		}
+    }
+}
+
+static void ws_updatepeer()
+{
+    LOCK(cs_vNodes);
```
alsala: The scope of the LOCK should be reduced just to include the loop on vNodes

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+
+
+class WsHandler
+{
+private:
+    boost::shared_ptr< websocket::stream<tcp::socket>> localWs;
+    boost::lockfree::queue<WsEvent*, boost::lockfree::capacity<1024>> wsq;
+    std::atomic<bool> exit_rwhandler_thread_flag { false };
+
+    void sendBlockEvent(int height, const std::string& strHash, const std::string& blockHex, WsEvent::WsEventType eventType, const CBlockIndex * pindex)
```
alsala: Consider using a const std::string& chainWorkStr instead of passing the pindex.
This would imply also changing the interface of send_tip_update().
The reason is that we already are passing other parameters such as: pindex->nHeight, pindex->GetBlockHash()
therefore we might as well pass pindex->nChainWork.GetHex().

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+                }
+                else
+                {
+                    ++it;
+                }
+            }
+        }
+    }
+
+    void send_tip_update(int height, const std::string& strHash, const std::string& blockHex, const CBlockIndex * pindex)
```
alsala: Consider using a const std::string& chainWorkStr instead of passing the pindex.
As commented above for sendBlockEvent(), the reason is that we already are passing other parameters such as: pindex->nHeight, pindex->GetBlockHash()
therefore we might as well pass pindex->nChainWork.GetHex().

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+    {
+		std::unique_lock<std::mutex> lck(wsmtx);
+		if (listWsHandler.size() )
+		{
+			LogPrint("ws", "%s():%d - update mempool loop on ws clients\n", __func__, __LINE__);
+			auto it = listWsHandler.begin();
+			while (it != listWsHandler.end())
+			{
+				LogPrint("ws", "%s():%d - call wshandler_send_mempool_update to connection[%u]\n", __func__, __LINE__, (*it)->t_id);
+				(*it)->send_mempool_update(hashes.size(), hashes);
```
alsala: It is no use passing also the first parameters: it can be got from the list itself.

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+            ++it;
+        }
+    }
+}
+
+bool StartWsServer()
+{
+    try
+    {
+        // ip address is not configurable as of now, it is locahost mandatorily
```
alsala: Either change the comment or set it statically to localhost

src/validationinterface.cpp
```diff
@@ -34,6 +37,9 @@ void UnregisterValidationInterface(CValidationInterface* pwalletIn) {
     g_signals.EraseTransaction.disconnect(boost::bind(&CValidationInterface::EraseFromWallet, pwalletIn, _1));
     g_signals.SyncTransaction.disconnect(boost::bind(&CValidationInterface::SyncTransaction, pwalletIn, _1, _2));
     g_signals.UpdatedBlockTip.disconnect(boost::bind(&CValidationInterface::UpdatedBlockTip, pwalletIn, _1));
+    g_signals.MempoolChanged.disconnect(boost::bind(&CValidationInterface::MempoolChanged, pwalletIn));
```
alsala: The order of unregistering should be the opposite of the registering

alsala (changes requested):

src/zen/websocket_server.cpp
MarcoOl94: We need because the explorer compares the transactions received the websocket with the transactions ids sent and it deletes the missing transactions in the response.

src/zen/websocket_server.cpp
MarcoOl94: Yes because the explorer compares the transactions received the websocket with the transactions ids sent and it deletes the missing transactions in the response.

src/zen/websocket_server.cpp
MarcoOl94: Because nBlocks is the number of blocks within which the transaction would like to be confirmed, at least 1.

alsala (approved):

src/main.cpp
```diff
@@ -2758,6 +2758,7 @@ bool static DisconnectTip(CValidationState &state) {
     BOOST_FOREACH(const CTransaction &tx, block.vtx) {
         SyncWithWallets(tx, NULL);
     }
+    GetMainSignals().MempoolChanged();
```
albertog78, line 2763: Notify the same signal also on ConnectTip after having modified the mempool

src/net.cpp
```diff
@@ -513,6 +514,7 @@ CNode* ConnectNode(CAddress addrConnect, const char *pszDest)
         {
             LOCK(cs_vNodes);
             vNodes.push_back(pnode);
+            GetMainSignals().PeersChanged();
```
albertog78, line 517: To optimize performances process the events on the explorer side on a regular basis

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+        rspPayload.pushKV("block", blockHex);
+        rspPayload.pushKV("chainwork",  chainwork);
+
+        UniValue* rv = wse->getPayload();
+        rv->pushKV("eventType", eventType);
+        rv->pushKV("eventPayload", rspPayload);
+        wsq.push(wse);
+    }
+
+    void sendBlock(int height, const std::string& strHash, const std::string& blockHex, const CBlockIndex* pblockindex,
```
albertog78: Instead of passing pblockindex pass directly chainWork

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+    int sendNodeStaticConfiguration(const std::string& clientRequestId) {
+        int version = CLIENT_VERSION;
+        int protocolVersion = PROTOCOL_VERSION;
+        int timeoffset = 0;
+        int relayfee = minRelayTxFee.GetFeePerK();
+        proxyType p;
+        GetProxy(NET_IPV4, p);
+        std::string proxy = p.IsValid() ? p.proxy.ToStringIPPort() : std::string();
+        std::string errors = GetWarnings("statusbar");
+        int blocks = (int)chainActive.Height();
```
albertog78: verify to lock properly where is needed

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
+#include "../util.h"
```
albertog78, line 1: align implementation to last zendoo version

src/zen/websocket_server.cpp
```diff
@@ -0,0 +1,1360 @@
...
+        } catch (const std::exception &e) {
+            LogPrint("ws", "%s():%d - %s\n", __func__, __LINE__, e.what());
+            return INVALID_PARAMETER;
+        }
+
+        if (nBlocks < 1) {
+            nBlocks = 1;
+        }
+
+        CFeeRate feeRate = mempool.estimateFee(nBlocks);
```
albertog78, line 785: verify if lock is needed

albertog78 (changes requested):

src/zen/websocket_server.cpp
MarcoOl94: The RPC command where the implementation is taken doesn't use any lock. (https://github.com/HorizenOfficial/zen/blob/master/src/rpc/mining.cpp#L786). However this event is never called from the explorer, maybe i can delete it.

PR #426, opened 2021-12-05
Add merkleTree and scTxsCommitment as return values of getblocktemplate.

ruyichang:
Make merkleTree and scTxsCommitment as return values of getblockemplate is convenient for use coinbasetx and other txs to build block directly. This avoids calling getblockmerkleroots for a same job.


cronicc:
Superseded by https://github.com/HorizenOfficial/zen/pull/432

Thanks for this contribution though!

PR #429, opened 2021-12-30
Custom fields validation fix

ptagl:
This PR contains a fix for the custom fields management. In particular, the validation function was using the wrong bit endianness to check custom fields against the configuration (bit length) specified during sidechain creation.


src/util.cpp
```diff
@@ -934,21 +935,15 @@ int GetNumCores()
...
-        inputByte = (inputByte ^ (inputByte - 1)) >> 1;  // Set inputByte's trailing 0s to 1s and zero rest
-        for (c = 0; inputByte; c++)
-        {
-            inputByte >>= 1;
-        }
-    }
-    return c;
+    // __builtin_clz() returns the number of leading zeros for an unsigned int, so we need to perform a
+    // conversion to use it for an unsigned char.
+    return __builtin_clz(inputByte) % (sizeof(unsigned int) * 8 - CHAR_BIT);
```
i-Alex: Seems that `__builtin_clz` function is compiler and platform dependent and works with `unsigned int` type. Is it safe to use it? With trailing zeros before, we had safer solution, as for me.

src/gtest/test_libzendoo.cpp
```diff
@@ -1973,4 +1973,27 @@ TEST(CctpLibrary, TestInvalidProofVkWhenOversized)
     EXPECT_FALSE(vkInvalid.IsValid());
 
     //TODO: Might be useful to test the same behaviour with bit vector
+}
+
+TEST(CctpLibrary, TestCustomFieldsValidation)
+{
+    for (uint8_t i = 1; i < CHAR_BIT; i++)
```
i-Alex, line 2042: Here we have all possible cases for up-to 1 byte size custom field.
Let's add also some cases with bigger config size in a separate test. I mean, assuming that field element is created  from int\long\full field element. So actually the cases which failed before the fix.

ptagl: Increased unit tests coverage with commit [7267da211e5e0fad639ca9cea7d20a0c6658daf8](https://github.com/HorizenOfficial/zen/commit/7267da211e5e0fad639ca9cea7d20a0c6658daf8).

src/util.cpp
ptagl: Implemented a new function with commit [db1a66f2005b9348c04ee9b06c1aef8907622837](https://github.com/HorizenOfficial/zen/commit/db1a66f2005b9348c04ee9b06c1aef8907622837).

I moved `__builtin_clz()` to a unit test to check the validity of the new implemented function.

src/gtest/test_libzendoo.cpp
```diff
@@ -19,6 +19,15 @@
 
 using namespace blockchain_test_utils;
 
+// Random uint used to generate a random custom field
```
alsala: It would be useful add a note explaining why this FE is created in this way (see also TestRandomCustomFieldGeneration below)

src/gtest/test_libzendoo.cpp
```diff
@@ -1973,4 +1982,140 @@ TEST(CctpLibrary, TestInvalidProofVkWhenOversized)
...
+TEST(CctpLibrary, TestGetLeadingZeros)
+{
+    ASSERT_EQ(8, getLeadingZeroBitsInByte(0));
+
+    for (uint8_t n = 1; n > 0; n++)
+    {
+        ASSERT_EQ(__builtin_clz(n) % (sizeof(unsigned int) * 8 - CHAR_BIT), getLeadingZeroBitsInByte(n));
+    }
+}
+
```
alsala, line 2039: Could be useful also add a unit test for the getBytesFromBits() api in util.cpp

src/gtest/test_libzendoo.cpp
```diff
@@ -1973,4 +1982,140 @@ TEST(CctpLibrary, TestInvalidProofVkWhenOversized)
...
+            rawBytes.back() = j;
+            FieldElementCertificateField certField = FieldElementCertificateField(rawBytes);
+            FieldElementCertificateFieldConfig config = FieldElementCertificateFieldConfig(i + 31 * CHAR_BIT);
+            CFieldElement fe = certField.GetFieldElement(config);
+
+            // The Field Element is valid if it matches the configuration (j < 1 << i)
+            // and if doesn't exceed the modulus, thus the last two bits cannot be set (j < 1 << 6).
+            if (j < 1 << i && j < 1 << 6)
+            {
+                EXPECT_TRUE(fe.IsValid());
```
alsala, line 2084: This is correct, even if according to the implementation of GetFieldElement(config) the returned fe is always the null fe if the bits of cfg are not consistent (null fe is invalid anyway)

alsala (approved):

VolodymyrBabaiev (approved):

i-Alex (approved):

src/gtest/test_libzendoo.cpp
ptagl: Done with commit [304694d2ecefdfe23ee5750d1c44e1b4562ca5ba](https://github.com/HorizenOfficial/zen/pull/429/commits/304694d2ecefdfe23ee5750d1c44e1b4562ca5ba).

I've also added a check to be sure that the first bit of the random custom field is set to "1".

src/gtest/test_libzendoo.cpp
ptagl: Done with commit [304694d2ecefdfe23ee5750d1c44e1b4562ca5ba](https://github.com/HorizenOfficial/zen/pull/429/commits/304694d2ecefdfe23ee5750d1c44e1b4562ca5ba).

src/gtest/test_libzendoo.cpp
```diff
@@ -2004,6 +2018,41 @@ TEST(CctpLibrary, TestGetLeadingZeros)
...
+ * @brief Check the correctness of the GetBytesFromBits utility function
+ * 
+ * In order to keep the test fast, the iteration is limited to the first 2^16 numbers.
+ */
+TEST(CctpLibrary, TestGetBytesFromBits)
+{
+    int reminder = 0;
+
+    // Check that the function works properly with the "0" input
+    ASSERT_EQ(0, getBytesFromBits(0, reminder));
```
alsala, line 2011: This api has an int for the number of bits, it could be useful to add also a oneliner for testing negative values

alsala (approved):

albertog78 (approved):

PR #432, opened 2022-01-28
Add 'merkleTree' and 'scTxsCommitment' to 'getblocktemplate' exposed by optional parameter

cronicc:


src/rpc/mining.cpp
```diff
@@ -711,8 +721,20 @@ UniValue getblocktemplate(const UniValue& params, bool fHelp)
     }
 
     UniValue result(UniValue::VOBJ);
+    pblock->hashMerkleRoot = pblock->BuildMerkleTree();
```
cronicc: @ptagl it would be better to move lines 724-728 inside of `if (includeMerkleRoots)`, that way we don't have the performance impact of calculating the MerkleTree and ScTxsCommitment with the parameter being `false`.

ptagl: Done with commit [c8f4d937640caee9607d0d1e817c1daf2e9b1e9c](https://github.com/HorizenOfficial/zen/pull/432/commits/c8f4d937640caee9607d0d1e817c1daf2e9b1e9c).

ptagl (approved):

alsala (approved):

albertog78 (approved):

toniteo:
Ok

PR #435, opened 2022-02-09
Sidechain creation version

ptagl:
This PR introduces the concept of "Sidechain version" in the Sidechain Creation Output.

To make the change easier, such version field has not been just added to `CTxScCreationOut` but it's been stored together with the `withdrawalEpochLength`.

So, before these changes, the serialization of a `CTxScCreationOut` was reading/writing 4 bytes (32 bits) for the `withdrawalEpochLength`, now instead the first (most significant byte) of the `withdrawalEpochLength` is used for the sidechain `version`. This is possible because even though the `withdrawalEpochLength` had space for 32 bits integers, the maximum allowed value is 4032 (using only 12 bits). For this reason, we can assume that the sidechain version before this PR was already set to 0.

```
              32 bits | 4 bytes                           8 bits | 1 byte      24 bits | 3 bytes

+-------------------------------------------+            +---------------+---------------------------+
|                                           |            |               |                           |
|                                           |            |               |                           |
|           withdrawalEpochLength           | +--------> |    version    |   withdrawalEpochLength   |
|                                           |            |               |                           |
|                                           |            |               |                           |
+-------------------------------------------+            +---------------+---------------------------+

```

List of changes:
- Added Fork 9 (Sidechain Version) that activates at block height 1100000, 1000000, 450 for Mainnet, Testnet and Regtest (**numbers to be set properly**)
- Added the `version` field in `CTxScCreationOut` and in `ScFixedParameters`
- Added the `sidechainVersion` parameter to the `GetFieldElement` function to select the proper validation logic
- Updated RPC commands `sc_create`, `createrawtransaction`, `getscinfo` to include the `version` argument (mandatory when creating a new sidechain)
- Added unit and regression tests

Additionally, while developing unit and regression tests, some helper classes have been created/extended to make it easier the definition of common scenarios (e.g. crafting a sidechain creation transaction, testing the `AcceptTxToMemoryPool()` function, etc.).


src/sc/sidechain.h
```diff
@@ -132,7 +130,17 @@ class CSidechain {
...
-        READWRITE(sidechainVersion);
+        /*
+            This is some reserved space that at beginning was intended to be used for storing the sidechain version.
+            However, since Zendoo was released without handling the version (in particular in CTxScCreationOut), such field
+            has been introduced later with fork 9 and "extracted" from the first byte of the withdrawalEpochLength inside
+            ScFixedParameters.
+            The initial "sidechainVersion" variable has then been removed to avoid confusion and inconsistencies and its
+            space in the serialization is held by this dummy/reserved variable.
+        */
+        uint32_t reserved = 0xffffffff;
```
alsala: Minor: probably it is best to keep 0 values in order not to have any difference with the value stored so far

qa/rpc-tests/sc_create.py
```diff
@@ -72,13 +72,39 @@ def run_test(self):
         constant = generate_random_field_element_hex()
 
         # ---------------------------------------------------------------------------------------
-        # Node 2 try creating a SC with insufficient funds
-        mark_logs("\nNode 2 try creating a SC with insufficient funds", self.nodes, DEBUG_MODE)
+        # Node 2 tries to create a SC without version
```
i-Alex, line 75: what if user will try to set `version` > 1?
let's add the test as well.

src/primitives/transaction.cpp
```diff
@@ -1069,13 +1072,24 @@ bool CTransaction::ContextualCheck(CValidationState& state, int nHeight, int dos
...
             //enforce empty joinsplit for transparent txs and sidechain tx
             if(!GetVjoinsplit().empty()) {
                 return state.DoS(dosLevel, error("ContextualCheck(): transparent or sc tx but vjoinsplit not empty"),
                                      CValidationState::Code::INVALID, "bad-txns-transparent-jsnotempty");
             }
+
+            //enforce that any eventual SC creation output is using a valid sidechain version for the current active fork
+            for (const CTxScCreationOut& scCreationOutput : vsc_ccout)
+            {
+                if (scCreationOutput.version > maxSidechainVersionAllowed)
```
i-Alex, line 1086: Why we should allow sc version `0` after the `sc_version_fork`?
Why not to allow only the sc with version `1`?

src/primitives/transaction.h
```diff
@@ -585,6 +585,20 @@ friend class CTransaction;
     void GenerateScId(const uint256& txHash, unsigned int pos) const;
 
 public:
+    /**
+     * @brief  Version of the sidechain.
+     * The sidechain version has been introduced with hard fork 9 to increase the flexibility of the sidechain mechanism.
+     * Since the Sidechain Creation Output was at the beginning designed without the version, a single byty for such field
```
i-Alex: `byty` -> `byte`

src/sc/sidechaintypes.h
```diff
@@ -509,6 +511,7 @@ struct ScFixedParameters
     ADD_SERIALIZE_METHODS;
     template <typename Stream, typename Operation>
     inline void SerializationOp(Stream& s, Operation ser_action, int nType, int nVersion) {
+        READWRITE(version);
```
i-Alex: I think we can't add `READWRITE(version)` for free for `ScFixedParameters` struct.
`ScFixedParameters` is used as a part of `CSidechain` class serialization.
The node without these changes will not be able to recover the old `CSidechain` records from the database after the update to the recent code base.
We should have the same serialization approach as inside Transaction.

src/gtest/test_mempool.cpp
```diff
@@ -351,6 +351,65 @@ TEST(Mempool, SproutNegativeVersionTx) {
...
+    // Go to the next block after the SidechainVersionFork
+    testManager.ExtendChainActiveToHeight(sidechainVersionForkHeight);
+
+    // Check that a Sidechain Creation transaction with version 0 is accepted
+    state = CValidationState();
+    EXPECT_EQ(MempoolReturnValue::VALID, testManager.TestAcceptTxToMemoryPool(state, mtx_v0));
+
+    // Check that a Sidechain Creation transaction with version 1 is accepted
+    state = CValidationState();
+    EXPECT_EQ(MempoolReturnValue::VALID, testManager.TestAcceptTxToMemoryPool(state, mtx_v1));
```
i-Alex, line 409: Check the behavior of the `CTxMemPool::removeStaleTransactions`. We should remove Txs with sc_creation version `1` in case of the rollback before the fork, for example: `testManager.ExtendChainActiveToHeight(sidechainVersionForkHeight - 1);`
Note: maybe `removeStaleTransactions` is not the best method where to put `sc cr version` check, but for sure we need it.

src/gtest/test_libzendoo.cpp
```diff
@@ -1030,580 +1058,593 @@ TEST(CctpLibrary, BitVectorCertificateFieldFullBzip2)
...
-    for (int i = 0; i < sizeof(field_bytes); i++)
-        printf("%02x", ((unsigned char*)field_bytes)[i]);
-    printf("]\n");
+        size_t bt_list_len = vbt_list.size();
 
-    for (const CBwtRequestOut& ccout : tx.GetVBwtRequestOut() )
-    {
-        wrappedFieldPtr sptrScId = CFieldElement(ccout.GetScId()).GetFieldElement();
-        field_t* scid_fe = sptrScId.get();
+        int custom_fields_len = cert.vFieldElementCertificateField.size() + cert.vBitVectorCertificateField.size();
```
i-Alex, line 1375: We need to discuss the cert custom field usage in the commitment tree before and after the fork.

i-Alex (changes requested):

src/wallet/rpcwallet.cpp
```diff
@@ -821,6 +822,20 @@ UniValue sc_create(const UniValue& params, bool fHelp)
             throw JSONRPCError(RPC_INVALID_PARAMETER, string("Duplicate key in input: ") + s);
     }
 
+    // ---------------------------------------------------------
+    int sidechainVersion;
+    if (setKeyArgs.count("version"))
+    {
+        sidechainVersion = find_value(inputObject, "version").get_int();
+
+        if (sidechainVersion < 0 || sidechainVersion > ForkManager::getInstance().getHighestFork()->getMaxSidechainVersion())
```
alsala: We could also check the specific fork version using `chainActive.Height()`

src/primitives/transaction.cpp
ptagl: We decided to allow the creation of old-style sidechains in order to let users who are prepared to create a sidechain to be able to do that also after the fork point with minimal changes (right now just updating the call to `sc_create` RPC command).

src/primitives/transaction.h
ptagl: Fixed with commit [4481d65fe92c2936e85e975017fdc573b40257e8](https://github.com/HorizenOfficial/zen/pull/435/commits/4481d65fe92c2936e85e975017fdc573b40257e8).

src/wallet/rpcwallet.cpp
ptagl: Fixed with commit [b35e836f000e35d3356fecb882fb6dc17b24ae96](https://github.com/HorizenOfficial/zen/pull/435/commits/b35e836f000e35d3356fecb882fb6dc17b24ae96).

qa/rpc-tests/sc_create.py
ptagl: Any `version > 1` will be rejected by the RPC command (and obviously there are the same checks in `AcceptTxToMemoryPool()` and `ConnectBlock()` functions).

Python test extended with commit [0b50c831acf78c8ee7ce4b28a3d9280bba83896b](https://github.com/HorizenOfficial/zen/pull/435/commits/0b50c831acf78c8ee7ce4b28a3d9280bba83896b).

src/sc/sidechain.h
ptagl: Fixed with commit [ccdf6c5a5feb54b6353030ee8afec4570281d25d](https://github.com/HorizenOfficial/zen/pull/435/commits/ccdf6c5a5feb54b6353030ee8afec4570281d25d).

src/sc/sidechaintypes.h
ptagl: Fixed with commit [e888104ef55902c02162b2a31abb7b6cfc679b13](https://github.com/HorizenOfficial/zen/pull/435/commits/e888104ef55902c02162b2a31abb7b6cfc679b13).

src/gtest/test_libzendoo.cpp
ptagl: In order to let forgers know the version of other sidechains, we added a new websocket API to retrieve such information given a list of sidechain IDs as input (see commit [967bdd0bdf8aa6d7998c718b8ad47659795971ac](https://github.com/HorizenOfficial/zen/pull/435/commits/967bdd0bdf8aa6d7998c718b8ad47659795971ac)).

src/primitives/transaction.h
```diff
@@ -610,7 +625,27 @@ friend class CTransaction;
...
+            
+            // Get the most significant byte
+            version = withdrawalEpochLengthAndVersion >> 24;
+
+            // Get the least significant 3 bytes
+            withdrawalEpochLength = withdrawalEpochLengthAndVersion & 0x00FFFFFF;
+        }
+        else
+        {
+            int withdrawalEpochLengthAndVersion = (version << 24) | withdrawalEpochLength;
```
albertog78, line 648: Would be useful to consider withdrawalEpochLength & 0x00FFFFFF  and version & 0xFF or assert they are in valid ranges?

src/sc/sidechaintypes.h
```diff
@@ -509,7 +511,25 @@ struct ScFixedParameters
...
+            
+            // Get the most significant byte
+            version = withdrawalEpochLengthAndVersion >> 24;
+
+            // Get the least significant 3 bytes
+            withdrawalEpochLength = withdrawalEpochLengthAndVersion & 0x00FFFFFF;
+        }
+        else
+        {
+            int withdrawalEpochLengthAndVersion = (version << 24) | withdrawalEpochLength;
```
albertog78, line 547: see comment in CTxScCreationOut serialization code

albertog78 (changes requested):

src/gtest/test_mempool.cpp
ptagl: After double checking the code, we noticed that we didn't include a similar check for the Zendoo fork point and there must be a reason for that.

In the end we recalled that the we don't ever have to include fork point checks in the `CTxMemPool::removeStaleTransactions` function since it is called only when disconnecting a tip after the generation of a blockchain fork. In this case, we are not only disconnecting blocks, but also reconnecting some others and we can be sure that the final height will be higher than the previous one (otherwise the fork couldn't have led to the disconnection of the tip) and thus we don't need to validate the transactions against the fork point activation.

src/primitives/transaction.h
ptagl: I think that we can skip the check for the `version` field due to its `uint8_t` datatype (so it will never be greater than 0xFF). It might instead be useful for the `withdrawalEpochLength` since it could overwrite the `version` during the OR bitwise operation, but I'm not sure if it's possible also in practice since the `withdrawalEpochLength` is limited by the chain parameters (see `nScMaxWithdrawalEpochLength` in `chainparams.cpp`).

Do you have any specific scenario in mind that we should take care of?

albertog78: Agree about version. For what regards withdrawalEpochLength the mask (or assert) would be useful to detect / prevent buggy clients generating messy txs (eg. when tx is not yet in a block / mempool then not validated yet by consensus rules).

albertog78 (approved):

alsala (approved):

i-Alex (approved):

src/zen/forks/fork9_sidechainversionfork.cpp
```diff
@@ -0,0 +1,14 @@
+#include "fork9_sidechainversionfork.h"
+
+namespace zen {
+
+SidechainVersionFork::SidechainVersionFork()
+{
+    // TODO: set proper fork height values.
+    setHeightMap({{CBaseChainParams::Network::MAIN,1100000},
+                  {CBaseChainParams::Network::REGTEST,450},
+                  {CBaseChainParams::Network::TESTNET,1000000}});
```
cronicc: Set higher temporary activation heights, these block heights have already passed on testnet/mainnet.

src/gtest/test_forkmanager.cpp
```diff
@@ -286,3 +287,20 @@ TEST(ForkManager, SidechainForkMainnet) {
 	EXPECT_EQ(ForkManager::getInstance().getNewBlockVersion(1047624), BLOCK_VERSION_SC_SUPPORT);
 	EXPECT_EQ(ForkManager::getInstance().getNewBlockVersion(1047625), BLOCK_VERSION_SC_SUPPORT);
 }
+
+TEST(ForkManager, SidechainVersionForkMainnet) {
+    SelectParams(CBaseChainParams::MAIN);
+
+    // TODO: set proper fork height value.
+    int sidechainVersionForkHeight = 1100000;
```
cronicc: Set higher temporary activation height. Extend test with testnet and regtest coverage.

cronicc (changes requested):

src/zen/forks/fork9_sidechainversionfork.cpp
ptagl: Fixed with commit [190d49622f7e7e70a9d60a74426dc82c813bbb20](https://github.com/HorizenOfficial/zen/pull/435/commits/190d49622f7e7e70a9d60a74426dc82c813bbb20).

src/gtest/test_forkmanager.cpp
ptagl: Fixed with commit [190d49622f7e7e70a9d60a74426dc82c813bbb20](https://github.com/HorizenOfficial/zen/pull/435/commits/190d49622f7e7e70a9d60a74426dc82c813bbb20).

cronicc (approved):

PR #439, opened 2022-03-14
Sc Version - Test for commitment tree root computation

ptagl:
This PR introduces additional tests related to the management of the sidechain version, in particular when computing the sidechain commitment tree of a block that includes certificates of both version 0 and 1.

List of changes:
- Extended a Python test to generate a raw block containing two certificates (version 0 and version 1) and a sidechain creation (version 0) and to display the commitment tree root
- Implemented a unit test that deserializes the raw block generated for the test above and then checks that the sidechain commitment tree root is computed as expected

It was not possible to rely on the Python test only because each time it's run it returns different values (different sidechain IDs, due to different transaction hashes, due to different transaction inputs, etc.).

Note that a [similar test](https://github.com/HorizenOfficial/Sidechains-SDK/commit/460ad1f0984e96925e20be8e6c434d2acda439d4) has been introduced on the SDK side (it uses the same data).


i-Alex (approved):

src/gtest/test_libzendoo.cpp
```diff
@@ -625,6 +626,71 @@ TEST(SidechainsField, NakedZendooFeatures_EmptyTreeCommitmentCalculation)
...
+ * 
+ * https://github.com/HorizenOfficial/Sidechains-SDK/commit/460ad1f0984e96925e20be8e6c434d2acda439d4
+ */
+TEST(SidechainsField, CommitmentComputationFromSerializedBlock)
+{
+    SelectParams(CBaseChainParams::REGTEST);
+    const BlockchainTestManager& testManager = BlockchainTestManager::GetInstance();
+
+    // Read hex string from file
+    std::ifstream t("./src/gtest/test-data/block_with_v0_and_v1_certificates.hex");
```
alsala: There might be problems if the UT is not executed from the standard path. In such case the resulting blockData would not be valid because the ifstream would not be valid too

alsala (changes requested):

src/gtest/test_libzendoo.cpp
ptagl: Fixed with commit [50747f8dbabe82f3c6f10f78431ea6e85b3f681b](https://github.com/HorizenOfficial/zen/pull/439/commits/50747f8dbabe82f3c6f10f78431ea6e85b3f681b).

Added a way to retrieve the path of the file containing the raw block hex string from the path of the `gtest` binary.

i-Alex (approved):

src/gtest/test_libzendoo.cpp
alsala: Ok, seems fine!

alsala (approved):

PR #446, opened 2022-04-04
Backport upstream improvements

cronicc:
Backports of:
 * https://github.com/zcash/zcash/pull/2873 - improve sync performance by increasing default value of dbcache
 * https://github.com/zcash/zcash/pull/2797 - Implement z_mergetoaddress for combining UTXOs and notes
 * https://github.com/zcash/zcash/pull/3080 - Return JoinSplit and JoinSplitOutput indexes in z_listreceivedbyaddress
 * https://github.com/zcash/zcash/pull/2913 - Add z_listunspent RPC call
 * https://github.com/zcash/zcash/pull/3106 - Add Note Locking to z_mergetoaddress


src/wallet/rpcwallet.cpp
```diff
@@ -4762,6 +4895,9 @@ UniValue z_listreceivedbyaddress(const UniValue& params, bool fHelp)
         obj.pushKV("amount", ValueFromAmount(CAmount(entry.plaintext.value())));
         std::string data(entry.plaintext.memo().begin(), entry.plaintext.memo().end());
         obj.pushKV("memo", HexStr(data));
+        // (txid, jsindex, jsoutindex) is needed to globally identify a note
+        obj.pushKV("jsindex", entry.jsop.js);
+        obj.pushKV("jsoutindex", entry.jsop.n);
```
a-petrini, line 4902: Please, add the description of the added fields to the function dcoumentation, as done in `UniValue z_listunspent(const UniValue& params, bool fHelp)`

cronicc: Fixed in https://github.com/HorizenOfficial/zen/pull/446/commits/5378dbf723b880a8bab27983ea0947356512973d

src/wallet/rpcwallet.cpp
```diff
@@ -4019,6 +4020,138 @@ UniValue listunspent(const UniValue& params, bool fHelp)
...
+
+        // Keep track of addresses to spot duplicates
+        set<std::string> setAddress;
+
+        // Sources
+        for (const UniValue& o : addresses.getValues()) {
+            if (!o.isStr()) {
+                throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, expected string");
+            }
+            string address = o.get_str();
```
drgora: I'd suggest to make this a const& to save one copy:
```suggestion
std::string const& address = o.get_str();
```

src/wallet/rpcwallet.cpp
```diff
@@ -5812,6 +5950,347 @@ UniValue z_shieldcoinbase(const UniValue& params, bool fHelp)
...
+
+    // Keep track of addresses to spot duplicates
+    std::set<std::string> setAddress;
+
+    // Sources
+    for (const UniValue& o : addresses.getValues()) {
+        if (!o.isStr())
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, expected string");
+
+        std::string address = o.get_str();
```
drgora: Just like above, I'd suggest:
`std::string const& address = o.get_str();`

src/wallet/asyncrpcoperation_mergetoaddress.cpp
```diff
@@ -0,0 +1,943 @@
...
+using namespace libzcash;
+
+int mta_find_output(UniValue obj, int n)
+{
+    UniValue outputMapValue = find_value(obj, "outputmap");
+    if (!outputMapValue.isArray()) {
+        throw JSONRPCError(RPC_WALLET_ERROR, "Missing outputmap for JoinSplit operation");
+    }
+
+    UniValue outputMap = outputMapValue.get_array();
```
drgora: I'd suggest to make this a const& to save one copy:
`UniValue const& outputMap = outputMapValue.get_array();`

src/wallet/rpcwallet.cpp
```diff
@@ -4019,6 +4020,138 @@ UniValue listunspent(const UniValue& params, bool fHelp)
...
+    bool fIncludeWatchonly = false;
+    if (params.size() > 2) {
+        fIncludeWatchonly = params[2].get_bool();
+    }
+
+    LOCK2(cs_main, pwalletMain->cs_wallet);
+
+    // User has supplied zaddrs to filter on
+    if (params.size() > 3) {
+        UniValue addresses = params[3].get_array();
```
drgora: Same as above, I'd suggest to make `addresses` a `const&` here

src/wallet/rpcwallet.cpp
```diff
@@ -5812,6 +5950,347 @@ UniValue z_shieldcoinbase(const UniValue& params, bool fHelp)
...
+
+    LOCK2(cs_main, pwalletMain->cs_wallet);
+
+    bool useAny = false;
+    bool useAnyUTXO = false;
+    bool useAnyNote = false;
+    std::set<CBitcoinAddress> taddrs = {};
+    std::set<libzcash::PaymentAddress> zaddrs = {};
+
+    UniValue addresses = params[0].get_array();
```
drgora: Same as above, I'd suggest to make `addresses` a `const&` here

src/wallet/rpcwallet.cpp
```diff
@@ -5812,6 +5950,347 @@ UniValue z_shieldcoinbase(const UniValue& params, bool fHelp)
...
+    std::string memo;
+    if (params.size() > 5) {
+        memo = params[5].get_str();
+        if (!isToZaddr) {
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "Memo can not be used with a taddr.  It can only be used with a zaddr.");
+        } else if (!IsHex(memo)) {
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, expected memo data in hexadecimal format.");
+        }
+        if (memo.length() > ZC_MEMO_SIZE*2) {
+            throw JSONRPCError(RPC_INVALID_PARAMETER,  strprintf("Invalid parameter, size of memo is larger than maximum allowed %d", ZC_MEMO_SIZE ));
```
drgora: Any reason why we print `ZC_MEMO_SIZE` here, while we check for `ZC_MEMO_SIZE*2`?

src/wallet/rpcwallet.cpp
```diff
@@ -5812,6 +5950,347 @@ UniValue z_shieldcoinbase(const UniValue& params, bool fHelp)
...
+        } else if (address == "ANY_TADDR") {
+            useAnyUTXO = true;
+        } else if (address == "ANY_ZADDR") {
+            useAnyNote = true;
+        } else {
+            CBitcoinAddress taddr(address);
+            if (taddr.IsValid()) {
+                // Ignore any listed t-addrs if we are using all of them
+                if (!(useAny || useAnyUTXO)) {
+                    taddrs.insert(taddr);
```
drgora: I'd suggest to `taddrs.insert(std::move(taddr));` here, we should save some mem copies.

src/wallet/rpcwallet.cpp
```diff
@@ -5812,6 +5950,347 @@ UniValue z_shieldcoinbase(const UniValue& params, bool fHelp)
...
+    bool useAnyNote = false;
+    std::set<CBitcoinAddress> taddrs = {};
+    std::set<libzcash::PaymentAddress> zaddrs = {};
+
+    UniValue addresses = params[0].get_array();
+    if (addresses.size()==0)
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, fromaddresses array is empty.");
+
+    // Keep track of addresses to spot duplicates
+    std::set<std::string> setAddress;
```
drgora, line 6038: `std::unordered_set` might be better suited for this use case

src/wallet/rpcwallet.cpp
```diff
@@ -5812,6 +5950,347 @@ UniValue z_shieldcoinbase(const UniValue& params, bool fHelp)
...
+            }
+        }
+
+        if (setAddress.count(address))
+            throw JSONRPCError(RPC_INVALID_PARAMETER, string("Invalid parameter, duplicated address: ") + address);
+        setAddress.insert(address);
+    }
+
+    // Validate the destination address
+    auto destaddress = params[1].get_str();
```
drgora: This could also be a `std::string const&`

src/wallet/asyncrpcoperation_mergetoaddress.cpp
```diff
@@ -0,0 +1,943 @@
...
+        }
+    }
+
+    throw std::logic_error("n is not present in outputmap");
+}
+
+AsyncRPCOperation_mergetoaddress::AsyncRPCOperation_mergetoaddress(
+    CMutableTransaction contextualTx,
+    std::vector<MergeToAddressInputUTXO> utxoInputs,
+    std::vector<MergeToAddressInputNote> noteInputs,
```
drgora: We make local copies of these vectors at line 62, so I'd suggest to make these `const&` in the signature to avoid unnecessary mem copies.

src/wallet/wallet.cpp
```diff
@@ -4431,7 +4480,7 @@ void CWallet::GetFilteredNotes(std::vector<CNotePlaintextEntry> & outEntries, st
             PaymentAddress pa = nd.address;
 
             // skip notes which belong to a different payment address in the wallet
-            if (fFilterAddress && !(pa == filterPaymentAddress)) {
+            if (!(filterAddresses.empty() || filterAddresses.count(pa))) {
```
AndreaLanfranchi: Seems to me a bit convoluted to read. What is the intention here ? 
Due to the negation it does not continue when:
- either `filterAddresses.empty() == true` (which implies `filterAddresses.count(pa) == 0` for sure)
- or `filterAddresses.count(pa) != 0` (which implies `filterAddresses.empty() == false` for sure)

It continues only if `filterAddresses.count(pa) == 0` or in other word if `pa ∉ filterAddesses`

So why not simply ?
```suggestion
            if (!filterAddresses.contains(pa)) {
```

AndreaLanfranchi: Which raises an immediate question ... 
why even enter the loop `for (auto & pair : wtx.mapNoteData) {` if `filterAddresses` is empty ?
Under such condition there is no point traversing it and do continuous lookups on an empty set

Edit. After better comprehension this comment is wrong

src/wallet/wallet.cpp
```diff
@@ -4658,3 +4712,78 @@ std::shared_ptr<CWalletTransactionBase> CWalletTransactionBase::MakeWalletObject
...
+        CWalletTransactionBase& wtx = *(p.second);
+
+        // Filter the transactions before checking for notes
+        if (!CheckFinalTx(*wtx.getTxBase()) || !wtx.HasMatureOutputs() || wtx.GetDepthInMainChain() < minDepth || wtx.GetDepthInMainChain() > maxDepth) {
+            continue;
+        }
+
+        if (wtx.mapNoteData.size() == 0) {
+            continue;
+        }
```
AndreaLanfranchi, line 4736: This is pleonastic and unneeded. When `wtx.mapNoteData.empty()` the following range loop does not even enter.

src/wallet/wallet.cpp
```diff
@@ -4658,3 +4712,78 @@ std::shared_ptr<CWalletTransactionBase> CWalletTransactionBase::MakeWalletObject
...
+
+        for (auto & pair : wtx.mapNoteData) {
```
AndreaLanfranchi, line 4738: Maybe structured binding adds more readibility. Also, imho, derived paired values should be marked const.
```suggestion
        for (const auto & [/*JSOutPoint*/ jsop, /*CNoteData*/ nd] : wtx.mapNoteData) {
```

src/wallet/wallet.cpp
```diff
@@ -4658,3 +4712,78 @@ std::shared_ptr<CWalletTransactionBase> CWalletTransactionBase::MakeWalletObject
...
 }
+
+/* Find unspent notes filtered by payment address, min depth and max depth */
+void CWallet::GetUnspentFilteredNotes(
+    std::vector<CUnspentNotePlaintextEntry>& outEntries,
+    std::set<PaymentAddress>& filterAddresses,
+    int minDepth,
+    int maxDepth,
+    bool requireSpendingKey)
+{
```
AndreaLanfranchi, line 4723: This opening block adds an unneeded indentation level : the lock stays active till end or early exit of function

Edit. Disregard this ... optical illusion caused my habit to see open braces for function on the same line of function definition

src/wallet/wallet.cpp
drgora: I agree the condition is a bit convoluted to read, but I believe your suggestion is not logically equal.
With the original loop, if `filterAddresses` is empty you do not execute `continue`, while with your suggestion you do execute `continue`.

I do not have the full picture clear yet, but my guess is that `filterAddresses` is an optional whitelist that you might not use. If that is the case, my suggestion would be:

```suggestion
            if (!filterAddresses.empty() && !filterAddresses.contains(pa)) {
```

src/wallet/asyncrpcoperation_mergetoaddress.cpp
```diff
@@ -0,0 +1,943 @@
...
+    }
+
+    set_state(OperationStatus::EXECUTING);
+    start_execution_clock();
+
+    bool success = false;
+
+#ifdef ENABLE_MINING
+#ifdef ENABLE_WALLET
+    GenerateBitcoins(false, NULL, 0);
```
drgora, line 128: It might be nice to add a brief comment on why we stop mining here, and resume later on

src/wallet/asyncrpcoperation_mergetoaddress.cpp
```diff
@@ -0,0 +1,943 @@
...
+    /**
+     * END SCENARIO #2
+     */
+
+
+    // Copy zinputs to more flexible containers
+    std::deque<MergeToAddressInputNote> zInputsDeque;
+    for (auto o : noteInputs_) {
+        zInputsDeque.push_back(o);
+    }
```
drgora: I'd suggest the following form, which is more compact and saves some mem copies:

```suggestion
    std::deque<MergeToAddressInputNote> zInputsDeque(noteInputs_.begin(), noteInputs_.end());
```

src/wallet/asyncrpcoperation_mergetoaddress.cpp
```diff
@@ -0,0 +1,943 @@
...
+    sign_send_raw_transaction(obj);
+    return true;
+}
+
+
+/**
+ * Sign and send a raw transaction.
+ * Raw transaction as hex string should be in object field "rawtxn"
+ */
+void AsyncRPCOperation_mergetoaddress::sign_send_raw_transaction(UniValue obj)
```
drgora: `obj` could also be passed by reference:

```suggestion
void AsyncRPCOperation_mergetoaddress::sign_send_raw_transaction(UniValue const& obj)
```

src/wallet/asyncrpcoperation_mergetoaddress.cpp
```diff
@@ -0,0 +1,943 @@
...
+    }
+    for (size_t i = 0; i < ZC_NUM_JS_OUTPUTS; i++) {
+        arrOutputMap.push_back(outputMap[i]);
+    }
+
+
+    // !!! Payment disclosure START
+    unsigned char buffer[32] = {0};
+    memcpy(&buffer[0], &joinSplitPrivKey_[0], 32); // private key in first half of 64 byte buffer
+    std::vector<unsigned char> vch(&buffer[0], &buffer[0] + 32);
```
drgora: It seems to me that there's no need for this `buffer` array, I'd suggest the following:

```suggestion
    std::vector<unsigned char> vch(32);
    memcpy(vch.data(), &joinSplitPrivKey_[0], 32);
```

src/wallet/asyncrpcoperation_mergetoaddress.cpp
```diff
@@ -0,0 +1,943 @@
...
+    UniValue obj(UniValue::VOBJ);
+    obj.pushKV("encryptednote1", encryptedNote1);
+    obj.pushKV("encryptednote2", encryptedNote2);
+    obj.pushKV("rawtxn", HexStr(ss.begin(), ss.end()));
+    obj.pushKV("inputmap", arrInputMap);
+    obj.pushKV("outputmap", arrOutputMap);
+    return obj;
+}
+
+std::array<unsigned char, ZC_MEMO_SIZE> AsyncRPCOperation_mergetoaddress::get_memo_from_hex_string(std::string s)
```
drgora: No need to copy the string here:

```suggestion
std::array<unsigned char, ZC_MEMO_SIZE> AsyncRPCOperation_mergetoaddress::get_memo_from_hex_string(std::string const& s)
```

drgora (commented):
The features introduced with this PR look reasonable to me.
The implementation looks functionally correct to me, even though I cannot vouch with strong confidence for the JoinSplit part.

I have left few comments here and there. These are mainly stylistic suggestions and to avoid unneeded memory copies, but given the limited impact on performance none of these is a real blocker.

Also, on a more general note, I believe it would be better to remove explicit initialization with the default values (e.g.  `src/wallet/rpcwallet.cpp#4085`}, and move hardcoded numeric literals into preprocessor macros, or enums, or something similar.

src/wallet/asyncrpcoperation_mergetoaddress.cpp
AndreaLanfranchi: Or even simpler use vector initialization directly just as
```
std::vector<unsigned char> vch(joinSplitPrivKey, 32);
```

src/wallet/asyncrpcoperation_mergetoaddress.cpp
AndreaLanfranchi: @drgora at first sight I don't agree on this.
I believe the copy is intentional : I found this object constructor used only in one shared_pointer and for this reason I suppose the object itself outlives the original vectors it has been costructed from (hence leaving the object to a reference which does not point to anything with obvious consequences).
What I would do instead is to change the constructor to `move` the ownership of the vectors to the class instance. In such way we don't have a copy and we've honored the original intention.

Edit: I misunderstood the intentions (thought would mean to change member of class to a const ref)

src/wallet/asyncrpcoperation_mergetoaddress.cpp
AndreaLanfranchi: I agree with @drgora with a comment addition though: this method is named "sign_send_raw_transaction" so I'm expecting the input parameter to be a "raw_transaction" ... why we do pass the whole "UniValue" object to pull the transaction from ?

src/wallet/asyncrpcoperation_mergetoaddress.cpp
AndreaLanfranchi: Agree on this with a hint: let's be consistent using `qualifier -> type -> name` (eg. `const std::string& s` instead of `std::string const& s`)

src/wallet/asyncrpcoperation_mergetoaddress.cpp
AndreaLanfranchi: And of course use the form `qualifier -> type -> name` (eg. `const UniValue& obj`)

src/wallet/rpcwallet.cpp
AndreaLanfranchi: Probably not worth here: in set there's no expetation of a high number of elements.

ptagl (approved):

drgora (approved):

AndreaLanfranchi (approved):
Ok for me too. Thank you.

PR #448, opened 2022-04-04
Fix: 'z_sendmany error=Witness for note commitment is null'

cronicc:


JackPiri:
Maybe a check on the actual necessity of calling `SetBestChain` can be introduced, something like:

```
if (pindex && pindex != pindexStart)  
  SetBestChain(chainActive.GetLocator(pindexPrev));
```

in order to avoid useless write-to-disk operations.

AndreaLanfranchi:
@JackPiri if I am reading correctly `pindex` is meant to be `nullptr` after line 2118 (exits the `while (pindex)`) hence the condition `if (pindex && pindex != pindexStart)` is meant to always fail (I don't see any `break` in that loop).

I'd rather change line 2032 from
`CBlockIndex* pindexPrev;`
to 
`CBlockIndex* pindexLast = pindexStart;`

and line 2113 from
`pindexPrev = pindex;`
to 
`pindexLast = pindex;`

and  finally implement the check at 2153 as 
```
if (pindexLast != pindexStart)
  SetBestChain(chainActive.GetLocator(pindexLast));
```

My two cents

drgora:
I'm not clear why we want to call `CWallet::SetBestChain()` on every call to `CWallet::ScanForWalletTransactions()`.

Would it not be enough to call `CWallet::SetBestChain()` during the shutdown, for example from `Shutdown()` in `init.cpp`? In this way we might avoid writing repeatedly to the storage at runtime?

cronicc:
> I'm not clear why we want to call `CWallet::SetBestChain()` on every call to `CWallet::ScanForWalletTransactions()`.
> 
> Would it not be enough to call `CWallet::SetBestChain()` during the shutdown, for example from `Shutdown()` in `init.cpp`? In this way we might avoid writing repeatedly to the storage at runtime?

CWallet::ScanForWalletTransactions() should only be called at runtime from init.cpp on startup, and when importing keys using RPC calls. So there shouldn't be any impact on performance unless rescanning.
My reasoning for flushing the witness cache at the end of the rescan is that there is a lower probability for an inconsistent cache in case zend crashes (e.g. OOM Kill) and doesn't go through the shutdown logic.

ptagl (approved):

drgora (approved):
This is ok for me, thanks

PR #451, opened 2022-05-11
Always send notifications about tip updates

drgora:
As of today, we generate _UpdatedBlockTip_ notifications in the main loop of ActivateBestChain().
This means that we do not actually generate _UpdatedBlockTip_ notifications for every tip update, but rather for
every completion of ActivateBestChainStep(), which might perform several block connections and disconnections at once.

As a consequence, when a new best chain is selected, we skip notifications about blocks being disconnected or connected in the process.
This behavior might be related to issue https://github.com/HorizenOfficial/Sidechains-SDK/issues/138.

This PR updates the code so that we always send _UpdatedBlockTip_ notifications on every change of the tip, including when
a block is being disconnected, and a new best chain is being selected.

Also notice that as of today we check whether the initial download is finished before we send notifications. With this patch, we
ignore this condition, and always send notifications.

In future, we might implement similar checks directly in the notification managers (e.g. rpc or ui).


ptagl:
Since we don't see any big advantage of changing the behavior of the `UpdatedBlockTip` notification, we prefer to close this PR for now.

PR #452, opened 2022-05-18
Update the build system to be compatible with latest Linux distributions

ptagl:
Zend currently fails to be compiled on the most updated Linux distributions, including Ubuntu 22.

This PR brings some changes to the build system to fix the issue, including:

- Update of the Boost dependency from v1.70 to v1.74
- Set a specific version of the `b2sdk` pip package to avoid Backblaze failured due to breaking updates
- Updated GMP lib dependency to version 6.2.1
- Added C++ 17 as default compilation version
- Updated `libzendoo` (MC Crypto Lib) to always statically include Bzip2 (it currently happens that it includes Bzip2 dynamically if it finds the related shared library installed on the system, consequently breaking the Zend linking phase)
- Fixed the syntax of the check for the presence of `libgmp` in `configure.ac` (the previous syntax was working with v2.69 but stopped from v2.71)

In the future we should discuss if it's the case to update also the continuous integration to compile on Ubuntu 22 (now it's using v18 and v20).

### DIFF
--- a/depends/patches/boost/signals2-noise.patch
+++ b/depends/patches/boost/signals2-noise.patch
@@ -1,0 +1,23 @@
+From fd27423fea5537bc857c1fa14bb0c25b994f77b3 Mon Sep 17 00:00:00 2001
+From: Frank Mori Hess <fmh6jj@gmail.com>
+Date: Mon, 20 Jul 2020 14:17:05 -0400
+Subject: [PATCH] Fix warning about deprecated
+ boost/function_output_iterator.hpp
+
+---
+ include/boost/signals2/detail/null_output_iterator.hpp | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/include/boost/signals2/detail/null_output_iterator.hpp b/include/boost/signals2/detail/null_output_iterator.hpp
+index 9e986959..dee4373c 100644
+--- a/include/boost/signals2/detail/null_output_iterator.hpp
++++ b/include/boost/signals2/detail/null_output_iterator.hpp
+@@ -11,7 +11,7 @@
+ #ifndef BOOST_SIGNALS2_NULL_OUTPUT_ITERATOR_HPP
+ #define BOOST_SIGNALS2_NULL_OUTPUT_ITERATOR_HPP
+ 
+-#include <boost/function_output_iterator.hpp>
++#include <boost/iterator/function_output_iterator.hpp>
+ 
+ namespace boost
+ {
--- a/src/random.h
+++ b/src/random.h
@@ -9,6 +9,7 @@
 #include "uint256.h"
 
 #include <functional>
+#include <limits>
 #include <stdint.h>
 
 /**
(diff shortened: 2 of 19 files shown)


src/random.h
```diff
@@ -9,6 +9,7 @@
 #include "uint256.h"
 
 #include <functional>
+#include <limits>
```
drgora, line 12: Since we are including `<limits>` in `src/random.h` now, I'd suggest to remove the same inclusion from `src/random.cpp`

depends/patches/boost/signals2-noise.patch
```diff
@@ -0,0 +1,23 @@
...
+index 9e986959..dee4373c 100644
+--- a/include/boost/signals2/detail/null_output_iterator.hpp
++++ b/include/boost/signals2/detail/null_output_iterator.hpp
+@@ -11,7 +11,7 @@
+ #ifndef BOOST_SIGNALS2_NULL_OUTPUT_ITERATOR_HPP
+ #define BOOST_SIGNALS2_NULL_OUTPUT_ITERATOR_HPP
+ 
+-#include <boost/function_output_iterator.hpp>
++#include <boost/iterator/function_output_iterator.hpp>
+
```
drgora, line 21: Tiny comment on the style here: it might be good to get rid of whitespaces at lines 17 and 21

src/validationinterface.cpp
```diff
@@ -6,6 +6,8 @@
 #include "validationinterface.h"
 #include <primitives/certificate.h>
 
+using namespace boost::placeholders;
```
drgora, line 9: Another small stylistic comment here: I believe using this namespace explicitly where needed might make the code  more readable. On the other hand, I see this is currently aligned with zcash

src/random.h
```diff
@@ -19,6 +20,34 @@ uint64_t GetRand(uint64_t nMax);
...
+        return std::numeric_limits<result_type>::max();
+    }
+
+    result_type operator()() {
+        result_type nRand = 0;
+        GetRandBytes((unsigned char*)&nRand, sizeof(nRand));
+        return nRand;
+    }
+
+    double entropy() const noexcept {
```
drgora: Just a notice that this function is not used anywhere, and might as well be `constexpr`.

drgora (approved):
I left few minor comments, none of which is a blocker.  This looks good to me.

depends/patches/boost/signals2-noise.patch
ptagl: As we discovered together, a white-space in patch files is automatically generated by `diff` and it's part of the syntax standard, so we keep it as it is.

src/validationinterface.cpp
ptagl: Even though using namespace in this way is not a best practice, I would keep this one as an exception just for two reasons:
- keeping the code aligned with Zcash (as you already mentioned)
- not cluttering the file with tens of namespaces for very small placeholder names (_1, _2, _3, etc.)

src/random.h
ptagl: Removed with commit [7fbbbd36aeca5cf515f2f6c66a22f83889eec988](https://github.com/HorizenOfficial/zen/pull/452/commits/7fbbbd36aeca5cf515f2f6c66a22f83889eec988).

src/random.h
ptagl: I removed that function with commit [7fbbbd36aeca5cf515f2f6c66a22f83889eec988](https://github.com/HorizenOfficial/zen/pull/452/commits/7fbbbd36aeca5cf515f2f6c66a22f83889eec988).

depends/patches/boost/signals2-noise.patch
drgora: Yes, makes sense.

src/validationinterface.cpp
drgora: Alright, makes sense.

cronicc (approved):

PR #453, opened 2022-05-26
update IsInitialBlockDownload in main.cpp

### LINKED ISSUE #74: zend sometimes warns of 0 blocks received in the last 4 hours during initial sync
Original issue: https://github.com/ZencashOfficial/zen_archived/issues/138

a-petrini:
modified IsInitialBlockDownload() in main.cpp:
- removed call to isAfterChainsplit() that generated a single warning "0 blocks received in the last 4 hours (96 expected)" at zend launch
- modified syntax to adhere to Zcash codebase

### DIFF
--- a/src/gtest/test_blockdownload.cpp
+++ b/src/gtest/test_blockdownload.cpp
@@ -14,8 +14,6 @@ TEST(initialblockdwnld, checkIBDState) {
     chainActive.SetTip(NULL);
     SelectParams(CBaseChainParams::MAIN);
     const CChainParams& chainParams = Params();
-    constexpr int MAIN_CHAIN_TEST_SIZE = 150000;
-    makeTemporaryMainChain(MAIN_CHAIN_TEST_SIZE);
 
     // 1.
     // fImporting, fReindex and fReindexFast supposedly initialized as false
@@ -74,7 +72,7 @@ TEST(initialblockdwnld, checkIBDState) {
     //
     // Set conditions so that all checks fail, and the end of function is reached so that:
     // - lockIBDState set to true
-    // - funciton returns false
+    // - function returns false
     pindexBestHeader->nTime = GetTime();
     EXPECT_FALSE(IsInitialBlockDownload());
 
@@ -94,30 +92,4 @@ TEST(initialblockdwnld, checkIBDState) {
 
     // Restore active chain tip
     chainActive.SetTip(originalTip);
-}
-
-
-void makeTemporaryMainChain(int trunk_size)
-{
-    // Create genesis block
-    CBlock b = Params().GenesisBlock();
-    CBlockIndex* genesis = AddToBlockIndex(b);
-    auto hashOfPrevBlock = b.GetHash();
-
-    // create the main trunk, from which some forks will possibly stem
-    for (int i = 0; i < trunk_size; i++)
-    {
-        CBlock b;
-        b.nVersion = MIN_BLOCK_VERSION;
-        b.nNonce = uint256(GetRandHash());  
-        b.nBits = arith_uint256(uint256(GetRandHash()).ToString() ).GetCompact();
-        b.hashPrevBlock = hashOfPrevBlock;
-        hashOfPrevBlock = b.GetHash();
-
-        CBlockIndex *bi = AddToBlockIndex(b);
-
-        chainActive.SetTip(bi);
-    }
-
-    return;
 }
--- a/src/gtest/test_blockdownload.cpp
+++ b/src/gtest/test_blockdownload.cpp
@@ -7,12 +7,25 @@ using namespace zen;
 
 // Checks IsInitialBlockDownload() conditions and lockIBDState latching to false
 TEST(initialblockdwnld, checkIBDState) {
-    // Init
+    ////// Init
     auto originalTip = chainActive.Tip();
-    chainActive.SetTip(nullptr);
     SelectParams(CBaseChainParams::MAIN);
     const CChainParams& chainParams = Params();
 
+    ///// Set initial conditions so that IsInitialBlockDownload() returns false
+    fImporting = fReindex = fReindexFast = fCheckpointsEnabled = false;
+    // Reset the active chain and add the genesis block
+    chainActive.SetTip(nullptr);
+    CBlock b = Params().GenesisBlock();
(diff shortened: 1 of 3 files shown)


src/gtest/test_blockdownload.cpp
```diff
@@ -0,0 +1,74 @@
...
+    // Expecting that fImporting, fReindex and fReindexFast are false, check if
+    // chainActive.Height() < Checkpoints::GetTotalBlocksEstimate(chainParams.Checkpoints())
+    fCheckpointsEnabled = true;
+    EXPECT_TRUE(IsInitialBlockDownload());
+
+    // Setting tip of active chain lower than checkpoint
+    auto block1 = CBlockIndex();
+    block1.nHeight = Checkpoints::GetTotalBlocksEstimate(chainParams.Checkpoints()) - 1;
+    chainActive.SetTip(&block1);
+    EXPECT_TRUE(IsInitialBlockDownload());
```
ptagl: Would it be possible to include this section in a loop and check that we always get `true` for every height in [0, 1, 2, ..., N] up to the checkpoint height?

I'm asking because in this way I guess we can spot also weird behaviors, like the one we had with the call to `isAfterChainsplit()` inside `IsInitialBlockDownload()`.

ptagl (changes requested):

src/gtest/test_blockdownload.cpp
a-petrini: The test has been updated for checking every value up to height N.

ptagl (approved):

src/gtest/test_blockdownload.cpp
```diff
@@ -0,0 +1,95 @@
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+
+#include "main.h"
+
+using namespace zen;
+
+void makeTemporaryMainChain(int trunk_size);
```
drgora: This seems to be a leftover from the previous implementation, we should probably remove it?

src/gtest/test_blockdownload.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+// Checks IsInitialBlockDownload() conditions and lockIBDState latching to false
+TEST(initialblockdwnld, checkIBDState) {
+    // Init
+    auto originalTip = chainActive.Tip();
+    chainActive.SetTip(NULL);
```
drgora: I'd suggest to use `nullptr` consistently for the new code that we write, as you do in the rest of the function

src/gtest/test_blockdownload.cpp
```diff
@@ -0,0 +1,95 @@
...
+    // Check each individual variable and restore its state
+    fImporting   = true; EXPECT_TRUE(IsInitialBlockDownload()); fImporting   = false;
+    fReindex     = true; EXPECT_TRUE(IsInitialBlockDownload()); fReindex     = false;
+    fReindexFast = true; EXPECT_TRUE(IsInitialBlockDownload()); fReindexFast = false;
+
+    // 2.
+    // Expecting that fImporting, fReindex and fReindexFast are false, check if:
+    // chainActive.Height() < Checkpoints::GetTotalBlocksEstimate(chainParams.Checkpoints())
+    fCheckpointsEnabled = true;
+    EXPECT_TRUE(IsInitialBlockDownload());
```
drgora: Please correct me if I'm wrong, but I unless I'm missing something I believe that this line, similarly to all the previous `EXPECT_TRUE` lines, is not necessarily testing the specific conditions mentioned in the comments.

It looks to me that since the tip of `chainActive` is null, all these lines would return `true` anyway. If that is the case, I'd suggest to set all the conditions so that `IsInitialBlockDownload()==false` at the beginning of the test, and the toggle each variable as you are doing.

src/main.cpp
```diff
@@ -2072,25 +2072,32 @@ CAmount GetBlockSubsidy(int nHeight, const Consensus::Params& consensusParams)
     return nSubsidy;
 }
 
+
```
drgora: Tiny comment, but the file is mostly formatted consistently with one empty line between functions, I think it's good to stick to that style :)

src/gtest/test_blockdownload.cpp
a-petrini: removed in commit 8036ea80c617507a1a3fd033f29bfeaa13b08ff0

src/gtest/test_blockdownload.cpp
a-petrini: changed to nullptr in commit 8036ea80c617507a1a3fd033f29bfeaa13b08ff0

src/gtest/test_blockdownload.cpp
a-petrini: I agree that comments are misleading, and have been improved in commit 8036ea80c617507a1a3fd033f29bfeaa13b08ff0
The same commit also change the order of the checks in `IsInitialBlockDownload()` to correctly check all the conditions.
However, I cannot set the conditions so that `IsInitialBlockDownload()==false` at the beginning of the test, because once the atomic variable `lockIDBState` has been set, the functions is designed to always return false.

src/main.cpp
a-petrini: fixed in commit 4f633fca786a66af6379f8f9027db1019ff32894 :)

src/gtest/test_blockdownload.cpp
drgora: Yeah I agree that you do not want to call `IsInitialBlockDownload()` when it returns `false` before the end of the test. I was suggesting to set the conditions for `IsInitialBlockDownload()` to return `false`, without actually calling it, and then modify one condition at a time to verify that it actually impacts the result.

This might break with future code updates without us noticing, given that you cannot actually test that you set the correct conditions for `IsInitialBlockDownload()` to return `false`. On the other hand, I believe this it is still stricter than the current tests.

a-petrini: I agree: commit 12d22553057f4c551d53a17a1b29974149b34c00 modifies the test so that all conditions that may return `false` are set in advance, then each flag / if is test one at a time

drgora (approved):

cronicc (approved):

PR #454, opened 2022-05-31
Add the dependencies list in the documentation

### LINKED ISSUE #436: Can we get an actual dependency list instead of apt-get command?
I'm trying to build on Arch Linux, but the build is failing for some reason that I can't quite figure out (GCC build errors being notoriously unhelpful and all)

I suspect that I'm just missing a dependency or have a wrong version of something, but I can't tell what's actually needed, since the dependency list isn't truly a dependency list, and is really just an apt-get command to be run under Ubuntu.

There's a lot of Linux distros out there beyond Ubuntu, so it would be really nice if we could get an actual dependency list, maybe with officially supported version numbers for each dependency as well.

ptagl:
This PR adds a new `dependencies` file in the `doc` section, listing the dependencies needed to build Zen on Linux platforms.

It close #436.

### DIFF
--- a/doc/dependencies.md
+++ b/doc/dependencies.md
@@ -15,4 +15,4 @@ These are the dependencies needed to build Zen on Linux.
 | [Automake](https://www.gnu.org/software/automake/) | [1.16.5] |
 | [Libtool](https://www.gnu.org/software/libtool/) | [2.4.6] |
 | [pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/) | [0.29.2] |
-| [hexdump](https://www.gnu.org/software/coreutils/) | [2.37.2] |
+| [hexdump](https://www.kernel.org/pub/linux/utils/util-linux/) | [2.37.2] |
--- a/doc/dependencies.md
+++ b/doc/dependencies.md
@@ -1,0 +1,18 @@
+# Dependencies
+
+These are the dependencies needed to build Zen on Linux.
+
+| Dependency | Tested version |
+| --- | --- |
+| [Make](https://www.gnu.org/software/make/) | [4.3] |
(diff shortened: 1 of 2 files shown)


doc/dependencies.md
```diff
@@ -0,0 +1,18 @@
...
+| [Curl](https://curl.se/) | [7.81] |
+| [Bzip2](https://sourceware.org/bzip2/) | [1.0.8] |
+| [GCC](https://gcc.gnu.org) | [11.2.0] |
+| [Patch](https://savannah.gnu.org/projects/patch/) | [2.7.6] |
+| [Perl](https://www.perl.org/) | [5.34.0] |
+| [Autoconf](https://www.gnu.org/software/autoconf/) | [2.71] |
+| [Automake](https://www.gnu.org/software/automake/) | [1.16.5] |
+| [Libtool](https://www.gnu.org/software/libtool/) | [2.4.6] |
+| [pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/) | [0.29.2] |
+| [hexdump](https://www.gnu.org/software/coreutils/) | [2.37.2] |
```
drgora: According to `man hexdump`, "The hexdump command is part of the util-linux package which can be downloaded from Linux Kernel Archive <https://www.kernel.org/pub/linux/utils/util-linux/>"

ptagl: Fixed with commit [53f1131d8608bcdf0d1c9864a07413b136e42417](https://github.com/HorizenOfficial/zen/pull/454/commits/53f1131d8608bcdf0d1c9864a07413b136e42417).

drgora (approved):

a-petrini (approved):

cronicc (approved):

PR #457, opened 2022-06-28
Fixed BlockchainTestManager transaction generation

ptagl:
Due to a bug in the BlockchainTestManager class, it was not possible to create transaction that required an automatic input generation after calling the Reset() function.

This commit fixes the bug and adds a unit test to check that the class behaves as expected.

### DIFF
--- a/src/gtest/test_utils.cpp
+++ b/src/gtest/test_utils.cpp
@@ -1,0 +1,67 @@
+#include <gtest/gtest.h>
+
+#include "tx_creation_utils.h"
+#include "zen/forks/fork9_sidechainversionfork.h"
+
+using namespace blockchain_test_utils;
+
+class BlockchainHelperTest: public ::testing::Test
+{
+protected:
+
+    int sidechainForkHeight;
+
+    BlockchainHelperTest()
+    {
+        SidechainFork sidechainFork = SidechainFork();
+        sidechainForkHeight = sidechainFork.getHeight(CBaseChainParams::REGTEST);
+    }
+
+    void SetUp() override
+    {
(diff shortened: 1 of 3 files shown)


src/gtest/test_utils.cpp
```diff
@@ -0,0 +1,67 @@
...
+TEST_F(BlockchainHelperTest, CoinGeneration)
+{
+    BlockchainTestManager& testManager = BlockchainTestManager::GetInstance();
+
+    // Initialize the sidechain keys
+    testManager.GenerateSidechainTestParameters(ProvingSystem::CoboundaryMarlin, TestCircuitType::Certificate);
+
+    // Go to the last block before the SidechainVersionFork
+    
+    testManager.ExtendChainActiveToHeight(sidechainForkHeight);
```
drgora, line 46: I'd suggest to add a comment to explain if and why a specific height is necessary for this test.

drgora (commented):
I just left a minor comment, but it looks good to me otherwise.

a-petrini (approved):

src/gtest/test_utils.cpp
ptagl: Reaching that height is not strictly needed for the test, it was just a workaround to be able to easily create a transaction as the `BlockchainTestManager` has functions to create sidechain outputs but not normal ones.

Thinking a little bit more about it, I don't like this workaround, so, instead of adding a comment to explain it I will add a new helper function to create normal outputs.

I'll set this PR as draft until such change is ready.

JackPiri:

PR #461, opened 2022-07-13
Add 2204 ci worker:

tvangundy-hl:
Added -Wnonull flag for C compiles
Removed ipfs
Added pycryptodome to elminate dependency on RIPEMD160

### DIFF
--- a/qa/rpc-tests/test_framework/script.py
+++ b/qa/rpc-tests/test_framework/script.py
@@ -25,7 +25,6 @@ if sys.version > '3':
     bord = lambda x: x
 
 import struct
-import hashlib
 from Crypto.Hash import RIPEMD160
 
 from test_framework import bignum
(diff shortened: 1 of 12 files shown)


tvangundy-hl:
@ptagl , Please confirm usage of this C compiler flag.   -Wnonull  I needed to add this because the warning messages were exceeding the travis-ci maximum log length.

pycryptodome  was needed because the old library is not supported in ubuntu jammy's version of python/openssl.

See,
https://stackoverflow.com/questions/42601709/hashlib-cant-find-ripemd160

cronicc (approved):

qa/rpc-tests/test_framework/script.py
```diff
@@ -36,7 +37,10 @@
 OPCODE_NAMES = {}
 
 def hash160(s):
-    return hashlib.new('ripemd160', sha256(s)).digest()
+    return RIPEMD160.new(sha256(s)).digest()
+
+# def hash160(s):
+#     return hashlib.new('ripemd160', sha256(s)).digest()
```
ptagl: Even though it's not strictly necessary, I would definitely remove the old code (not sure if it's also possible to remove the import of `hashlib`).

ptagl (approved):

PR #458, opened 2022-06-29
Prepare management for fork 10

a-petrini:
This PR introduces the mandatory changes to the ForkManager class for the upcoming fork #10:
- updates to ForkManager class
- updates to the corresponding test unit
- updates to Python RPC test
In Python tests, height-related global variables have been gathered into a single dictionary.

PR #463, opened 2022-08-01
Non-ceasable sidechains - RC1

ptagl:
A detailed description of all changes will follow soon.

Currently this PR contains all the main changes needed to handle non-ceasable sideechains, including new validation rules for blocks and mempool.

Changes to the fork manager to introduce the new hard fork have been already merged with a [specific PR](https://github.com/HorizenOfficial/zen/pull/458).

In the next days, we are going to work on minor changes:
- Update RPC commands to perform checks over non-ceasable sidechains (e.g. do not allow to create a non-ceasable sidechains with a CSW verification key)
- Update RPC commands to include additional data for non-ceasable sidechains or to remove meaningless fields (e.g. `ceasingHeight` or `certSubmissionWindowLength`)
- Increase test coverage


src/sc/sidechain.h
```diff
@@ -179,18 +199,28 @@ class CSidechain {
...
+    }
+
+    static bool isNonCeasingSidechain(int version, int withdrawalEpochLength) {
+        // SC v2 can be either ceasing (as v0 and v1), or non-ceasing.
+        // Non ceasing sidechains are identified by fixedParams.withdrawalEpochLength == 0, which is not allowed for
+        // other sidechain versions.
+        return version == 2 && withdrawalEpochLength == 0;
+    }
+
+    bool isNonCeasing() const {
```
alsala, line 222: Cosmetic: consider adding a `canCease() `method to be used in test conditions instead of a double negation `if (!isNonCeasing())`

src/bitcoin-tx.cpp
```diff
@@ -23,6 +25,8 @@
 
 using namespace std;
 
+CCoinsViewCache *pcoinsTip = nullptr;
```
alsala, line 28: Why these lines have been added?

src/sc/sidechain.cpp
```diff
@@ -99,20 +226,24 @@ int CSidechain::GetCertSubmissionWindowEnd(int certEpoch) const
...
+} 
 
-int CSidechain::GetCertMaturityHeight(int certEpoch) const
+int CSidechain::GetCertMaturityHeight(int certEpoch, int includingBlockHeight) const
 {
     if (!isCreationConfirmed()) //default value
         return -1;
 
-    return GetCertSubmissionWindowEnd(certEpoch+1);
+    //return GetCertSubmissionWindowEnd(certEpoch+1);
```
alsala, line 240: commented code may be removed

src/sc/sidechaintypes.cpp
```diff
@@ -217,6 +217,9 @@ wrappedFieldPtr CFieldElement::GetFieldElement() const
 
 uint256 CFieldElement::GetLegacyHash() const
 {
+    if (this->byteVector.size() < 32)
```
alsala, line 220: Maybe we could use `Sidechain::SC_FE_SIZE_IN_BYTES`? Also below.

src/txmempool.h
```diff
@@ -119,7 +119,7 @@ struct CSidechainMemPoolEntry
 {
     uint256 scCreationTxHash;
     std::set<uint256> fwdTxHashes; 
-    std::map<int64_t, uint256> mBackwardCertificates; //quality -> certHash
+    std::map<std::pair<int64_t, int>, uint256> mBackwardCertificates; //quality -> certHash
```
alsala, line 122: The comment now is obsolete, this now is key=(quality/epoch), value=certHash

src/txmempool.cpp
```diff
@@ -1556,6 +1586,85 @@ bool CTxMemPool::checkIncomingTxConflicts(const CTransaction& incomingTx) const
...
+    int prevHeight = -1;
+
+    for (auto const& cert_it: sc_it->second.mBackwardCertificates)
+    {
+        // std::pair<int x, int y> is ordered as
+        // (x1,y1) > (x2,y2) if x1==x2 and y1 > y2
+
+        // retrieve referencedHeight
+        const CScCertificate& mapCert = mapCertificate.at(cert_it.second).GetCertificate();
+        if (mapCert.quality != 0)
```
alsala, line 1638: `quality != 0` means that the SC of the certificate can cease.
SC SDK minimum cert quality for ceasing SC is 1 (at least 1 cert signer), but in general for MC `quality==0` is valid also for certificate of a SC that can cease.
Would it be an option using a negative value, for instance `CScCertificate::QUALITY_NOT_CEASING = -2` for non ceasing SCs?

src/txmempool.cpp
```diff
@@ -1556,6 +1586,85 @@ bool CTxMemPool::checkIncomingTxConflicts(const CTransaction& incomingTx) const
...
+            return true; // isNonCeasing() == false
+
+        assert(mapCert.epochNumber > prevEpoch);
```
alsala, line 1641: We might assert also that `mapCert` quality and epoch is the same as the key of the iterator we are using:
```
assert(mapCert.quality == cert_it.first.first)
assert(mapCert.epochNumber == cert_it.first.second)
```

src/txmempool.cpp
```diff
@@ -1865,9 +1978,36 @@ bool CCoinsViewMemPool::GetSidechain(const uint256& scId, CSidechain& info) cons
...
+        // certificates of later epochs.
+        if (info.isNonCeasing() && mempool.mapSidechains.at(scId).mBackwardCertificates.size() > 0)
+        {
+            const uint256& topQualHash = mempool.mapSidechains.at(scId).GetTopQualityCert()->second;
+            const CScCertificate& certTopQual = mempool.mapCertificate[topQualHash].GetCertificate();
+            const auto map_it = mapCumtreeHeight.find(certTopQual.endEpochCumScTxCommTreeRoot.GetLegacyHash());
+            
+            if (map_it == mapCumtreeHeight.end())
+            {
+                // TODO: should we set an invalid height or return false?
```
alsala, line 2001: See comment above

src/txmempool.cpp
```diff
@@ -1556,6 +1586,85 @@ bool CTxMemPool::checkIncomingTxConflicts(const CTransaction& incomingTx) const
...
+        // (x1,y1) > (x2,y2) if x1==x2 and y1 > y2
+
+        // retrieve referencedHeight
+        const CScCertificate& mapCert = mapCertificate.at(cert_it.second).GetCertificate();
+        if (mapCert.quality != 0)
+            return true; // isNonCeasing() == false
+
+        assert(mapCert.epochNumber > prevEpoch);
+
+        int mcertHeight = mapCumtreeHeight.at(mapCert.endEpochCumScTxCommTreeRoot.GetLegacyHash());
```
alsala, line 1643: Are we sure at this point that the entry is here? In other occurrences (for instance in `GetSidechain()` around line 1999 we use find() and a check on result, but the situation should be the same.

src/sc/sidechain.cpp
```diff
@@ -57,11 +57,138 @@ void Sidechain::ClearSidechainsFolder()
...
+    int certWindowEndHeight   = isNonCeasing() ? INT_MAX : GetCertSubmissionWindowEnd(certEpoch);
+
+    if (inclusionHeight < certWindowStartHeight || inclusionHeight > certWindowEndHeight)
+    {
+        return error("%s():%d - ERROR: certificate cannot be accepted, cert received outside safeguard\n",
+            __func__, __LINE__);
+    }
+
+    // Check ordering of references
+    if (isNonCeasing()) {
```
alsala, line 166: Maybe the code of `CheckCertTiming() `method could be simplified a bit by splitting in if (isNotCeasing) {... } else {...}. In this way also traces could be customized.

src/sc/sidechain.cpp
```diff
@@ -266,23 +400,33 @@ bool Sidechain::checkTxSemanticValidity(const CTransaction& tx, CValidationState
                     CValidationState::Code::INVALID, "sidechain-sc-creation-invalid-constant");
         }
 
-        if (sc.wCeasedVk.is_initialized() )
+        if (sc.withdrawalEpochLength != 0)
```
alsala, line 403: Consider using the same `CSidechain::isNonCeasingSidechain() `method for checking this (via if / else). The condition could be set at the start of the current loop to a helper local bool var.

src/coins.cpp
```diff
@@ -1824,24 +1875,43 @@ bool CCoinsViewCache::UpdateSidechain(const CScCertificate& cert, CBlockUndo& bl
...
+    }
 
     currentSc.lastTopQualityCertHash            = certHash;
     currentSc.lastTopQualityCertReferencedEpoch = cert.epochNumber;
     currentSc.lastTopQualityCertQuality         = cert.quality;
     currentSc.lastTopQualityCertBwtAmount       = bwtTotalAmount;
     currentSc.lastTopQualityCertView            = CScCertificateView(cert, currentSc.fixedParams);
 
+    if (currentSc.isNonCeasing()) {
+        currentSc.lastInclusionHeight = blockHeight != -1 ? blockHeight : GetHeight()+1;
```
alsala, line 1906: Pls check comment on coins.h: here the -1 default value passed by the caller could be avoided probably

src/coins.h
```diff
@@ -697,16 +697,14 @@ class CCoinsViewCache : public CCoinsViewBacked
...
-    CValidationState::Code IsCertApplicableToState(const CScCertificate& cert, bool* banSenderNode = nullptr) const;
+    CValidationState::Code IsCertApplicableToState(const CScCertificate& cert, bool checkMempool, bool* banSenderNode = nullptr) const;
 
     CValidationState::Code CheckEndEpochCumScTxCommTreeRoot(
-        const CSidechain& sidechain, int epochNumber, const CFieldElement& endCumScTxCommTreeRoot) const;
+        const CSidechain& sidechain, int epochNumber, const CFieldElement& endCumScTxCommTreeRoot, int &referencedHeight) const;
 
-    bool CheckCertTiming(const uint256& scId, int certEpoch) const;
-    bool UpdateSidechain(const CScCertificate& cert, CBlockUndo& blockUndo);
+    bool UpdateSidechain(const CScCertificate& cert, CBlockUndo& blockUndo, int nHeight = -1);
```
alsala, line 707: The default -1 value for nHeight is used only in gtests, right? In this case, consider setting this value to `view.GetHeight()+1` directly in all gtest calls, in this way the implementation does not have to handle this -1 value on behalf of gtest only.

src/sc/sidechain.h
```diff
@@ -50,13 +51,18 @@ class CSidechainEvents {
...
+
 public:
     CSidechain():
         creationBlockHeight(-1), creationTxHash(),
         pastEpochTopQualityCertView(), lastTopQualityCertView(), lastTopQualityCertHash(),
         lastTopQualityCertReferencedEpoch(CScCertificate::EPOCH_NULL),
         lastTopQualityCertQuality(CScCertificate::QUALITY_NULL), lastTopQualityCertBwtAmount(0),
-        balance(0), maxSizeOfScFeesContainers(-1) {}
+        balance(0), maxSizeOfScFeesContainers(-1),
+        lastReferencedHeight(-1), lastInclusionHeight(-1) {}
```
alsala, line 65: Pls add these new data members to the `ToString()` method

src/coins.cpp
```diff
@@ -1263,27 +1258,56 @@ CValidationState::Code CCoinsViewCache::IsCertApplicableToState(const CScCertifi
...
+    {
+        const auto map_it = mapCumtreeHeight.find(endEpochCumScTxCommTreeRoot.GetLegacyHash());
+        if (map_it == mapCumtreeHeight.end()) {
+            LogPrintf("%s():%d - ERROR: cannot find commTreeRoot for certificate epoch %d (height %d)\n",
+                __func__, __LINE__, epochNumber, chainActive.Height());
+            return CValidationState::Code::INVALID;
+        }
+        referencedHeight = map_it->second;
+
+        if (referencedHeight <= sidechain.lastReferencedHeight)
```
alsala, line 1287: In case of a chain reorg, depending on the number of reverted blocks and certs, can this check fail when the reverted certificates gets back to the mempool? Probably a py test for this would be useful.

src/undo.h
```diff
@@ -162,6 +163,11 @@ struct CSidechainUndoData
     // CEASED_CERTIFICATE_DATA
     std::vector<CTxInUndo> ceasedBwts;
 
+    // NONCEASING_CERT_DATA
+    int prevInclusionHeight;
```
alsala, line 167: Pls add these new data members to `ToString()` method below

src/wallet/wallet.cpp
```diff
@@ -2070,33 +2070,31 @@ int CWallet::ScanForWalletTransactions(CBlockIndex* pindexStart, bool fUpdate)
...
+
+                bool bTopQualityCert = visitedScIds.count(itCert->GetScId()) == 0 || sidechain.isNonCeasing();
+                visitedScIds.insert(itCert->GetScId());
+
+                int bwtMaxDepth = sidechain.GetCertMaturityHeight(itCert->epochNumber, pindex->nHeight) - nHeight;
 
                 if (AddToWalletIfInvolvingMe(*itCert, &block, bwtMaxDepth, fUpdate))
                 {
                     ret++;
-                    if (fUpdate)
```
alsala, line 2084: Why `if (fUpdate)` has been removed? I see that it is mostly set as true by callers but it is set=false in the 'importwallet' rpc command

src/main.h
```diff
@@ -134,6 +134,8 @@ extern CCriticalSection cs_main;
 extern CTxMemPool mempool;
 typedef boost::unordered_map<uint256, CBlockIndex*, ObjectHasher> BlockMap;
 extern BlockMap mapBlockIndex;
+typedef boost::unordered_map<uint256, int, ObjectHasher> ScCumTreeRootMap;
+extern ScCumTreeRootMap mapCumtreeHeight;
```
alsala, line 138: This is an ever growing map, is there any way to do some optimization? For instance, do we need all heights or can we live with, say, latest 2000 block data?

alsala (changes requested):

src/bitcoin-tx.cpp
drgora: These were added to avoid undefined referenced in `zen-tx` when `pcoinsTip` was referenced in the sidechain compilation unit. Such changes are now superseded, so these lines were removed.

src/sc/sidechain.cpp
drgora: Yes, removed in branch [ap/nonCeasingSC](https://github.com/HorizenOfficial/zen/tree/ap/nonCeasingSC)

src/sc/sidechaintypes.cpp
drgora: Good point, done in branch [ap/nonCeasingSC](https://github.com/HorizenOfficial/zen/tree/ap/nonCeasingSC), commit 294ac79ce6e288b66989e83b90cbcde53a54b2f3

src/txmempool.h
drgora: Done, thanks

src/txmempool.cpp
drgora: Done with commit 294ac79ce6e288b66989e83b90cbcde53a54b2f3

src/txmempool.cpp
drgora: Well, we are sure that this is a non ceasing sc, so there should be an entry, unless some referenced past blocks were invalidated. This can happen with commit 777a4efe60a61ad24940db040b81c68d5c33065d, so now we use a `find` here. Thanks.

src/sc/sidechain.cpp
drgora: Done in branch  [ap/nonCeasingSC](https://github.com/HorizenOfficial/zen/tree/ap/nonCeasingSC)

src/sc/sidechain.cpp
drgora: Done with 294ac79ce6e288b66989e83b90cbcde53a54b2f3

src/coins.cpp
drgora: Done with 294ac79ce6e288b66989e83b90cbcde53a54b2f3

src/coins.h
drgora: Done with 294ac79ce6e288b66989e83b90cbcde53a54b2f3

src/sc/sidechain.h
drgora: Good point, done with b52d36e061c80799e6073e5ac2fc8816a8bb0fb0

src/undo.h
drgora: Done with 294ac79ce6e288b66989e83b90cbcde53a54b2f3

src/main.h
drgora: That's a good point. This is indeed an ever growing map. Right now we just store info on blocks whose height is compatible with non ceasing sidechains, but this might be optimized further as you are suggesting (even though the map itself should be quite small in memory in absolute an relative terms, compared to the overall memory usage of zend).

I'd suggest to leave this as it is for now, an elaborate on optimization options in future.

src/txmempool.cpp
drgora: This case has been addressed in a8c6b3277dd25cf2621976901d7351d1d15492d0

PR #464, opened 2022-08-03
Ap/python log improv

a-petrini:
This pull request improves the readability of logs generated by Python tests by:
- changing the order of parameters passed to assert_equal() to follow a more natural ordering
- adding support for colored log messages printed on terminal

### DIFF
--- a/qa/rpc-tests/test_framework/util.py
+++ b/qa/rpc-tests/test_framework/util.py
@@ -572,27 +572,28 @@ def colorize(color: str, text: str) -> str:
         return text
 
     N_C = "\033[0m"
-    colorshortcuts = ['e', 'r', 'g', 'y', 'b', 'p', 'c', 'n']
+    colorshortcuts = ['e', 'r', 'g', 'y', 'b', 'p', 'c']
     color = color[:1]   # just get the first char of 'color' input param
 
-    if color not in colorshortcuts:
+    if (color == 'n') or (color not in colorshortcuts):
         return text
 
-    COLORS = dict(zip(colorshortcuts, list(range(90, 97)) + [0]))
+    COLORS = dict(zip(colorshortcuts, range(90, 97)))
     return "\033[%d;1m" % COLORS[color] + text + N_C
 
-def mark_logs(msg, nodes, debug = 0, strip_escape = True):
+def strip_escape_seq(text: str) -> str:
+    """
+    Returns the string 'text' stripped from the escape sequences added by colorize()
+    """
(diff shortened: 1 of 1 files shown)


qa/rpc-tests/test_framework/util.py
```diff
@@ -443,7 +443,7 @@ def random_transaction(nodes, amount, min_fee, fee_increment, fee_variants):
 
     return (txid, signresult["hex"], fee)
 
-def assert_equal(expected, actual, message = ""):
+def assert_equal(actual, expected, message = ""):
```
ptagl, line 446: As a personal preference, I like the change that switched `actual` and `expected` parameters. I guess we also need at this point to do the same switch for the log string, so that it gets printed like:

`current value <actual>, expected <expected>`

As an alternative, we might also consider looking at the changes made by [ZCash](https://github.com/zcash/zcash/blob/f84a57e71bded0774bcb63e8c205c6db6ce442e1/qa/rpc-tests/test_framework/util.py#L612) and [Bitcoin](https://github.com/bitcoin/bitcoin/blob/d36bec9b3bf5c8730f14485944a035378451ae01/test/functional/test_framework/util.py#L54) to the same function.

To keep uniformity with the other util functions, I would vote to partially integrate the change by Bitcoin, in particular:
- renaming the two input parameters as `thing1` and `thing2`
- changing the debug string as `failed <thing1> == <thing2>`

drgora (approved):
This looks good and useful to me, thanks!

qa/rpc-tests/test_framework/util.py
drgora: I definitely do not have a strong opinion on this, but I personally like the log message to point out explicitly which value is expected and which one is observed, so I'd say my weak preference goes to the first option.

a-petrini: Commit e8ccdb14cc643288b109ad004751f700ada9eef8 modifies the returned string of `assert_equal(actual, expected)` function, so that values are reported in the same order of the input parameters:

`current value [left]: <actual>, expected [right]: <expected>`

Also the same commit modifies the `mark_logs()` function for supporting strings that already underwent the `colorize()` function, without propagating the ANSI escape sequences to the `debug.log` files. This is necessary for supporting fancy multicolored strings, i.e.:

`mark_logs("This is a red colored string embellished with some " + colorize('b', "blue words."), nodes, debug = 0, 'r')`

ptagl (approved):

PR #470, opened 2022-09-26
Gp/depends building fix

JackPiri:
Depends building bug fix.
In case of repo with local path too long, a makefile error was prompted; now the construction of command arguments results in shorter string.


a-petrini (approved):
Looks good

kamil871 (approved):

ptagl (approved):

PR #478, opened 2022-10-07
added sc v2 support for scFees list

a-petrini:


src/sc/sidechain.cpp
```diff
@@ -713,17 +713,30 @@ int CSidechain::getNumBlocksForScFeeCheck()
...
-        maxSizeOfScFeesContainers = numBlocks / epochLength;
-        if (maxSizeOfScFeesContainers == 0 || (numBlocks % epochLength != 0) )
-        {
-            maxSizeOfScFeesContainers++;
+    if (maxSizeOfScFeesContainers == -1) {
+        maxSizeOfScFeesContainers = getNumBlocksForScFeeCheck();
+
+        if (!isNonCeasing()) {
+            int epochLength = fixedParams.withdrawalEpochLength;
+            int numBlocks = getNumBlocksForScFeeCheck();
```
AndreaLanfranchi: Shouldn't `epochLength` and `numBlocks` be `const` ?
Also wouldn't be better to instantiate `numBlocks` in the `else` branch ? (it has no use in the `if` branch

src/sc/sidechain.cpp
```diff
@@ -713,17 +713,30 @@ int CSidechain::getNumBlocksForScFeeCheck()
...
+            if (epochLength == 0) {
+                maxSizeOfScFeesContainers = 1;
+            } else {
+                maxSizeOfScFeesContainers = numBlocks / epochLength;
```
AndreaLanfranchi: Is this safe ? Any chance `epochLength` assumes a negative value ?

src/sc/sidechain.cpp
```diff
@@ -713,17 +713,30 @@ int CSidechain::getNumBlocksForScFeeCheck()
...
+                }
+            }
+        }
+        else {
+            // As defined in chainparams.cpp, maxSizeOfScFeesContainers is 200 for mainnet and testnet,
+            // and 10 for regtest. For regtest, this value can be overridden with the command line option
+            // -blocksforscfeecheck=<n>. In this case, we provide a safe guard of five slots for the scFees
+            // size. This value has been chosen as it was the max possible size obtainable by the previous
+            // implementation.
+            if (maxSizeOfScFeesContainers < 5)
```
AndreaLanfranchi: I find semantically better 
```
maxSizeOfScFeesContainers = std::max(maxSizeOfScFeesContainers, 5)
```

src/sc/sidechain.cpp
```diff
@@ -757,61 +778,139 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView)
...
+                    if (scFees_v2.empty())
+                        break;
+
+                    const auto& entry = scFees_v2.front();
+                    LogPrint("sc", "%s():%d - popping f=%d, m=%d, h=%d from list, as scFees maxsize has been reached\n",
+                        __func__, __LINE__, entry.forwardTxScFee, entry.mbtrTxScFee, entry.submissionHeight);
+                    scFees_v2.pop_front();
+                    delta--;
+                }
+            }
```
AndreaLanfranchi: The `if (delta > 0)` is not needed here imho. Having no `else` branch the `while (delta > 0)` is enough

src/sc/sidechain.cpp
AndreaLanfranchi: Also ... what is that magic constant 5 ? Can we give a meaningful value name using `static constexpr` somewhere ?

src/sc/sidechain.cpp
```diff
@@ -757,61 +778,139 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView)
...
-    int delta = scFees.size() - maxSizeOfScFeesContainers;
-    if (delta > 0)
-    {
-        // remove from the front as many elements are needed to be within the circular buffer size
-        // --
-        // usually this is just one element, but in regtest a node can set the max size via a startup option
-        // therefore such size might be lesser than scFee size after a node restart
-        while (delta > 0)
+        int delta = scFees.size() - maxSizeOfScFeesContainers;
+        if (delta > 0)
```
AndreaLanfranchi: The `if (delta > 0)` is not needed here imho. Having no `else` branch the `while (delta > 0)` is enough

src/sc/sidechain.h
```diff
@@ -170,10 +172,13 @@ class CSidechain {
...
+            READWRITE(scFees_v2);
+        else
+            READWRITE(scFees);
         if (ser_action.ForRead())
         {
-            if (!scFees.empty())
+            if (!scFees.empty() || !scFees_v2.empty())
             {
                 maxSizeOfScFeesContainers = getMaxSizeOfScFeesContainers();
             }
```
AndreaLanfranchi: What is this for ?
I thought a serialization op was not meant to change the state of the object.
Why we do change `maxSizeOfScFeesContainers` and, btw, we dont' serialize it ?

src/sc/sidechain.cpp
```diff
@@ -757,61 +778,139 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView)
...
+            while (delta > 0)
+            {
+                if (scFees.empty())
+                    break;
+
+                const auto& entry = scFees.front();
+                LogPrint("sc", "%s():%d - popping f=%d, m=%d from list, as scFees maxsize has been reached\n",
+                    __func__, __LINE__, entry.forwardTxScFee, entry.mbtrTxScFee);
+                scFees.pop_front();
+                delta--;
```
AndreaLanfranchi: Prefer pre-increments rather than post-increments when value is not immediately used. 

Rationale https://google.github.io/styleguide/cppguide.html#Preincrement_and_Predecrement

src/sc/sidechain.cpp
```diff
@@ -739,14 +752,22 @@ void CSidechain::InitScFees()
...
         assert(scFees.empty());
+        assert(scFees_v2.empty());
 
-        Sidechain::ScFeeData defaultData(
-            lastTopQualityCertView.forwardTransferScFee, lastTopQualityCertView.mainchainBackwardTransferRequestScFee);
-        scFees.push_back(defaultData);
+        if (!isNonCeasing()) {
+            Sidechain::ScFeeData defaultData(
+                lastTopQualityCertView.forwardTransferScFee, lastTopQualityCertView.mainchainBackwardTransferRequestScFee);
+            scFees.push_back(defaultData);
```
AndreaLanfranchi: Prefer use `emplace_back` instead of `push_back` in cases like this.
We save a temporary object in memory and constructor is done in-place within the container.

src/sc/sidechain.cpp
```diff
@@ -757,61 +778,139 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView)
...
-    // check if we are past the buffer size
-    int delta = scFees.size() - maxSizeOfScFeesContainers;
-    if (delta > 0)
-    {
-        // remove from the front as many elements are needed to be within the circular buffer size
-        // --
-        // usually this is just one element, but in regtest a node can set the max size via a startup option
-        // therefore such size might be lesser than scFee size after a node restart
-        while (delta > 0)
+        int delta = scFees.size() - maxSizeOfScFeesContainers;
```
AndreaLanfranchi: This is dangerous for various reasons:

- sign mismatch amongst operators
- implicit narrowing conversion from size_t to int

Raising warnings to errors in compiler causes a fail cause what the compiler actually does is `(int)(scFees.size() - ((size_t)maxSizeOfScFeesContainers))` which may cause an underflow (e.g. 1-5) which can return a "meaningful result" only in some cases.
See https://godbolt.org/z/37eTsafrf

AndreaLanfranchi (changes requested):

src/sc/sidechain.cpp
a-petrini: Non negativity of epoch length is checked in `Sidechain::checkTxSemanticValidity()` before sidechain creation, hence in the current configuration `epochLenght` cannot be negative. Things may change if in the future we would allow `epochLenght` to assume negative values for managing special cases. I propose putting a preemptive assert.

src/sc/sidechain.cpp
a-petrini: The value 5 is derived from the default values set in `chainparams.cpp`. Specifically, for regtest `nScNumBlocksForScFeeCheck` (the values returned by `CSidechain::getNumBlocksForScFeeCheck()`) is set to 10, and the withdrawal epoch length can range from 2 to 4032 (`nScMinWithdrawalEpochLength` and `nScMaxWithdrawalEpochLength` also in `chainparams.cpp`).

These values allow the `scFees` list size to range from 1 to 5, as evaluated by `numBlocks / epochLength` (and the subsequent `if` block).

I was not sure whether all these details were meaningful, so I skipped them in the comment altogether, but I realize that by looking at the value it is not clear where it comes from. I doubt that a meaningful name could help here, so instead I suggest I could update the comment, so that its meaning should not be lost.

AndreaLanfranchi: Great insight .... thank you. 
So let me ask ... can't we have a function deriving the max value instead of setting as a constant ?

src/sc/sidechain.h
a-petrini: This looks super weird and caused an issue during tests, as I originally forgot to check non-emptiness of `scFees_v2`.

It seems that `maxSizeOfScFeesContainers` is not serialized, hence it must be recovered and re-evaluated when a node is reactivated after a shutdown, as this variable is initialized with a proper value only when a sc creation transaction is received (see `CSidechain::InitScFees()`).

Given this context, it is possible that `scFees` are correctly emplaced in the lists when deserializing a new block, but `maxSizeOfScFeesContainers` is still set as -1.

src/sc/sidechain.cpp
a-petrini: Investigating this issue, and `CSidechain::UpdateScFees()` already assumes that `maxSizeOfScFeesContainers` is positive (verified by assertion). One possible unsafe solution would be to cast `scFees.size()` to int64_t before doing the subtraction, given that `maxSizeOfScFeesContainers` cannot be greater than 4032 and the size of `scFees` list is not expected to grow bigger than INTMAX element when receiving a block.

src/sc/sidechain.cpp
a-petrini: While I like the idea of writing a function for this, I think that moving the evaluation logic inside a function adds more overhead than writing the constant with a brief explanation in the comment.

Moving the evaluation of this minimal value inside a function involves performing the division `CSidechain::getNumBlocksForScFeeCheck() / params.nScMinWithdrawalEpochLength`, that for mainnet and testnet is equal to 2, so the `std::max` will always return 200.
All in all, checking the minimum size of `maxSizeOfScFeesContainers` only makes sense for regtest, because for mainnet and testnet its value is assigned at the beginning of the very same function, with no possibility to set something different than those imposed by default in `chainparams.cpp`.

src/sc/sidechain.cpp
a-petrini: Both comments fixed in f0bab99cd7559a9bab4c8dc65103df6eea2b4b76

src/sc/sidechain.cpp
a-petrini: Assertion added in f0bab99cd7559a9bab4c8dc65103df6eea2b4b76

src/sc/sidechain.cpp
a-petrini: `if..then` replaced by `std::max`, and comment improved in f0bab99cd7559a9bab4c8dc65103df6eea2b4b76

src/sc/sidechain.cpp
a-petrini: fixed in f0bab99cd7559a9bab4c8dc65103df6eea2b4b76

src/sc/sidechain.cpp
a-petrini: Fixed in f0bab99cd7559a9bab4c8dc65103df6eea2b4b76

src/sc/sidechain.cpp
a-petrini: All occurrences of `delta--` replaced with `--delta` in f0bab99cd7559a9bab4c8dc65103df6eea2b4b76

src/sc/sidechain.cpp
a-petrini: All occurrences of `push_back` replaced with `emplace_back` in f0bab99cd7559a9bab4c8dc65103df6eea2b4b76

src/sc/sidechain.h
AndreaLanfranchi: So let me understand ... we're using serialization here as a constructor ?

src/sc/sidechain.cpp
```diff
@@ -757,12 +760,12 @@ void CSidechain::InitScFees()
         if (!isNonCeasing()) {
             Sidechain::ScFeeData defaultData(
                 lastTopQualityCertView.forwardTransferScFee, lastTopQualityCertView.mainchainBackwardTransferRequestScFee);
-            scFees.push_back(defaultData);
+            scFees.emplace_back(defaultData);
```
AndreaLanfranchi: Hmmm ... sorry that's not what I meant.
By using emplace_back in this way your constructing a temporary object and then in the container is copy constructed the final object.
It should be

`scFees.emplace_back(lastTopQualityCertView.forwardTransferScFee, lastTopQualityCertView.mainchainBackwardTransferRequestScFee)
`
This forwards constructor arguments to the container which will construct object in place

AndreaLanfranchi (changes requested):

src/sc/sidechain.cpp
AndreaLanfranchi: Yet a small note on this
In both loops I see 

```
        int delta = scFees.size() - maxSizeOfScFeesContainers;
        while (delta > 0)
        {
            if (scFees.empty())
                break;

            const auto& entry = scFees.front();
            LogPrint("sc", "%s():%d - popping f=%d, m=%d from list, as scFees maxsize has been reached\n",
                __func__, __LINE__, entry.forwardTxScFee, entry.mbtrTxScFee);
            scFees.pop_front();
            --delta;
        }
```

As a result the inner check `if (scFees.empty())` can really never happen : we enter the while loop only if vector's size is >  maxSizeOfScFeesContainers and by consequence we never empty the vector in the loop
As a suggestion I'd rewrite as : 

```
        const size_t max_size { static_cast<size_t>(maxSizeOfScFeesContainers) };
        while (scFees.size() > max_size)
        {
            const auto& entry = scFees.front();
            LogPrint("sc", "%s():%d - popping f=%d, m=%d from list, as scFees maxsize has been reached\n",
                __func__, __LINE__, entry.forwardTxScFee, entry.mbtrTxScFee);
            scFees.pop_front();
        }

```

Which is more readable and clearly expresses the intention without delta and substractions

src/sc/sidechain.cpp
```diff
@@ -773,45 +797,119 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView)
...
-    for (const auto& entry : scFees)
-    {
-        std::cout << "[" << std::setw(2) << entry.forwardTxScFee
-                  << "/" << std::setw(2) << entry.mbtrTxScFee << "]";
+    if (!isNonCeasing()) {
+        for (const auto& entry : scFees) {
+            std::cout   << "[" << std::setw(2) << entry.forwardTxScFee
+                        << "/" << std::setw(2) << entry.mbtrTxScFee << "]";
+        }
+        std::cout << std::endl;
```
AndreaLanfranchi: Only one `std::cout << std::endl;` is actually needed just before function end

src/sc/sidechain.cpp
```diff
@@ -773,45 +797,119 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView)
...
+            }
+        }
+        // On the other hand, if we found a scFeeData element for the current height, we just update it with lower
+        // ft and/or mbtr fee rates. All the checks on the container size and element age have been performed
+        // at the moment of the element insertion, hence their are not required here
+        else {
+            if (ftScFee < scFeeIt->forwardTxScFee)
+                scFeeIt->forwardTxScFee = ftScFee;
+            if (mbtrScFee < scFeeIt->mbtrTxScFee)
+                scFeeIt->mbtrTxScFee = mbtrScFee;
```
AndreaLanfranchi: We have semantics available which match the description

```suggestion
            scFeeIt->forwardTxScFee = std::min(scFeeIt->forwardTxScFee, ftScFee);
            scFeeIt->mbtrTxScFee = std::min(scFeeIt->mbtrTxScFee, mbtrScFee);
```

AndreaLanfranchi (changes requested):

src/sc/sidechain.cpp
```diff
@@ -713,17 +713,33 @@ int CSidechain::getNumBlocksForScFeeCheck()
...
-            maxSizeOfScFeesContainers++;
+    if (maxSizeOfScFeesContainers == -1) {
+        maxSizeOfScFeesContainers = getNumBlocksForScFeeCheck();
+
+        if (!isNonCeasing()) {
+            const int epochLength = fixedParams.withdrawalEpochLength;
+            assert(epochLength > 0);
+            if (epochLength == 0) {
+                maxSizeOfScFeesContainers = 1;
+            } else {
```
AndreaLanfranchi: `assert`  voids the following `if`.
If we assert for `epochLength > 0`  why we implement a case when `epochLenght == 0` ?

src/sc/sidechain.cpp
```diff
@@ -713,17 +713,33 @@ int CSidechain::getNumBlocksForScFeeCheck()
...
+                const int numBlocks = getNumBlocksForScFeeCheck();
+                maxSizeOfScFeesContainers = numBlocks / epochLength;
+                if (maxSizeOfScFeesContainers == 0 || (numBlocks % epochLength != 0) )
+                {
+                    maxSizeOfScFeesContainers++;
+                }
```
AndreaLanfranchi: Easy way to round up (ceil) integers dvisions (assuming operands are positive)

```suggestion
                const int numBlocks = getNumBlocksForScFeeCheck();
                maxSizeOfScFeesContainers = (numBlocks + epochLength - 1) / epochLength;
```
We save modulus operator to compute the remainder and an if branch
See https://godbolt.org/z/4nKq9sWbd

src/sc/sidechain.cpp
```diff
@@ -773,45 +797,119 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView)
...
+            // considered - this also covers ceasable sidechain, as certificate submission
+            // happens at regular intervals
+            auto& entry = scFees_v2.front();
+            while (entry.submissionHeight <= blockHeight - getMaxSizeOfScFeesContainers())
+            {
+                LogPrint("sc", "%s():%d - popping f=%d, m=%d, h=%d from list, as entry is too old\n",
+                    __func__, __LINE__, entry.forwardTxScFee, entry.mbtrTxScFee, entry.submissionHeight);
+                scFees_v2.pop_front();
+                entry = scFees_v2.front();
+            }
```
AndreaLanfranchi: This loop (reading the implementation **and** the description) strictly requires list `scFees_v2` is sorted by submission height (ascending) but I'm not sure it is so.
Can you clarify ?

AndreaLanfranchi (changes requested):

src/sc/sidechain.cpp
AndreaLanfranchi: Also `blockHeight - getMaxSizeOfScFeesContainers()` should be stored into a const variable to prevent recompute on every loop

src/sc/sidechain.cpp
a-petrini: My bad. 2522a6032b25bc34777cfd408f489058fb189cee and d39a7f376cd683ff02304198fef02b53898b8375 improved the object emplacement in the lists.

src/sc/sidechain.cpp
a-petrini: Fixed in 2522a6032b25bc34777cfd408f489058fb189cee

src/sc/sidechain.cpp
a-petrini: Agree. Fixed in 2522a6032b25bc34777cfd408f489058fb189cee

src/sc/sidechain.cpp
a-petrini: `if` condition removed in 2522a6032b25bc34777cfd408f489058fb189cee, as no longer useful

src/sc/sidechain.cpp
a-petrini: This whole part has been updated and rewritten in 2522a6032b25bc34777cfd408f489058fb189cee and updated in 5b469af6a62aee85c0dfebacb847b343c38d0e55 to manage the edge case where -blocksforscfeecheck is 0

src/sc/sidechain.cpp
a-petrini: This section has been rewritten in 2522a6032b25bc34777cfd408f489058fb189cee to exploit the `remove_if` method included in STL. This solution does not require the list to be ordered.

src/sc/sidechain.cpp
a-petrini: This part has been updated in 2522a6032b25bc34777cfd408f489058fb189cee

src/sc/sidechain.cpp
a-petrini: `delta` variable has been removed in 2522a6032b25bc34777cfd408f489058fb189cee

src/sc/sidechain.cpp
```diff
@@ -757,61 +769,119 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView)
...
+
+            // Also purge all elements submitted within a block that now is too old to be considered
+            scFees_v2.remove_if([blockHeight, this](Sidechain::ScFeeData_v2 entry) {
+                const bool testCondition = (entry.submissionHeight <= blockHeight - this->maxSizeOfScFeesContainers);
+                if (testCondition)
+                    LogPrint("sc", "%s():%d - popping f=%d, m=%d, h=%d from list, as entry is too old\n",
+                        __func__, __LINE__, entry.forwardTxScFee, entry.mbtrTxScFee, entry.submissionHeight);
+                return testCondition;
+            });
```
AndreaLanfranchi: Humble suggestion (capture as little as possible in lambdas - especially `this` if not needed)
For all the rest LGTM
Thank you !!!

```suggestion
            // Also purge all elements submitted within a block that now is too old to be considered
            const auto threshold{blockHeight - this->maxSizeOfScFeesContainers};
            scFees_v2.remove_if([threshold](Sidechain::ScFeeData_v2 entry) {
                const bool testCondition = (entry.submissionHeight <= threshold);
                if (testCondition)
                    LogPrint("sc", "%s():%d - popping f=%d, m=%d, h=%d from list, as entry is too old\n",
                        __func__, __LINE__, entry.forwardTxScFee, entry.mbtrTxScFee, entry.submissionHeight);
                return testCondition;
            });
```

AndreaLanfranchi (approved):

src/sc/sidechain.cpp
a-petrini: fixed in c3fa17e38bb8677927f9b87bf564d3f8ec9954fc

src/sc/sidechaintypes.h
```diff
@@ -483,6 +483,26 @@ typedef struct sScFeeData_tag
...
+    int submissionHeight;
+    ScFeeData_v2() : ScFeeData(), submissionHeight(-1) {};
+    ScFeeData_v2(CAmount f, CAmount m, int submHeight) : ScFeeData(f, m), submissionHeight(submHeight) {};
+
+    ADD_SERIALIZE_METHODS;
+
+    template <typename Stream, typename Operation>
+    inline void SerializationOp(Stream& s, Operation ser_action, int nType, int nVersion) {
+        READWRITE(forwardTxScFee);
+        READWRITE(mbtrTxScFee);
```
drgora: Just a small tip here, feel free to ignore. Have you considered something like the following?
```suggestion
        ScFeeData::SerializationOp(s, ser_action, nType, nVersion);
```
To me this also seems more in line with `ScFeeData_v2`'s constructors.

src/sc/sidechaintypes.h
```diff
@@ -483,6 +483,26 @@ typedef struct sScFeeData_tag
...
+        READWRITE(submissionHeight);
+    }
+
+    inline bool operator==(const ScFeeData_v2& rhs) const
+    {
+        return (forwardTxScFee == rhs.forwardTxScFee && mbtrTxScFee == rhs.mbtrTxScFee && submissionHeight == rhs.submissionHeight);
```
drgora: Similar to the comment above, how about
```suggestion
        return ScFeeData::operator=(rhs) && submissionHeight == rhs.submissionHeight;
```

src/rpc/blockchain.cpp
```diff
@@ -1654,12 +1654,25 @@ bool FillScRecordFromInfo(const uint256& scId, const CSidechain& info, CSidechai
...
-            sf.push_back(o);
+        // v2 non-ceasable sc
+        if (info.isNonCeasing()) {
+            for(const auto& entry: info.scFees_v2)
+            {
+                UniValue o(UniValue::VOBJ);
+                o.pushKV("forwardTxScFee", ValueFromAmount(entry.forwardTxScFee));
+                o.pushKV("mbtrTxScFee", ValueFromAmount(entry.mbtrTxScFee));
+                o.pushKV("submissionHeight", entry.submissionHeight);
+                sf.push_back(o);
```
drgora: I'd suggest to 
```suggestion
                sf.push_back(std::move(o));
```
here, to save some mem copies.

AndreaLanfranchi: That would be a wise hint for a method on Univalue implementing `&&` in the 	`push_back` signature.
I fear it'd have no effect here.
(Just my two cents)

a-petrini: Yes, I also changed all the occurrences in the same function in cb80200cd4274e56654dd8b0860c37dcb80cc076

src/sc/sidechaintypes.h
a-petrini: Thanks, I updated both function as suggested in cb80200cd4274e56654dd8b0860c37dcb80cc076

qa/rpc-tests/sc_ft_and_mbtr_fees_update.py
```diff
@@ -0,0 +1,591 @@
...
+        self.sync_all()
+
+        ### Create a network split:
+        ### 1. disconnect nodes 1 and 2, so that the the nets are nodes 0, 1 and 2, 3
+        ### 2. in the first round, node 0 sends many small txs to sc; save the node balances
+        ### 3. node 2 mines block
+        ### 4. rejoin the network
+        ### 5. mined block should be broadcasted to nodes 0 and 1, but transactions shouldn't
+        ### This process should repeat until we enter epoch 2
+        nodesToBeGen = self.nodes[0].getscinfo(scid)['items'][0]['endEpochHeight'] - self.nodes[0].getblockcount() + 1
```
drgora: I believe there is a typo. `nodesToBeGen` should actually be something like `blocksToBeGen`, right?

qa/rpc-tests/sc_ft_and_mbtr_fees_update.py
```diff
@@ -0,0 +1,591 @@
...
+            self.split_network()
+
+            # 2.
+            if loopfirstround:
+                mark_logs_temp(f"Node 0 adds {NUM_TXS} transactions to nodes 0 and 1 mempools", self.nodes, DEBUG_MODE, color='g')
+                initial_balances = [self.nodes[x].getbalance() for x in range(NUMB_OF_NODES)]
+                ft_address_1 = self.nodes[1].getnewaddress() # = mc_return_address
+                forwardTransferOuts = [{'toaddress': address, 'amount': Decimal(float(FT_SC_FEE) + 0.05), "scid": scid, "mcReturnAddress": ft_address_1}]
+                for _ in range(NUM_TXS):
+                    # self.nodes[0].sendtoaddress(ft_address_1, 0.03)
```
drgora: Should we remove this line entirely?

drgora (approved):
Looks good to me, thanks

qa/rpc-tests/sc_ft_and_mbtr_fees_update.py
a-petrini: Yep, my mistake. Corrected in 609432fde5cb879e3941d85880f9fe42557168f9

qa/rpc-tests/sc_ft_and_mbtr_fees_update.py
a-petrini: Line removed in 609432fde5cb879e3941d85880f9fe42557168f9

kamil871 (approved):

src/sc/sidechain.cpp
```diff
@@ -713,20 +713,27 @@ int CSidechain::getNumBlocksForScFeeCheck()
...
+        if (!isNonCeasing()) {
+            const int numBlocks = getNumBlocksForScFeeCheck();
+            const int epochLength = fixedParams.withdrawalEpochLength;
+            assert(epochLength > 0);
+            maxSizeOfScFeesContainers = (numBlocks + epochLength - 1) / epochLength;
+            // CSidechain::getNumBlocksForScFeeCheck() may return 0 for regtest...
+            // This was managed in the same way in the old version
+            maxSizeOfScFeesContainers = std::max(maxSizeOfScFeesContainers, 1);
+        } else {
+            // As defined in chainparams.cpp, maxSizeOfScFeesContainers is 200 for mainnet and testnet,
```
JackPiri: I guess `maxSizeOfScFeesContainers` inside comment at line #726 should be changed to `ScNumBlocksForScFeeCheck` (or `nScNumBlocksForScFeeCheck`)

src/sc/sidechain.h
```diff
@@ -198,7 +200,8 @@ class CSidechain {
                (this->balance                                    == rhs.balance)                           &&
                (this->fixedParams                                == rhs.fixedParams)                       &&
                (this->mImmatureAmounts                           == rhs.mImmatureAmounts)                  &&
-               (this->scFees                                     == rhs.scFees);
+               (this->scFees                                     == rhs.scFees)                            &&
```
JackPiri: Can all `this->` be removed (as well as for `CSidechainEvents` `operator==` inside same file)?

src/sc/sidechain.cpp
```diff
@@ -713,20 +713,27 @@ int CSidechain::getNumBlocksForScFeeCheck()
...
-    //assert(epochLength > 0);
-    if (epochLength == 0) {
-        maxSizeOfScFeesContainers = 1;
-    } else {
-        maxSizeOfScFeesContainers = numBlocks / epochLength;
-        if (maxSizeOfScFeesContainers == 0 || (numBlocks % epochLength != 0) )
-        {
-            maxSizeOfScFeesContainers++;
+    if (maxSizeOfScFeesContainers == -1) {
+        if (!isNonCeasing()) {
```
JackPiri, line 718: What about moving `getNumBlocksForScFeeCheck` outside the if-else, being used in both branches?

```
const int numBlocks = getNumBlocksForScFeeCheck();
if (!isNonCeasing()) {
...
} else {
...
}
```

src/undo.h
```diff
@@ -310,11 +317,20 @@ struct CSidechainUndoData
             res += strprintf("prevInclusionHeight=%d\n", prevInclusionHeight);
         }
 
-        res += strprintf("scFees.size()=%u\n", scFees.size());
-        for(const auto& entry: scFees)
-        {
-           res += strprintf("scFtFee=%d.%08d - ", entry.forwardTxScFee / COIN, entry.forwardTxScFee % COIN);
-           res += strprintf("scMbtrFee=%d.%08d\n", entry.mbtrTxScFee / COIN, entry.mbtrTxScFee % COIN);
+        if (contentBitMask & AvailableSections::NONCEASING_CERT_DATA) {
+            res += strprintf("scFees.size()=%u\n", scFees_v2.size());
```
JackPiri: I see in the v2-case the res variable would contain `submissionHeight` (conversely from non v2-case), but why not emphasizing it with `strprintf("scFees_v2.size()=%u\n", scFees_v2.size())`?

src/sc/sidechain.cpp
```diff
@@ -757,61 +769,120 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView)
...
+                    __func__, __LINE__, entry.forwardTxScFee, entry.mbtrTxScFee, entry.submissionHeight);
+                scFees_v2.pop_front();
+            }
+
+            // Also purge all elements submitted within a block that now is too old to be considered
+            const auto threshold{blockHeight - this->maxSizeOfScFeesContainers};
+            scFees_v2.remove_if([threshold](Sidechain::ScFeeData_v2 entry) {
+                const bool testCondition = (entry.submissionHeight <= threshold);
+                if (testCondition)
+                    LogPrint("sc", "%s():%d - popping f=%d, m=%d, h=%d from list, as entry is too old\n",
```
JackPiri: Maybe adding also threshold value can be useful when analyzing logs

```
LogPrint("sc", "%s():%d - popping f=%d, m=%d, h=%d from list, as entry is too old (t=%d)\n",
                        __func__, __LINE__, entry.forwardTxScFee, entry.mbtrTxScFee, entry.submissionHeight, threshold);
```

JackPiri (approved):
The PR looks good to me, thanks

src/undo.h
a-petrini: Keeping the same name for the new structure version should not break the compatibility. For v2 SC, the basic fee data will be listed under the same structure as v1, with just an additional field.

src/sc/sidechain.h
a-petrini: I am not sure why "this" has been used in a similar context, but I saw it elsewhere in the code. We might clean up this - HA! - in a future refactoring.

src/sc/sidechain.cpp
a-petrini: Comment updated, to better explain the origin of the value and avoid confusion, in c8b2941bc959c29cd0db69eb8df66aa425bccda8

src/sc/sidechain.cpp
a-petrini: Yes, seems reasonable. Added in c8b2941bc959c29cd0db69eb8df66aa425bccda8

src/sc/sidechain.cpp
a-petrini: Done in c8b2941bc959c29cd0db69eb8df66aa425bccda8

PR #479, opened 2022-10-11
Gp/size estimation z operations

### LINKED ISSUE #475: Size estimation for shielded transactions is not working correctly
https://github.com/HorizenOfficial/zen/blob/530675d3e8cadf50685a781265549446829568bb/src/wallet/rpcwallet.cpp#L6207-L6208

JackPiri:
Pull request associated to issue https://github.com/HorizenOfficial/zen/issues/475

### DIFF
--- a/qa/pull-tester/rpc-tests.sh
+++ b/qa/pull-tester/rpc-tests.sh
@@ -40,6 +40,7 @@ testScripts=(
   'wallet_protectcoinbase.py'
   'wallet_shieldcoinbase.py'
   'wallet_mergetoaddress.py'
+  'wallet_mergetoaddress_2.py'
   'wallet.py'
   'wallet_nullifiers.py'
   'wallet_1941.py'
--- a/qa/pull-tester/rpc-tests.sh
+++ b/qa/pull-tester/rpc-tests.sh
@@ -16,10 +16,6 @@ for i in "$@"; do
       EXTENDED="true"
       shift
       ;;
-    -specific=*)
-      SPECIFIC="${i#*=}"
-      shift
-      ;;
     -exclude=*)
       EXCLUDE="${i#*=}"
       shift
@@ -204,10 +200,6 @@ if [ ! -z "$EXCLUDE" ]; then
       fi
     done
   done
-fi
-
-if [ ! -z "$SPECIFIC" ]; then
-	testScripts=( "$SPECIFIC" )
 fi
 
 # split array into m parts and only run tests of part n where SPLIT=m:n
--- a/qa/pull-tester/rpc-tests.sh
+++ b/qa/pull-tester/rpc-tests.sh
@@ -16,6 +16,10 @@ for i in "$@"; do
       EXTENDED="true"
       shift
       ;;
+    -specific=*)
+      SPECIFIC="${i#*=}"
+      shift
+      ;;
     -exclude=*)
       EXCLUDE="${i#*=}"
       shift
@@ -200,6 +204,10 @@ if [ ! -z "$EXCLUDE" ]; then
       fi
     done
   done
+fi
+
+if [ ! -z "$SPECIFIC" ]; then
+	testScripts=( "$SPECIFIC" )
 fi
 
 # split array into m parts and only run tests of part n where SPLIT=m:n
--- a/qa/rpc-tests/wallet_mergetoaddress_2.py
+++ b/qa/rpc-tests/wallet_mergetoaddress_2.py
@@ -1,0 +1,223 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+import shutil
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.authproxy import JSONRPCException
+from test_framework.util import assert_equal, initialize_chain_clean, \
+    start_node, connect_nodes_bi, sync_blocks, sync_mempools, \
+    wait_and_assert_operationid_status
+
+import time
+from decimal import Decimal
+from distutils.dir_util import copy_tree
+
+class WalletMergeToAddressTest (BitcoinTestFramework):
+
+    def import_data_to_data_dir(self):
+        #importing datadir resource
+        import os
+        os.chdir(os.path.dirname(__file__))
+        resource_file = os.getcwd() + "/resources/wallet_mergetoaddress_2/test_setup_.zip"
+        import zipfile
+        with zipfile.ZipFile(resource_file, 'r') as zip_ref:
+            zip_ref.extractall(self.options.tmpdir)
+        copy_tree(self.options.tmpdir + "/test_setup_", self.options.tmpdir)
+        shutil.rmtree(self.options.tmpdir + "/test_setup_")
+
+    def generate_notes(self, quantity):
+        myzaddr0 = self.nodes[0].z_getnewaddress()
+        results = []
+        for i in range(quantity):
+            self.nodes[0].generate(1)
+            self.sync_all()
+            self.nodes[1].generate(100)
+            self.sync_all()
+            results.append(self.nodes[0].z_shieldcoinbase("*", myzaddr0, 0))
+            wait_and_assert_operationid_status(self.nodes[0], results[i]['opid'])
+        self.nodes[1].generate(1)
+        self.sync_all()
+        assert_equal(self.nodes[0].getblockcount(), quantity * (1 + 100) + 1)
+        assert_equal(self.nodes[0].getwalletinfo()["balance"] +
+                     self.nodes[0].getwalletinfo()["unconfirmed_balance"] +
+                     self.nodes[0].getwalletinfo()["immature_balance"], 0)
+
+    def send_transparent(self, quantity):
+        for i in range(quantity):
+            self.nodes[1].sendmany("", {self.nodes[0].listaddresses()[0]: 10}, 100)
+        self.nodes[1].generate(100)
+        self.sync_all
+
+    def get_merged_transaction(self, unspent_transactions_before_merge, unspent_transactions_after_merge):
+        txid = unspent_transactions_after_merge[0]["txid"]
+        for aft in range(len(unspent_transactions_after_merge)):
+            for bef in range(len(unspent_transactions_before_merge)):
+                if (unspent_transactions_after_merge[aft]["txid"] == unspent_transactions_before_merge[bef]["txid"]):
+                    break
+                if (bef == len(unspent_transactions_before_merge) - 1):
+                    txid = unspent_transactions_after_merge[aft]["txid"]
+        return self.nodes[2].gettransaction(txid)
+
+    def z_mergetoaddress_t2t(self, utxos_quantity):
+        mytaddr2 = self.nodes[2].listaddresses()[0]
+        listunspent_before_merge = self.nodes[2].listunspent(0)
+        balance_before_merge = self.nodes[2].z_getbalance(mytaddr2)
+        result = self.nodes[0].z_mergetoaddress(["ANY_TADDR"], mytaddr2, 0, utxos_quantity, 0)        
+        wait_and_assert_operationid_status(self.nodes[0], result['opid'], "success", "", 3600)
+        self.nodes[1].generate(100)
+        self.sync_all()
+        tx = self.get_merged_transaction(listunspent_before_merge, self.nodes[2].listunspent(0))
+        assert_equal(result["mergingUTXOs"], utxos_quantity)
+        assert_equal(result["mergingTransparentValue"], self.nodes[2].z_getbalance(mytaddr2) - balance_before_merge)
+        assert_equal(result["mergingNotes"], 0)
+        #no joinsplit is required because every input is transparent and every output is transparent
+        assert_equal(len(tx["vjoinsplit"]), 0)
+
+    def z_mergetoaddress_t2z(self, utxos_quantity):
+        if (len(self.nodes[2].z_listaddresses()) == 0):
+            myzaddr2 = self.nodes[2].z_getnewaddress()
+        else:
+            myzaddr2 = self.nodes[2].z_listaddresses()[0]
+        listunspent_before_merge = self.nodes[2].z_listunspent(0)
+        balance_before_merge = self.nodes[2].z_getbalance(myzaddr2)
+        result = self.nodes[0].z_mergetoaddress(["ANY_TADDR"], myzaddr2, 0, utxos_quantity, 0)        
+        wait_and_assert_operationid_status(self.nodes[0], result['opid'], "success", "", 3600)
+        self.nodes[1].generate(100)
+        self.sync_all()
+        tx = self.get_merged_transaction(listunspent_before_merge, self.nodes[2].z_listunspent(0))
+        assert_equal(result["mergingUTXOs"], utxos_quantity)
+        assert_equal(result["mergingTransparentValue"], self.nodes[2].z_getbalance(myzaddr2) - balance_before_merge)
+        assert_equal(result["mergingNotes"], 0)
+        #only 1 joinsplit is required because every input is transparent and there is only a single shielded output
+        assert_equal(len(tx["vjoinsplit"]), 1)
+
+    def z_mergetoaddress_z2t(self, notes_quantity):
+        mytaddr2 = self.nodes[2].listaddresses()[0]
+        listunspent_before_merge = self.nodes[2].listunspent(0)
+        balance_before_merge = self.nodes[2].z_getbalance(mytaddr2)
+        result = self.nodes[0].z_mergetoaddress(["ANY_ZADDR"], mytaddr2, 0, 0, notes_quantity)        
+        wait_and_assert_operationid_status(self.nodes[0], result['opid'], "success", "", 3600)
+        self.nodes[1].generate(100)
+        self.sync_all()
+        tx = self.get_merged_transaction(listunspent_before_merge, self.nodes[2].listunspent(0))
+        assert_equal(result["mergingUTXOs"], 0)
+        assert_equal(result["mergingNotes"], notes_quantity)
+        assert_equal(result["mergingShieldedValue"], self.nodes[2].z_getbalance(mytaddr2) - balance_before_merge)
+        #(notes_quantity - 1) joinsplits are required because first two shielded inputs are merged in a shielded output,
+        #which is then set as shielded input (change) in the subsequent joinsplit and paired with another shielded input
+        #until the final shielded output is obtained; eventually, the last joinsplit contains the total amount of funds
+        #flowinf back to public pool
+        assert_equal(len(tx["vjoinsplit"]), notes_quantity - 1)
+
+    def z_mergetoaddress_z2z(self, notes_quantity):
+        if (len(self.nodes[2].z_listaddresses()) == 0):
+            myzaddr2 = self.nodes[2].z_getnewaddress()
+        else:
+            myzaddr2 = self.nodes[2].z_listaddresses()[0]
+        listunspent_before_merge = self.nodes[2].z_listunspent(0)
+        balance_before_merge = self.nodes[2].z_getbalance(myzaddr2)
+        result = self.nodes[0].z_mergetoaddress(["ANY_ZADDR"], myzaddr2, 0, 0, notes_quantity)        
+        wait_and_assert_operationid_status(self.nodes[0], result['opid'], "success", "", 3600)
+        self.nodes[1].generate(100)
+        self.sync_all()
+        tx = self.get_merged_transaction(listunspent_before_merge, self.nodes[2].z_listunspent(0))
+        assert_equal(result["mergingUTXOs"], 0)
+        assert_equal(result["mergingNotes"], notes_quantity)
+        assert_equal(result["mergingShieldedValue"], self.nodes[2].z_getbalance(myzaddr2) - balance_before_merge)
+        #(notes_quantity - 1) joinsplits are required because first two shielded inputs are merged in a shielded output,
+        #which is then set as shielded input (change) in the subsequent joinsplit and paired with another shielded input
+        #until the final shielded output is obtained
+        assert_equal(len(tx["vjoinsplit"]), notes_quantity - 1)
+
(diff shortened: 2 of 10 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -5946,7 +5946,7 @@ UniValue z_shieldcoinbase(const UniValue& params, bool fHelp)
 #define MERGE_TO_ADDRESS_DEFAULT_TRANSPARENT_LIMIT 50
 #define MERGE_TO_ADDRESS_DEFAULT_SHIELDED_LIMIT 10
 
-#define JOINSPLIT_SIZE JSDescription::getNewInstance(GROTH_TX_VERSION).GetSerializeSize(SER_NETWORK, PROTOCOL_VERSION, GROTH_TX_VERSION)
+#define JOINSPLIT_SIZE(TX_VER) JSDescription::getNewInstance(TX_VER == GROTH_TX_VERSION).GetSerializeSize(SER_NETWORK, PROTOCOL_VERSION, TX_VER)
```
AndreaLanfranchi: Macros like this should be avoided. Let's take the opportunity to remove them and transform in a static function

qa/rpc-tests/wallet_mergetoaddress_2.py
```diff
@@ -0,0 +1,348 @@
...
+        #importing datadir resource
+        import os
+        os.chdir(os.path.dirname(__file__))
+        resource_file = os.getcwd() + "/resources/wallet_mergetoaddress_2/test_setup_.zip"
+        import zipfile
+        with zipfile.ZipFile(resource_file, 'r') as zip_ref:
+            zip_ref.extractall(self.options.tmpdir)
+        copy_tree(self.options.tmpdir + "/test_setup_", self.options.tmpdir)
+        shutil.rmtree(self.options.tmpdir + "/test_setup_")
+
```
a-petrini, line 37: `import os` and `import zipfile` can be moved as global import. Also, be removing the `test_setup_` as root of the zip file (i.e. the root of the folder will contain the directories `node0`, `node1` and `node2`) it is possible to use the extractall() method without the need of do `chdir()`, `copy_tree()` and `rmtree()`.
Finally, using the `join()` function with `os.sep()` as join element will result in a more portable code
```suggestion
    def import_data_to_data_dir(self):
        #importing datadir resource
        resource_file = os.sep.join([os.path.dirname(__file__), 'resources', 'wallet_mergetoaddress_2', 'test_setup_.zip'])
        with zipfile.ZipFile(resource_file, 'r') as zip_ref:
            zip_ref.extractall(self.options.tmpdir)
```

qa/rpc-tests/wallet_mergetoaddress_2.py
```diff
@@ -0,0 +1,348 @@
...
+    def import_data_to_data_dir(self):
+        #importing datadir resource
+        import os
+        os.chdir(os.path.dirname(__file__))
+        resource_file = os.getcwd() + "/resources/wallet_mergetoaddress_2/test_setup_.zip"
+        import zipfile
+        with zipfile.ZipFile(resource_file, 'r') as zip_ref:
+            zip_ref.extractall(self.options.tmpdir)
+        copy_tree(self.options.tmpdir + "/test_setup_", self.options.tmpdir)
+        shutil.rmtree(self.options.tmpdir + "/test_setup_")
```
a-petrini, line 36: `import os` and `import zipfile` can be moved as global import. Also, be removing the `test_setup_` as root of the zip file (i.e. the root of the folder will contain the directories `node0`, `node1` and `node2`) it is possible to use the extractall() method without the need of do `chdir()`, `copy_tree()` and `rmtree()`.
Finally, using the `join()` function with `os.sep()` as join element will result in a more portable code
```suggestion
    def import_data_to_data_dir(self):
        #importing datadir resource
        resource_file = os.sep.join([os.path.dirname(__file__), 'resources', 'wallet_mergetoaddress_2', 'test_setup_.zip'])
        with zipfile.ZipFile(resource_file, 'r') as zip_ref:
            zip_ref.extractall(self.options.tmpdir)
```

a-petrini: `import os` and `import zipfile` can be moved as global import. Also, be removing the `test_setup_` folder as root of the zip file (i.e. the root of the zip will contain the directories `node0`, `node1` and `node2`) it is possible to use the `extractall()` method without the need of do `chdir()`, `copy_tree()` and `rmtree()`.
Finally, using the `join()` function with `os.sep()` as join element will result in a more portable code
```suggestion
    def import_data_to_data_dir(self):
        #importing datadir resource
        resource_file = os.sep.join([os.path.dirname(__file__), 'resources', 'wallet_mergetoaddress_2', 'test_setup_.zip'])
        with zipfile.ZipFile(resource_file, 'r') as zip_ref:
            zip_ref.extractall(self.options.tmpdir)
```

qa/rpc-tests/wallet_mergetoaddress_2.py
```diff
@@ -0,0 +1,348 @@
...
+
+import shutil
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.authproxy import JSONRPCException
+from test_framework.util import assert_equal, initialize_chain_clean, \
+    start_node, connect_nodes_bi, sync_blocks, sync_mempools, \
+    wait_and_assert_operationid_status
+
+import time
+from decimal import Decimal
```
a-petrini: There are a few unused imports that can be removed, in particular `JSONRPCException`, `sync_blocks, sync_mempools`, `time` and `Decimal`

qa/rpc-tests/wallet_mergetoaddress_2.py
```diff
@@ -0,0 +1,348 @@
...
+            results.append(self.nodes[0].z_shieldcoinbase("*", myzaddr0, 0))
+            wait_and_assert_operationid_status(self.nodes[0], results[i]['opid'])
+        self.nodes[1].generate(1)
+        self.sync_all()
+
+    def send_transparent(self, quantity):
+        for i in range(quantity):
+            self.nodes[1].sendmany("", {self.nodes[0].listaddresses()[0]: 10}, 100)
+        self.nodes[1].generate(100)
+        self.sync_all
```
a-petrini: This is a bug, as it is not invoking the sync_all function
```suggestion
        self.sync_all()
```

qa/rpc-tests/wallet_mergetoaddress_2.py
```diff
@@ -0,0 +1,348 @@
...
+
+    def get_merged_transaction(self, unspent_transactions_before_merge, unspent_transactions_after_merge):
+        txid = unspent_transactions_after_merge[0]["txid"]
+        for aft in range(len(unspent_transactions_after_merge)):
+            for bef in range(len(unspent_transactions_before_merge)):
+                if (unspent_transactions_after_merge[aft]["txid"] == unspent_transactions_before_merge[bef]["txid"]):
+                    break
+                if (bef == len(unspent_transactions_before_merge) - 1):
+                    txid = unspent_transactions_after_merge[aft]["txid"]
+        return self.nodes[2].gettransaction(txid)
```
a-petrini: This can be simplified using the `in` keyword to check if a specific element is present in a list or dictionary, given that the input lists differ for at least one element. The suggested function returns `None` if the lists contain the same elements.
```suggestion
    def get_merged_transaction(self, unspent_transactions_before_merge, unspent_transactions_after_merge):
        for tx in unspent_transactions_after_merge:
            if tx not in unspent_transactions_before_merge:
                return self.nodes[2].gettransaction(tx["txid"])
        return None
```

qa/rpc-tests/wallet_mergetoaddress_2.py
```diff
@@ -0,0 +1,348 @@
...
+
+    def z_mergetoaddress_t2t(self, utxos_quantity):
+        mytaddr2 = self.nodes[2].listaddresses()[0]
+        listunspent_before_merge = self.nodes[2].listunspent(0)
+        balance_before_merge = self.nodes[2].z_getbalance(mytaddr2)
+        result = self.nodes[0].z_mergetoaddress(["ANY_TADDR"], mytaddr2, 0, utxos_quantity, 0)        
+        wait_and_assert_operationid_status(self.nodes[0], result['opid'], "success", "", 3600)
+        self.nodes[1].generate(100)
```
a-petrini: I don't know if mining 100 blocks is really a necessity. If it is done to prevent some timing issues, maybe it is possible to main blocks (i.e. mine a block and check the condition) until the issue is solved.

qa/rpc-tests/wallet_mergetoaddress_2.py
a-petrini: As a follow up, the previous suggested change compares the txs in both lists. If you need to compare only the "txid" of each transaction, the function should be this one:
```suggestion
    def get_merged_transaction(self, unspent_transactions_before_merge, unspent_transactions_after_merge):
        unspent_txid_before = [x["txid"] for x in unspent_transactions_before_merge]
        for txid in [x["txid"] for x in unspent_transactions_after_merge]:
            if txid not in unspent_txid_before:
                return self.nodes[2].gettransaction(txid)
        return None
```

qa/rpc-tests/wallet_mergetoaddress_2.py
JackPiri: Done

qa/rpc-tests/wallet_mergetoaddress_2.py
JackPiri: Done

qa/rpc-tests/wallet_mergetoaddress_2.py
JackPiri: Done

src/wallet/rpcwallet.cpp
JackPiri: Done

qa/rpc-tests/wallet_mergetoaddress_2.py
JackPiri: Done

qa/rpc-tests/wallet_mergetoaddress_2.py
JackPiri: Moved to single block mining on node1 (apart from `generate_notes` method).

AndreaLanfranchi (approved):
LGTM

a-petrini:
Do not merge into master

a-petrini (approved):

src/chainparams.cpp
```diff
@@ -394,6 +394,15 @@ bool SelectParamsFromCommandLine()
     return true;
 }
 
+int64_t CChainParams::MaxTipAge() const {
+    if ( Params().NetworkIDString() == "regtest" )
+    {
+        int maxTipAge = (int)GetArg("-maxtipage", nMaxTipAge);
+        return maxTipAge;
```
drgora: Any reason for the cast to `int` and then to `int64_t`? It looks like you can make it more compact as:

```suggestion
        return GetArg("-maxtipage", nMaxTipAge);
```
Also, please try to keep a style consistent with the rest of the file.

src/wallet/rpcwallet.cpp
```diff
@@ -5946,7 +5946,9 @@ UniValue z_shieldcoinbase(const UniValue& params, bool fHelp)
 #define MERGE_TO_ADDRESS_DEFAULT_TRANSPARENT_LIMIT 50
 #define MERGE_TO_ADDRESS_DEFAULT_SHIELDED_LIMIT 10
 
-#define JOINSPLIT_SIZE JSDescription::getNewInstance(GROTH_TX_VERSION).GetSerializeSize(SER_NETWORK, PROTOCOL_VERSION, GROTH_TX_VERSION)
+int GetJoinSplitSize(int shieldedTxVersion) {
+    return JSDescription::getNewInstance(shieldedTxVersion == GROTH_TX_VERSION).GetSerializeSize(SER_NETWORK, PROTOCOL_VERSION, shieldedTxVersion);
+}
```
drgora: Just a small thing here; I'd suggest to put this function in an anonymous namespace (or alternatively make it `static`), so that its visibility remains limited to this file, given that it's not used elsewhere.

```suggestion
namespace {

int GetJoinSplitSize(int shieldedTxVersion) {
    return JSDescription::getNewInstance(shieldedTxVersion == GROTH_TX_VERSION).GetSerializeSize(SER_NETWORK, PROTOCOL_VERSION, shieldedTxVersion);
}

}
```

qa/pull-tester/rpc-tests.sh
```diff
@@ -16,6 +16,10 @@ for i in "$@"; do
       EXTENDED="true"
       shift
       ;;
+    -specific=*)
```
ptagl: I thinks it's great now to have the flexibility to run one specific test now, I just wonder if we can rename the `-specific` keyword to something else to improve clarity.

I don't have any particular keyword in mind, so I would like to have feedback from other people.

ptagl:
This PR introduces a zip file `qa/rpc-tests/resources/wallet_mergetoaddress_2/test_setup_.zip` that is used by the Python test `wallet_mergetoaddress_2.py` as initial state of the datadir to save time spent running the test on the CI.

This zipped data folder baiscally contains a regtest chain state with a lot of shielded notes to be used as input of the `z_mergetoaddress` RPC command.

The new file takes around 15 MB on the repo, I wonder if it is worth to place it somewhere else, like a cloud bucket.

@cronicc what do you think about it?

qa/zcash/full_test_suite.py
```diff
@@ -192,6 +192,8 @@ def main():
                         action='store', help='comma separated string of rpc tests to exclude, see qa/rpc-tests/README.md for more')
     parser.add_argument('--rpc-split', dest='split',
                         action='store', help='string in format m:n, see qa/rpc-tests/README.md for more')
+    parser.add_argument('--rpc-specific', dest='specific',
+                        action='store', help='execute only a specific python test, see qa/rpc-tests/README.md for more')
```
ptagl, line 196: Shall we add information to the README file about the usage of this argument (eventually after its renaming)?

src/init.cpp
```diff
@@ -609,6 +609,9 @@ std::string HelpMessage(HelpMessageMode mode)
         strUsage += HelpMessageOpt("-metricsrefreshtime", strprintf(_("Number of seconds between metrics refreshes (default: %u if running in a console, %u otherwise)"), 1, 600));
     }
 
+    strUsage += HelpMessageOpt("-maxtipage=<n>",
+        "regtest only - Set the maximum tip age)");
```
ptagl:
```suggestion
        "regtest only - Set the maximum tip age used for the relay of chain blocks)");
```

qa/rpc-tests/wallet_mergetoaddress_2.py
```diff
@@ -0,0 +1,341 @@
...
+        self.send_transparent(4)
+        #merge up to 4 utxos (available more than 4) from n0 to n2
+        result = self.z_mergetoaddress_t2t(4)
+        assert_equal(result[self.op_result_k][self.mergingUTXOs_k], 4)
+        assert_equal(result[self.op_result_k][self.mergingNotes_k], 0)
+        assert_equal(result[self.op_result_k][self.mergingTransparentValue_k], result[self.balance_delta_k])
+        assert_equal(result[self.op_result_k][self.mergingShieldedValue_k], 0)
+        #no joinsplit is required because every input is transparent and every output is transparent
+        assert_equal(result[self.joinsplit_len_k], 0)
+        # ---------- T to T ----------
```
ptagl: Is there any specific reason to test `z_mergetoaddress_t2t(3)` and then `z_mergetoaddress_t2t(4)` immediately after? Do they represent special cases?

qa/rpc-tests/wallet_mergetoaddress_2.py
```diff
@@ -0,0 +1,341 @@
...
+    op_result_k = "op_result"
+    balance_delta_k = "balance_delta"
+    joinsplit_len_k = "joinsplit_len"
+    mergingUTXOs_k = "mergingUTXOs"
+    mergingNotes_k = "mergingNotes"
+    mergingTransparentValue_k = "mergingTransparentValue"
+    mergingShieldedValue_k = "mergingShieldedValue"
+
+    def import_data_to_data_dir(self):
+        #importing datadir resource
```
ptagl: Can we add a white space at the beginning of any comment to improve readability?

qa/rpc-tests/wallet_mergetoaddress_2.py
```diff
@@ -0,0 +1,341 @@
...
+        args2 = ['-debug=zrpcunsafe', '-experimentalfeatures', '-zmergetoaddress', '-mempooltxinputlimit=7', "-maxtipage=36000000"]
+        self.nodes.append(start_node(2, self.options.tmpdir, args2))
+        connect_nodes_bi(self.nodes,0,1)
+        connect_nodes_bi(self.nodes,1,2)
+        connect_nodes_bi(self.nodes,0,2)
+        self.is_network_split=False
+        self.sync_all()
+
+    def run_test (self):
+
```
ptagl, line 89: What about extending the current description to explain why this Python test has been developed? If I remember correctly, it was to test specifically a change to the `z_mergetoaddress()` RPC command that fixed the estimation of the final transaction size. Is it right?

qa/rpc-tests/wallet_mergetoaddress_2.py
```diff
@@ -0,0 +1,341 @@
...
+        #merge up to 60 notes (available more than 60, but transaction size limit admits up to 58) from n0 to n2
+        result = self.z_mergetoaddress_z2z(60)
+        assert_equal(result[self.op_result_k][self.mergingUTXOs_k], 0)
+        assert_equal(result[self.op_result_k][self.mergingNotes_k], 58)
+        assert_equal(result[self.op_result_k][self.mergingTransparentValue_k], 0)
+        assert_equal(result[self.op_result_k][self.mergingShieldedValue_k], result[self.balance_delta_k])
+        #57 joinsplits are required because first two shielded inputs are merged in a shielded output,
+        #which is then set as shielded input (change) in the subsequent joinsplit and paired with another shielded input
+        #until the final shielded output is obtained
+        assert_equal(result[self.joinsplit_len_k], 57)
```
ptagl: I guess that this test is important to check the behavior of the implemented fix. I would expect that this test was previously failing due to a wrong estimation of the transaction size, while after the fix works as expected. Am I right?

If so, I'm not sure if all the previous tests and the last one are relevant to this PR.

ptagl (changes requested):

qa/rpc-tests/wallet_mergetoaddress_2.py
```diff
@@ -0,0 +1,341 @@
...
+
+    op_result_k = "op_result"
+    balance_delta_k = "balance_delta"
+    joinsplit_len_k = "joinsplit_len"
+    mergingUTXOs_k = "mergingUTXOs"
+    mergingNotes_k = "mergingNotes"
+    mergingTransparentValue_k = "mergingTransparentValue"
+    mergingShieldedValue_k = "mergingShieldedValue"
+
+    def import_data_to_data_dir(self):
```
drgora, line 23: I think it would be nice to add, either here or to the `run_test`, a description of the state being imported, e.g. in terms of chains height, num of utxos for each node and so...

Also, using a more specific name, compared to `test_setup_.zip`, might be more future proof in case we want to implement a similar approach in other tests.

qa/rpc-tests/wallet_mergetoaddress_2.py
```diff
@@ -0,0 +1,341 @@
...
+        self.is_network_split=False
+        self.sync_all()
+
+    def run_test (self):
+
+        # Node0 acts as "from" (merge_to_address sender)
+        # Node1 acts as miner and funds distributor
+        # Node2 acts ad "to" (merge_to_address receiver)
+
+        #(generate 1b on n0, generate 100bs on node1, shield 1 b on n0) * 100 + generate 1b on n1
```
drgora: Does this line refer to the state imported with `import_data_to_data_dir`? If so, could you please make it explicit?

src/init.cpp
JackPiri: Done

qa/pull-tester/rpc-tests.sh
JackPiri: Maybe something like `-runsingletest` can be more clear?

qa/rpc-tests/wallet_mergetoaddress_2.py
JackPiri: Done

src/chainparams.cpp
JackPiri: Done

qa/rpc-tests/wallet_mergetoaddress_2.py
JackPiri: Yes, it is related to the state imported with `import_data_to_data_dir`, added explicit comment on that function.

qa/pull-tester/rpc-tests.sh
ptagl: As a temporary proposal we could use `-runonly`. We could also consider to refactor the script to have a regex parameter to filter the list of tests.

JackPiri: Done

qa/rpc-tests/wallet_mergetoaddress_2.py
JackPiri: Done

src/wallet/rpcwallet.cpp
JackPiri: Done

qa/rpc-tests/wallet_mergetoaddress_2.py
JackPiri: Yes, added desciption of the test at the start of `run_test` (both for `wallet_mergetoaddress.py` as for `wallet_mergetoaddress`)

qa/rpc-tests/wallet_mergetoaddress_2.py
JackPiri: Removed other tests, moved tz2t test to `wallet_mergetoaddress.py`

qa/rpc-tests/wallet_mergetoaddress_2.py
JackPiri: Removed all tests apart from oversized transaction test

ptagl (approved):

PR #480, opened 2022-10-11
Gp/support indexing for p2pk

### LINKED ISSUE #352: Coinbase transactions are only indexed if their outputs have been spent as inputs of a new tx
If an address has only ever received coinbase transactions, it is not indexed by the block explorer.

Early testnet blocks are a good example of this:
https://explorer-testnet.zensystem.io/block/00143a02152cf0cf3b6cbb3d211f51d748c5dbd433c2033bbcafd60402582874 - 12.5 zen going to address ztWuqiGBEHkKSMBgbrbzTZbHudPzY2jE4x8
https://explorer-testnet.zensystem.io/address/ztWuqiGBEHkKSMBgbrbzTZbHudPzY2jE4x8 - address not showing a balance

```
zen-cli --testnet getaddresstxids '{"addresses": ["ztWuqiGBEHkKSMBgbrbzTZbHudPzY2jE4x8"]}'
[
]
zen-cli --testnet getaddressbalance '{"addresses": ["ztWuqiGBEHkKSMBgbrbzTZbHudPzY2jE4x8"]}'
{
  "balance": 0,
  "received": 0
}
zen-cli --testnet getaddressdeltas '{"addresses": ["ztWuqiGBEHkKSMBgbrbzTZbHudPzY2jE4x8"]}'
[
]
zen-cli --testnet getaddressutxos '{"addresses": ["ztWuqiGBEHkKSMBgbrbzTZbHudPzY2jE4x8"]}'
[
]
```

If the coinbase tx vout become vin of a new transaction, they get indexed.

The issue likely is in this part of the code https://github.com/HorizenOfficial/zen/blob/3aaeab91ea4608fbb3f92915d0ac588aafc0f2f2/src/main.cpp#L2565-L2578

Only if a tx is `!tx.IsCoinBase()` it gets added to AddressIndex and SpentIndex, we would need an else specific for AddressIndexing here where coinbase transactions are handled.

JackPiri:
Pull request associated to issue https://github.com/HorizenOfficial/zen/issues/352


src/addressindex.h
```diff
@@ -21,20 +27,20 @@ struct CAddressUnspentKey {
     }
     template<typename Stream>
     void Serialize(Stream& s, int nType, int nVersion) const {
-        ser_writedata8(s, type);
+        ser_writedata8(s, (uint8_t)type); //uint8_t is used for backward compatibility
```
AndreaLanfranchi: I strongly suggest to begin avoiding C-Style casts.
I understand they're shorter to write but actually they subtle inject (at runtime) a mix of different casts (const, reinterpret) and therefore can lead to unexpected results.
When we know the type we're casting from and the type we're casting to use `static_cast`

```suggestion
        ser_writedata8(s, static_cast<uint8_t>(type)); //uint8_t is used for backward compatibility
```

src/addressindex.h
```diff
@@ -367,4 +373,17 @@ struct CMempoolAddressDeltaKeyCompare
...
+{
+    AddressType addressType = AddressType::UNKNOWN;
+    if (scriptType == CScript::ScriptType::P2PKH || scriptType == CScript::ScriptType::P2PK)
+        addressType = AddressType::PUBKEY;
+    else if (scriptType == CScript::ScriptType::P2SH)
+        addressType = AddressType::SCRIPT;
+    return addressType;
+}
+
+#endif // BITCOIN_ADDRESSINDEX_H
```
AndreaLanfranchi: Missing new line at end of file

src/base58.cpp
```diff
@@ -155,17 +156,19 @@ CTxDestination CBitcoinAddress::Get() const
...
-        type = 1;
+        type = AddressType::PUBKEY;
         return true;
-    } else if (vchVersion == Params().Base58Prefix(CChainParams::SCRIPT_ADDRESS)) {
+    } else if (vchVersion == Params().Base58Prefix(CChainParams::SCRIPT_ADDRESS)
+               || vchVersion == Params().Base58Prefix(CChainParams::SCRIPT_ADDRESS_OLD)) {
         memcpy(&hashBytes, &vchData[0], 20);
-        type = 2;
+        type = AddressType::SCRIPT;
         return true;
```
AndreaLanfranchi: This is a bit ugly ... as each `if` branch implies a return I'd suggest simply 
```
    if (!IsValid()) {
        return false;
    }

    if (vchVersion == Params().Base58Prefix(CChainParams::PUBKEY_ADDRESS) || vchVersion == Params().Base58Prefix(CChainParams::PUBKEY_ADDRESS_OLD)) {
        memcpy(&hashBytes, &vchData[0], 20);
        type = AddressType::PUBKEY;
        return true;
    }
    
    if (vchVersion == Params().Base58Prefix(CChainParams::SCRIPT_ADDRESS) || vchVersion == Params().Base58Prefix(CChainParams::SCRIPT_ADDRESS_OLD)) {
        memcpy(&hashBytes, &vchData[0], 20);
        type = AddressType::SCRIPT;
        return true;
    }

    return false;
```
Is more readable, less (unneeded) `else`s and less indentation of code

src/rpc/misc.cpp
```diff
@@ -484,24 +485,24 @@ UniValue setmocktime(const UniValue& params, bool fHelp)
...
-    if (type == 2) {
+    if (type == AddressType::SCRIPT) {
         address = CBitcoinAddress(CScriptID(hash)).ToString();
-    } else if (type == 1) {
+    } else if (type == AddressType::PUBKEY) {
         address = CBitcoinAddress(CKeyID(hash)).ToString();
     } else {
         return false;
     }
     return true;
```
AndreaLanfranchi: When dissecting cases over an enum prefer the use of `switch` 

```suggestion
    switch (type) {
    case AddressType::SCRIPT:
        address = CBitcoinAddress(CScriptID(hash)).ToString();
        return true;
    case AddressType::PUBKEY:
        address = CBitcoinAddress(CKeyID(hash)).ToString();
        return true;
    default:
        return false;
    }
```

src/rpc/misc.cpp
```diff
@@ -666,15 +667,15 @@ UniValue getaddressutxos(const UniValue& params, bool fHelp)
...
         throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid address");
     }
 
     std::vector<std::pair<CAddressUnspentKey, CAddressUnspentValue> > unspentOutputs;
 
-    for (std::vector<std::pair<uint160, int> >::iterator it = addresses.begin(); it != addresses.end(); it++) {
+    for (std::vector<std::pair<uint160, AddressType> >::iterator it = addresses.begin(); it != addresses.end(); it++) {
         if (!GetAddressUnspent((*it).first, (*it).second, unspentOutputs)) {
             throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "No information available for address");
         }
```
AndreaLanfranchi: Prefer use of range-loop to improve readability
```suggestion
    for (const auto& [addressHash, addressType] : addresses) {
        if (!GetAddressUnspent(addressHash, addressType, unspentOutputs)) {
            throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "No information available for address");
        }
    }
```

src/rpc/misc.cpp
```diff
@@ -818,15 +819,15 @@ UniValue getaddressdeltas(const UniValue& params, bool fHelp)
...
+    std::vector<std::pair<uint160, AddressType> > addresses;
 
     if (!getAddressesFromParams(params, addresses)) {
         throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid address");
     }
 
     std::vector<std::pair<CAddressIndexKey, CAddressIndexValue> > addressIndex;
 
-    for (std::vector<std::pair<uint160, int> >::iterator it = addresses.begin(); it != addresses.end(); it++) {
+    for (std::vector<std::pair<uint160, AddressType> >::iterator it = addresses.begin(); it != addresses.end(); it++) {
```
AndreaLanfranchi: Same as above comment

src/script/script.cpp
```diff
@@ -262,6 +263,21 @@ bool CScript::IsPayToScriptHash() const
...
 }
 
+bool CScript::IsPayToPublicKey() const
+{
+    // Extra-fast test for pay-to-pubkey CScripts:
+
+    // Check if this script is P2PK without OP_CHECKBLOCKATHEIGHT
+    bool p2pk = (this->size() == 35 &&
+               (*this)[0] == 0x21 &&
+               (*this)[34] == OP_CHECKSIG);
```
AndreaLanfranchi: Why dereference `size` ? CScript inherits from vector and `size` is public method
```suggestion
    bool p2pk = (size() == 35 &&
               (*this)[0] == 0x21 &&
               (*this)[34] == OP_CHECKSIG);
```

src/script/script.cpp
```diff
@@ -262,6 +263,21 @@ bool CScript::IsPayToScriptHash() const
...
+
+     // Check if this script is P2PKH with OP_CHECKBLOCKATHEIGHT
+    bool p2pkWithReplay = false; //TODO: check format
```
AndreaLanfranchi: Can we solve this TODO ?

src/script/script.cpp
```diff
@@ -306,27 +322,49 @@ std::string CScript::ToString() const
 CScript::ScriptType CScript::GetType() const
 {
     if (this->IsPayToPublicKeyHash())
-        return CScript::P2PKH;
+        return CScript::ScriptType::P2PKH;
     if (this->IsPayToScriptHash())
-        return CScript::P2SH;
+        return CScript::ScriptType::P2SH;
+    if (this->IsPayToPublicKey())
+        return CScript::ScriptType::P2PK;
```
AndreaLanfranchi: Please remove all those `this->`

src/addressindex.h
JackPiri: Done

src/addressindex.h
JackPiri: Done

src/base58.cpp
JackPiri: Done

src/rpc/misc.cpp
JackPiri: Generally I'd agree with your opinion but I am not so sure in case of a very short set valued enum (like in this case, just 3 possible values); can you please share your view on this?

src/rpc/misc.cpp
JackPiri: Done (totally agree!)

src/rpc/misc.cpp
JackPiri: Done

src/script/script.cpp
JackPiri: Done

src/script/script.cpp
JackPiri: Done

src/rpc/misc.cpp
AndreaLanfranchi: There are many reasons to prefer that:

- an `if ... else` uses multiple statements for  multiple choices while `switch` uses a single expression
- `if` can evaluate multiple conditions for **any** logical expression while `case` can compare **only** for equality of chars or integer values
- as the main goal (in this case) is traverse an enum, as the switch can't have two identical `case` values we're implicitly safe against possible errors like ` if (type == AddressType::SCRIPT) {...} else if (type == AddressType::SCRIPT) { ...}` which builds correctly but obviously is not what we want to see.
- efficiency : compiler optimizes (at any O level) `switch` over `if ... else` (see https://godbolt.org/z/8r4shrKPG)

JackPiri: Done

src/script/script.cpp
JackPiri: I have added following code, trying to replicate from p2pkh->p2pkhWithReplay and p2sh->p2shWithReplay:


`

     // Check if this script is P2PKH with OP_CHECKBLOCKATHEIGHT

     // The overall size should not be more then 74:

     // 35 bytes for common P2PK + 1 byte size + 32 bytes of block hash + 1 byte size + 4 bytes of block height + 1 byte opcode

    bool p2pkWithReplay = (size() > 35 &&
                           size() < 75 &&
                           at(0) == 0x21 &&
                           at(34) == OP_CHECKSIG &&
                           at(size() - 1) == OP_CHECKBLOCKATHEIGHT);
`

unfortunately I was not able to find any useful documentation supporting this script scheme, @ptagl can you please confirm on correctness?

a-petrini (approved):

src/script/script.cpp
AndreaLanfranchi: ping @ptagl

src/txmempool.cpp
```diff
@@ -5,6 +5,7 @@
 
 #include "txmempool.h"
 
+#include "addressindex.h"
```
ptagl: I wonder if this include should go under a conditionally compiled section since it is used by the Address Indexing only.
```suggestion
#ifdef ENABLE_ADDRESS_INDEXING
#include "addressindex.h"
#endif // ENABLE_ADDRESS_INDEXING
```

src/txdb.h
```diff
@@ -6,6 +6,7 @@
 #ifndef BITCOIN_TXDB_H
 #define BITCOIN_TXDB_H
 
+#include "addressindex.h"
```
ptagl: I wonder if this include should go under a conditionally compiled section since it is used by the Address Indexing only.

```suggestion
#ifdef ENABLE_ADDRESS_INDEXING
#include "addressindex.h"
#endif // ENABLE_ADDRESS_INDEXING
```

src/txdb.cpp
```diff
@@ -5,6 +5,7 @@
 
 #include "txdb.h"
 
+#include "addressindex.h"
```
ptagl: I wonder if this include should go under a conditionally compiled section since it is used by the Address Indexing only.

```suggestion
#ifdef ENABLE_ADDRESS_INDEXING
#include "addressindex.h"
#endif // ENABLE_ADDRESS_INDEXING
```

src/main.cpp
```diff
@@ -7,6 +7,7 @@
 
 #include "sodium.h"
 
+#include "addressindex.h"
```
ptagl: Can this include be removed as already present in `main.h`?

```suggestion
```

src/main.h
```diff
@@ -15,6 +15,7 @@
 #include "timestampindex.h"
 #endif // ENABLE_ADDRESS_INDEXING
 
+#include "addressindex.h"
```
ptagl: I wonder if this include should go under a conditionally compiled section since it is used by the Address Indexing only.

```suggestion
#ifdef ENABLE_ADDRESS_INDEXING
#include "addressindex.h"
#endif // ENABLE_ADDRESS_INDEXING
```

src/rpc/blockchain.cpp
```diff
@@ -3,6 +3,7 @@
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
+#include "addressindex.h"
```
ptagl, line 7: I wonder if this include should go under a conditionally compiled section since it is used by the Address Indexing only.

```suggestion
#ifdef ENABLE_ADDRESS_INDEXING
#include "addressindex.h"
#endif // ENABLE_ADDRESS_INDEXING
```

src/rpc/misc.cpp
```diff
@@ -3,6 +3,7 @@
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
+#include "addressindex.h"
```
ptagl, line 7: I wonder if this include should go under a conditionally compiled section since it is used by the Address Indexing only.

```suggestion
#ifdef ENABLE_ADDRESS_INDEXING
#include "addressindex.h"
#endif // ENABLE_ADDRESS_INDEXING
```

src/script/script.cpp
```diff
@@ -305,28 +327,50 @@ std::string CScript::ToString() const
...
         // unknown script type; return zeros (this can happen)
         vector<unsigned char> hashBytes;
         hashBytes.resize(20, 0);
         return uint160(hashBytes);
     }
-    vector<unsigned char> hashBytes(this->begin() + start, this->begin() + start + 20);
+    
+    vector<unsigned char> hashBytes(this->begin() + start, this->begin() + start + size);
+
+    if (toBeHashed)
```
ptagl: Just for readability, what about renaming this boolean variable as `isPayToPublicKey`?

src/spentindex.h
```diff
@@ -53,11 +54,19 @@ struct CSpentIndexValue {
...
-        READWRITE(addressType);
+
+        //int is used for backward compatibility
+        int addressTypeInt = 0;
+        if (!ser_action.ForRead()) //serialize
+            addressTypeInt = (int)addressType;
+        READWRITE(addressTypeInt);
+        if (ser_action.ForRead()) //unserialize
+            addressType = (AddressType)addressTypeInt;
+
```
ptagl: Do you think this code would work properly also if we remove the distinction between serialization and deserialization?

```suggestion

        // int is used for backward compatibility
        int addressTypeInt = static_cast<int>(addressType);
        READWRITE(addressTypeInt);
        addressType = static_cast<AddressType>(addressTypeInt);

```

ptagl (changes requested):
About the inclusion of `addressindex.h`, we could alternatively try to use Forward Declarations.

src/script/script.cpp
JackPiri: Done

src/spentindex.h
JackPiri: Done

src/txmempool.cpp
JackPiri: Removed because already included by `txmempool.h`

src/txdb.h
JackPiri: Done (added forward declaration for `AddressType`)

src/txdb.cpp
JackPiri: Removed because already included by `txdb.h`

src/main.cpp
JackPiri: Done

src/main.h
JackPiri: Done (`"addressindex.h"` include replaced by `AddressType` forward declaration)

src/rpc/blockchain.cpp
JackPiri: Done

src/rpc/misc.cpp
JackPiri: Done

src/script/script.cpp
JackPiri: Done

src/spentindex.h
JackPiri: Done

src/script/script.cpp
ptagl: It seems fine to me as it reflects the same logic in the `Solver` function inside `standard.cpp`.

JackPiri: Agree

src/main.h
```diff
@@ -13,6 +13,7 @@
 #ifdef ENABLE_ADDRESS_INDEXING
 #include "spentindex.h"
 #include "timestampindex.h"
+enum class AddressType;
```
ptagl: Personally I would prefer to see this forward declaration in its proper section (e.g. immediately before the declaration of `class CTransaction;` or after `class CTxInUndo;`), even though we need to add an `#ifdef` just for it, but feel free to ignore this comment.

ptagl (approved):
I see a lot of commits like "Refactoring" or "Fixes", could you please try to squash them to make the history more readable before merging?

src/main.h
JackPiri: Done

src/script/script.cpp
```diff
@@ -217,24 +218,24 @@ bool CScript::IsPayToPublicKeyHash() const
...
+    bool p2pkh = (size() == 25 &&
+                  at(0) == OP_DUP &&
+                  at(1) == OP_HASH160 &&
+                  at(2) == 0x14 &&
+                  at(23) == OP_EQUALVERIFY &&
+                  at(24) == OP_CHECKSIG);
 
      // Check if this script is P2PKH with OP_CHECKBLOCKATHEIGHT
-     // The overall size should not be more then 62:
+     // The overall size should not be more then 64:
```
drgora: Just a small typo, but since we're changing this already... `then` -> `than`

src/rpc/misc.cpp
```diff
@@ -484,24 +487,26 @@ UniValue setmocktime(const UniValue& params, bool fHelp)
     return NullUniValue;
 }
 
-bool getAddressFromIndex(const int &type, const uint160 &hash, std::string &address)
+bool getAddressFromIndex(const AddressType &type, const uint160 &hash, std::string &address)
```
drgora: I believe it does not quite make sense to pass `type` (which is an `int`) as a reference in this case. I'd suggest
```suggestion
bool getAddressFromIndex(const AddressType type, const uint160 &hash, std::string &address)
```

src/script/script.cpp
```diff
@@ -244,24 +245,45 @@ bool CScript::IsPayToScriptHash() const
...
+{
+    // Extra-fast test for pay-to-pubkey CScripts:
+
+    // Check if this script is P2PK without OP_CHECKBLOCKATHEIGHT
+    bool p2pk = (size() == 35 &&
+                 at(0) == 0x21 &&
+                 at(34) == OP_CHECKSIG);
+
+     // Check if this script is P2PKH with OP_CHECKBLOCKATHEIGHT
+     // The overall size should not be more then 74:
```
drgora: This should be P2PK, right?
```suggestion
     // Check if this script is P2PK with OP_CHECKBLOCKATHEIGHT
     // The overall size should not be more than 74:
```

src/script/script.cpp
```diff
@@ -305,28 +327,50 @@ std::string CScript::ToString() const
...
+    }
     else {
         // unknown script type; return zeros (this can happen)
         vector<unsigned char> hashBytes;
         hashBytes.resize(20, 0);
         return uint160(hashBytes);
     }
-    vector<unsigned char> hashBytes(this->begin() + start, this->begin() + start + 20);
+    
+    vector<unsigned char> hashBytes(this->begin() + start, this->begin() + start + size);
```
drgora, line 364: I understand it's not really in the scope of this PR, but I see an inconsistency in this file when using the `std` namespace. Given that it should be real quick, I'd take this chance to fix it.

My personal preference, and also the most common style in our codebase, goes for being explicit with `std::` each time it's necessary, and removing the `using namespace std` at the beginning of this file.

drgora (approved):
This PR is ok for me. I have left few minor comments, none of which is a real blocker. Feel free to apply any subset of those :)

src/script/script.cpp
```diff
@@ -305,28 +327,50 @@ std::string CScript::ToString() const
...
+    {
+        start = 1;
+        size = 33;
+        isPayToPublicKey = true;
+    }
     else {
         // unknown script type; return zeros (this can happen)
         vector<unsigned char> hashBytes;
         hashBytes.resize(20, 0);
         return uint160(hashBytes);
```
drgora: How about
```suggestion
        return uint160(std::vector<unsigned char>(size));
```
where `size` is initialized to `20` when it's declared?

src/script/script.cpp
JackPiri: Done

src/rpc/misc.cpp
JackPiri: Done

src/script/script.cpp
JackPiri: Correct, done

src/script/script.cpp
JackPiri: Make sense, more compact. Done

src/script/script.cpp
JackPiri: Done

PR #481, opened 2022-10-17
Key rotation

drgora:
The main goal of this PR is introducing support for key rotation. Key rotation is enabled by default for `v2` sidechains' certificates, while certificates for `pre-v2` sidechains remain unchanged.

To such extent, more changes are introduced along the way:
- `mc-cryptolib` (and in turn `cctp-lib`) are updated to their most recent development versions
- `mcTestCall` is refactored (see https://github.com/HorizenOfficial/zendoo-mc-cryptolib/pull/63), and compiled as a library in `depends`
- python test utilities are updated to align with the new interface to `mcTestCall`
- few tests are updated to be more robust wrt variable performance for proofs creation and verification


src/sc/proofverifier.cpp
```diff
@@ -33,7 +33,13 @@ std::string ProofVerificationResultToString(ProofVerificationResult res)
...
  * @return CCertProofVerifierInput The structure containing all the data needed for the proof verification of the certificate.
  */
-CCertProofVerifierInput CScProofVerifier::CertificateToVerifierItem(const CScCertificate& certificate, const Sidechain::ScFixedParameters& scFixedParams, CNode* pfrom)
+#ifdef BITCOIN_TX
+CCertProofVerifierInput CScProofVerifier::CertificateToVerifierItem(const CScCertificate& certificate, const Sidechain::ScFixedParameters& scFixedParams, CNode* pfrom, const CCoinsViewCache* view)
+{
+    return CCertProofVerifierInput();
+}
+#else
+CCertProofVerifierInput CScProofVerifier::CertificateToVerifierItem(const CScCertificate& certificate, const Sidechain::ScFixedParameters& scFixedParams, CNode* pfrom, const CCoinsViewCache* view)
```
AndreaLanfranchi, line 42: Please insert pre-processor directives inside function body when function signature is identical

src/sc/proofverifier.cpp
```diff
@@ -81,8 +87,15 @@ CCertProofVerifierInput CScProofVerifier::CertificateToVerifierItem(const CScCer
     certData.proof = certificate.scProof;
     certData.verificationKey = scFixedParams.wCertVk;
 
+    if (view != nullptr) {
+        if (scFixedParams.version >= 2) {
+            auto const& cdh = view->GetActiveCertView(certData.scId).certDataHash;
+            certData.lastCertHash = cdh.IsNull() ? CFieldElement::GetPhantomHash() : cdh;
+        }
+    }
```
AndreaLanfranchi: Save one indentation level if possible

```suggestion
    if (view && scFixedParams.version >= 2) {
        auto const& cdh = view->GetActiveCertView(certData.scId).certDataHash;
        certData.lastCertHash = cdh.IsNull() ? CFieldElement::GetPhantomHash() : cdh;
    }
```

src/sc/sidechaintypes.cpp
```diff
@@ -273,10 +274,13 @@ CFieldElement CFieldElement::ComputeHash(const CFieldElement& lhs, const CFieldE
 
 const CFieldElement& CFieldElement::GetPhantomHash()
 {
-    // TODO call an utility method to retrieve from zendoo_mc_cryptolib a constant phantom hash
-    // field element and use it everywhere it is needed a constant value whose preimage has to
-    // be unknown
-    static CFieldElement ret{std::vector<unsigned char>(CFieldElement::ByteSize(), 0x00)};
+    // zendoo_mc_cryptolib allocates a new buffer with static lifetime on each call to zendoo_get_phantom_cert_data_hash().
+    static field_t* phantom_ptr = zendoo_get_phantom_cert_data_hash();
+    size_t size = (int)CFieldElement::ByteSize() / sizeof(unsigned char);
```
AndreaLanfranchi: Why convert sign when operands and result are meant to be unsigned ?

AndreaLanfranchi: Also not very clear to me how _staticness_ is enforced. Apparently at every call a buffer is allocated

src/txmempool.cpp
```diff
@@ -2024,23 +2024,27 @@ bool CCoinsViewMemPool::GetSidechain(const uint256& scId, CSidechain& info) cons
...
-            const CScCertificate& certTopQual = mempool.mapCertificate[topQualHash].GetCertificate();
-            const auto map_it = mapCumtreeHeight.find(certTopQual.endEpochCumScTxCommTreeRoot.GetLegacyHash());
-            
-            if (map_it == mapCumtreeHeight.end())
-            {
-                LogPrint("mempool", "%s():%d - could not find referenced block for certTopQual %s\n", __func__, __LINE__, certTopQual.GetHash().ToString());
-                info.lastUnconfirmedReferencedHeight = -1; // -1 stands for no valid values in mempool
-            }
-            else
+        if (!mempool.mapSidechains.at(scId).mBackwardCertificates.empty()) {
```
AndreaLanfranchi: Just a few lines above we do 
```
(mempool.mapSidechains.at(scId)
```
Maybe is worth to save a reference and avoid a second lookup on the map

src/txmempool.cpp
```diff
@@ -2024,23 +2024,27 @@ bool CCoinsViewMemPool::GetSidechain(const uint256& scId, CSidechain& info) cons
...
+
+            if (info.isNonCeasing())
```
AndreaLanfranchi: I'd put this check above this.
```
if (!mempool.mapSidechains.at(scId).mBackwardCertificates.empty())
```
 or join conditions within a single `if`. We actualy don't care about emptyness if sc is not nonceasing

AndreaLanfranchi (commented):
Also ... do you guys have clang format plugins installed in your IDE/editor ?

AndreaLanfranchi:
Another thing. Why merge into `ap/nonCeasingSC` ?

src/txmempool.cpp
drgora: Merged in 7e39749d9b01b1ee9557e5bc221d95963163f968

src/txmempool.cpp
drgora: Yeah you're right, nice catch! I've reworked this section to remove all the unneeded accesses, thanks (7e39749d9b01b1ee9557e5bc221d95963163f968)

src/sc/sidechaintypes.cpp
drgora: Woops, this is a leftover from a previous implementation. I updated the function entirely to avoid memcopies on each call, please see 7e39749d9b01b1ee9557e5bc221d95963163f968

src/sc/proofverifier.cpp
drgora: Ok, done in 7e39749d9b01b1ee9557e5bc221d95963163f968

src/sc/proofverifier.cpp
drgora: The pattern I used is quite common all over our codebase. I'd rather keep it consistent for now, and refactor all these cases in another PR if we want to.

AndreaLanfranchi: Fair enough. Thank you.

qa/rpc-tests/test_framework/mc_test/mc_test.py
```diff
@@ -27,20 +27,42 @@ def __init__(self, datadir, srcdir, ps_type = "cob_marlin"):
...
+        args.append(op)
+
+        args += ["-c", str(circuit_type), "-p", str(ps_type), "-s", str(segment_size), "-n", str(num_constraints)]
+
+        if constant is not None:
+            args.append("-k")
+            if constant == True:
+                args.append("CONSTANT_PLACEHOLDER")
+            else:
+                args.append(str(constant))
```
a-petrini, line 43: What does happen when `constant == False`?

qa/rpc-tests/test_framework/mc_test/mc_test.py
```diff
@@ -27,20 +27,42 @@ def __init__(self, datadir, srcdir, ps_type = "cob_marlin"):
         assert(ps_type == "darlin" or ps_type == "cob_marlin")
         self.ps_type = ps_type
 
-    def _generate_params(self, id, circuit_type, ps_type, file_prefix, num_constraints, segment_size):
-        params_dir = self._get_params_dir(id)
+    def _get_file_prefix(self, circ_type, constant):
+        if constant != None:
+            return str(self.ps_type) + "_" + circ_type + "_"
+        return str(self.ps_type) + "_" + circ_type + "_no_const_"
```
a-petrini: This could be shortened to:
```suggestion
    def _get_file_prefix(self, circ_type, constant):
        return str(self.ps_type) + "_" + circ_type + ("_" if constant != None else "_no_const_")
```

qa/rpc-tests/sc_big_block.py
```diff
@@ -140,8 +141,8 @@ def advance_sidechains_epoch(num_of_scs):
         certMcTest = CertTestUtils(self.options.tmpdir, self.options.srcdir, CERT_PROVING_SYSTEM)
         cswMcTest = CSWTestUtils(self.options.tmpdir, self.options.srcdir, CSW_PROVING_SYSTEM)
 
-        certVk = certMcTest.generate_params('scs', 'cert', CERT_NUM_CONSTRAINTS, SEGMENT_SIZE)
-        cswVk  = cswMcTest.generate_params('scs', 'csw', CSW_NUM_CONSTRAINTS, SEGMENT_SIZE)
+        certVk = certMcTest.generate_params('scs', 'cert', True, False, CERT_NUM_CONSTRAINTS, SEGMENT_SIZE)
+        cswVk  = cswMcTest.generate_params('scs', 'csw', True, False, CSW_NUM_CONSTRAINTS, SEGMENT_SIZE)
```
ptagl: What about using parameters name to improve readability?

```suggestion
        certVk = certMcTest.generate_params('scs', 'cert', constant = True, keyrot = False, CERT_NUM_CONSTRAINTS, SEGMENT_SIZE)
        cswVk  = cswMcTest.generate_params('scs', 'csw', constant = True, keyrot = False, CSW_NUM_CONSTRAINTS, SEGMENT_SIZE)
```

qa/rpc-tests/sc_cert_base.py
```diff
@@ -148,15 +148,15 @@ def run_test(self):
...
+        epoch_number, epoch_cum_tree_hash, _ = get_epoch_data(scid, self.nodes[0], EPOCH_LENGTH)
         mark_logs("epoch_number = {}, epoch_cum_tree_hash = {}".format(epoch_number, epoch_cum_tree_hash), self.nodes, DEBUG_MODE)
 
         addr_node1 = self.nodes[1].getnewaddress()
 
         #Create proof for WCert
         quality = 10
         proof = mcTest.create_test_proof(
-            "sc1", scid_swapped, epoch_number, quality, MBTR_SC_FEE, FT_SC_FEE, epoch_cum_tree_hash, constant, [addr_node1], [bwt_amount])
+            "sc1", scid_swapped, epoch_number, quality, MBTR_SC_FEE, FT_SC_FEE, epoch_cum_tree_hash, None, constant, [addr_node1], [bwt_amount])
```
ptagl: Can we use arguments name to avoid using `None` when calling `create_test_proof()` as in other tests?

That should apply also to other calls to the same function in this file.

```suggestion
            "sc1", scid_swapped, epoch_number, quality, MBTR_SC_FEE, FT_SC_FEE, epoch_cum_tree_hash, constant = constant, pks = [addr_node1], amounts = [bwt_amount])
```

qa/rpc-tests/test_framework/blockchainhelper.py
```diff
@@ -173,7 +173,7 @@ def create_sidechain(self, sc_name, sidechain_version, optional_parameters = {},
         and the version are mandatory.
         """
 
-        sc_input = SidechainCreationInput.from_rpc_args(self, sc_name, optional_parameters)
+        sc_input = SidechainCreationInput.from_rpc_args(self, sc_name, sidechain_version, optional_parameters)
         sc_input.version = sidechain_version
```
a-petrini: This should no longer be needed, as `from_rpc_args` passes the value `sidechain_version` to `SidechainCreationInput` constructor. All in all, maybe it was possible to correctly set the SC version by passing a proper key-value in the `optional_parameters` dictionary

qa/rpc-tests/sc_csw_memcleanup_split.py
```diff
@@ -152,15 +152,15 @@ def run_test(self):
...
+        epoch_number, epoch_cum_tree_hash, _ = get_epoch_data(scid, self.nodes[2], sc_epoch_len)
 
         bt_amount = Decimal("5.0")
         addr_node1 = self.nodes[1].getnewaddress()
         quality = 10
         scid_swapped = str(swap_bytes(scid))
 
         proof = certMcTest.create_test_proof(
-            "sc1", scid_swapped, epoch_number, quality, MBTR_SC_FEE, FT_SC_FEE, epoch_cum_tree_hash, constant, [addr_node1], [bt_amount])
+            "sc1", scid_swapped, epoch_number, quality, MBTR_SC_FEE, FT_SC_FEE, epoch_cum_tree_hash, None, constant, [addr_node1], [bt_amount])
```
ptagl:
```suggestion
            "sc1", scid_swapped, epoch_number, quality, MBTR_SC_FEE, FT_SC_FEE, epoch_cum_tree_hash, constant = constant, pks = [addr_node1], amounts = [bt_amount])
```

qa/rpc-tests/sc_getcertmaturityinfo.py
```diff
@@ -77,7 +77,7 @@ def run_test(self):
         mark_logs("Node 0 creates a certificate for the 3 sidechains", self.nodes, DEBUG_MODE)
         cert1_sc1: str = test_helper.send_certificate(v1_sc1_name, 10) # Epoch 0
         cert1_sc2: str = test_helper.send_certificate(v2_ceasing_sc2_name, 10) # Epoch 0
-        cert1_sc3: str = test_helper.send_certificate(v2_non_ceasing_sc3_name, 0) # Epoch 0
+        cert1_sc3: str = test_helper.send_certificate(v2_non_ceasing_sc3_name, 0, self.nodes[0].getblockcount()-1) # Epoch 0, reference one block in the past
```
ptagl, line 80: Why such a change was needed? Did the test stop working for any reason?

qa/rpc-tests/test_framework/blockchainhelper.py
```diff
@@ -304,9 +298,4 @@ def send_certificate(self, sc_name, quality, referenced_height = None):
             custom_fields,
             [RANDOM_BITVECTOR])
 
-        sc_info["last_certificate_epoch"] = epoch_number
-
-        if is_non_ceasing_sidechain:
-            sc_info["last_referenced_height"] = ref_height
-
```
a-petrini: Is there a reason for no longer updating there values in the `sc_info` dictionary after submitting the certificate?

qa/rpc-tests/sc_cert_getblocktemplate.py
```diff
@@ -185,16 +186,15 @@ def run_test_with_scversion(self, scversion, ceasable = True):
...
         else:
-            epoch_number = epoch_number + 1
-            current_reference_height = self.nodes[0].getblockcount()
-            epoch_cum_tree_hash = self.nodes[0].getblock(str(current_reference_height))['scCumTreeHash']
+            epoch_number, epoch_cum_tree_hash, prev_cert_data_hash = get_epoch_data(scid, self.nodes[0], epoch_length, not ceasable)
             proof = mcTest.create_test_proof(
-                sc_name, scid_swapped, epoch_number, quality, MBTR_SC_FEE, FT_SC_FEE, epoch_cum_tree_hash, constant,
-                [addr_node1], [bwt_amount])
+                sc_name, scid_swapped, epoch_number, quality, MBTR_SC_FEE, FT_SC_FEE, epoch_cum_tree_hash,
+                prev_cert_hash = prev_cert_data_hash if scversion >= 2 else None, constant = constant, pks = [addr_node1], amounts = [bwt_amount])
```
a-petrini: Thanks for uniforming the proof generation among SC versions. However, now the only difference is in the increase of quality for ceasable SC, so everything else can be removed from the `if..else` blocks

qa/rpc-tests/test_framework/blockchainhelper.py
```diff
@@ -89,7 +89,7 @@ def __init__(self, blockchainHelper, name):
     def is_non_ceasable(self) -> bool:
         return self.version == 2 and self.withdrawalEpochLength == 0
 
-    def from_rpc_args(blockchainHelper, name, args):
+    def from_rpc_args(blockchainHelper, name, version, args):
```
ptagl, line 92: Why do we need an additional `version` parameter? Can't we rely on the one already available in the `args` structure?

qa/rpc-tests/test_framework/blockchainhelper.py
ptagl: Considering the new behavior of the `SidechainCreationInput` contructor, can't we remove this line?

qa/rpc-tests/walletbackup.py
```diff
@@ -238,8 +238,21 @@ def run_test_with_scversion(self, scversion, ceasable = True):
...
+                    prev_cert_hash = None
+
+                proof = mcTest.create_test_proof(sc_name,
+                                                 scid_swapped,
+                                                 epoch_number,
+                                                 quality,
+                                                 MBTR_SC_FEE,
+                                                 FT_SC_FEE,
+                                                 epoch_cum_tree_hash,
+                                                 prev_cert_hash,
```
a-petrini: Is it possible to do as in [qa/rpc-tests/sc_cert_getblocktemplate.py]? This would also remove the if at lines 242-243
```suggestion
                                                 prev_cert_hash if scversion >=2 else None,
```

qa/rpc-tests/test_framework/util.py
```diff
@@ -611,12 +620,14 @@ def advance_epoch(mcTest, node, sync_call,
 
     sync_call()
 
-    epoch_number, epoch_cum_tree_hash = get_epoch_data(scid, node, epoch_length)
+    epoch_number, epoch_cum_tree_hash, _ = get_epoch_data(scid, node, epoch_length)
 
     scid_swapped = str(swap_bytes(scid))
 
+    print(constant)
```
ptagl: I guess this print was used only for debugging purpose, can we remove it now?

src/txmempool.cpp
```diff
@@ -2013,23 +2013,28 @@ bool CCoinsViewMemPool::GetSidechain(const uint256& scId, CSidechain& info) cons
...
         // certificates of later epochs.
-        if (info.isNonCeasing() && mempool.mapSidechains.at(scId).mBackwardCertificates.size() > 0)
-        {
-            const uint256& topQualHash = mempool.mapSidechains.at(scId).GetTopQualityCert()->second;
+        if (info.isNonCeasing() && !sc.mBackwardCertificates.empty()) {
+            const uint256& topQualHash = sc.GetTopQualityCert()->second;
             const CScCertificate& certTopQual = mempool.mapCertificate[topQualHash].GetCertificate();
+            info.lastTopQualityCertView.certDataHash = certTopQual.GetDataHash(info.fixedParams);
+            info.lastTopQualityCertView.forwardTransferScFee = certTopQual.forwardTransferScFee;
+            info.lastTopQualityCertView.mainchainBackwardTransferRequestScFee = certTopQual.mainchainBackwardTransferRequestScFee;
```
ptagl, line 2032: It's quite clear to me why we need to include the `certDataHash` into the sidechain info, but I don't understand why we do the same for `forwardTransferScFee` or `mainchainBackwardTransferRequestScFee`.

As far as I remember we only take into account the fees published with the last block and not the fees that are set by unconfirmed certificates.

src/sc/sidechaintypes.cpp
```diff
@@ -273,11 +278,18 @@ CFieldElement CFieldElement::ComputeHash(const CFieldElement& lhs, const CFieldE
...
+    CctpErrorCode err_code;
+    zendoo_serialize_field(zendoo_get_phantom_cert_data_hash(), buf.data(), &err_code);
+    if (err_code != CctpErrorCode::OK) {
+        LogPrintf("%s():%d - ERROR: could not create phantom hash, code[0x%x]\n", __func__, __LINE__, err_code);
+    }
+
+    phantom_element.SetByteArray(buf);
+    return phantom_element;
 }
 #endif
```
ptagl: I think that the new behavior of this function is more correct than the previous one, but I wonder if it breaks the compatibility with previous versions.

We should perform additional checks to make sure that it's safe to do so.

qa/rpc-tests/sc_cert_nonceasing.py
```diff
@@ -46,13 +46,13 @@ def setup_network(self, split=False):
...
-        self.is_network_split = split
         self.sync_all()
 
     def try_send_certificate(self, node_idx, scid, epoch_number, quality, ref_height, mbtr_fee, ft_fee, bt, expect_failure, failure_reason=None):
         scid_swapped = str(swap_bytes(scid))
-        ep, epoch_cum_tree_hash = get_epoch_data(scid, self.nodes[node_idx], 0, True, ref_height)
-        proof = self.mcTest.create_test_proof(PARAMS_NAME, scid_swapped, epoch_number, quality, mbtr_fee, ft_fee, epoch_cum_tree_hash, self.constant, [bt["address"]], [bt["amount"]])
+        ep, epoch_cum_tree_hash, prev_cert_hash = get_epoch_data(scid, self.nodes[node_idx], 0, True, ref_height)
+        proof = self.mcTest.create_test_proof(PARAMS_NAME, scid_swapped, epoch_number, quality, mbtr_fee, ft_fee, epoch_cum_tree_hash, prev_cert_hash,
+            constant = self.constant, pks = [bt["address"]], amounts = [bt["amount"]])
```
a-petrini:
```suggestion
        proof = self.mcTest.create_test_proof(PARAMS_NAME, scid_swapped, epoch_number, quality, mbtr_fee, ft_fee, epoch_cum_tree_hash, prev_cert_hash = prev_cert_hash,
            constant = self.constant, pks = [bt["address"]], amounts = [bt["amount"]])
```

src/gtest/tx_creation_utils.cpp
```diff
@@ -753,7 +753,7 @@ void BlockchainTestManager::GenerateSidechainTestParameters(ProvingSystem provin
 {
     CctpErrorCode errorCode;
     zendoo_generate_mc_test_params(
-        circuitType, provingSystem, 1 << 10,
+        circuitType, provingSystem, 1 << 10, false, //FIXME?
```
ptagl: Maybe we can expose the argument `with_prev_hash` (currently set to false) to the callers of `GenerateSidechainTestParameters()` as optional.

That would also let us to extend the unit tests currently available in `src/gtest/test_libzendoo.cpp`.

qa/rpc-tests/sc_keyrot.py
```diff
@@ -0,0 +1,226 @@
...
+            "amount": Decimal("1"),
+            "wCertVk": vk,
+            "constant": self.constant,
+        }
+        return self.nodes[0].sc_create(cmdInput)
+
+
+    def run_test(self):
+        '''
+        '''
```
ptagl: Can we add a summary description of what is tested here?

qa/rpc-tests/test_framework/mc_test/mc_test.py
drgora: Applied your suggestion to 3518be16ce52befbdfc21398187b4961582107f2

qa/rpc-tests/test_framework/blockchainhelper.py
drgora: My mistake, thanks for catching this. Restored in 3518be16ce52befbdfc21398187b4961582107f2

qa/rpc-tests/test_framework/blockchainhelper.py
drgora: I removed line 177 in f1aa8bb237c592621a4f635666ee5a4761bbb988. We may update to only use `optional_parameters` in future.

qa/rpc-tests/sc_cert_getblocktemplate.py
drgora: Nice catch! Done in f1aa8bb237c592621a4f635666ee5a4761bbb988, thanks.

qa/rpc-tests/walletbackup.py
drgora: Done in f1aa8bb237c592621a4f635666ee5a4761bbb988

qa/rpc-tests/test_framework/mc_test/mc_test.py
drgora: As you suggest, when `constant == False` you actually use a constant. That is counterintuitive, and I updated `mc_test.py` to properly handle `False` as `None` in f1aa8bb237c592621a4f635666ee5a4761bbb988. Thanks!

qa/rpc-tests/sc_keyrot.py
```diff
@@ -0,0 +1,226 @@
...
+                                  epoch_number,
+                                  quality,
+                                  self.nodes[0].getblockcount(),
+                                  MBTR_SC_FEE,
+                                  FT_SC_FEE,
+                                  amount_cert_1,
+                                  0, # 0 means that PHANTOM_HASH will be used
+                                  False
+                                 )
+
```
ptagl, line 237: I see here that we test passing the PHANTOM_HASH for the proof of the first certificate. Should we add more tests to cover also the generation of proof for certificates with epoch other than the first one (so that we don't pass the PHANTOM_HASH) or this scenario is already taken into account in other Python tests?

ptagl (changes requested):

qa/rpc-tests/sc_getcertmaturityinfo.py
ptagl: Already addressed with [this comment](https://github.com/HorizenOfficial/zen/pull/481#discussion_r999213311).

qa/rpc-tests/test_framework/blockchainhelper.py
ptagl: Already addressed with [this comment](https://github.com/HorizenOfficial/zen/pull/481#discussion_r999213311).

qa/rpc-tests/test_framework/blockchainhelper.py
ptagl: Already addressed with [this comment](https://github.com/HorizenOfficial/zen/pull/481#discussion_r999213311).

qa/rpc-tests/sc_keyrot.py
drgora: Such scenario is indeed tested in all the other tests that send v2 sc certificates, such as `sc_cert_nonceasing.py` and `sc_getcertmaturityinfo.py`.

qa/rpc-tests/sc_big_block.py
drgora: Good point, done with f663cac6ada8f2b0e099497b23ac96371e611b14

qa/rpc-tests/test_framework/util.py
drgora: Right, this should have been removed, nice catch! f663cac6ada8f2b0e099497b23ac96371e611b14

src/sc/sidechaintypes.cpp
drgora: You're right,this indeed breaks compatibility. f663cac6ada8f2b0e099497b23ac96371e611b14 restores this zero hash for compatibility, and also implements `GetPhantomHash` properly to return the phantom hash as expected.

src/gtest/tx_creation_utils.cpp
drgora: Good point, I implemented your suggestion in f663cac6ada8f2b0e099497b23ac96371e611b14, which also adds key rotation specific tests in `test_libzendoo.cpp`

qa/rpc-tests/sc_keyrot.py
drgora: Woops, right, I forgot this, thanks. Added in f663cac6ada8f2b0e099497b23ac96371e611b14

qa/rpc-tests/sc_cert_base.py
drgora: @a-petrini did this in 64a018ce829d1a900692a9e41d2589cbb3518939, thanks a lot!

qa/rpc-tests/sc_csw_memcleanup_split.py
drgora: @a-petrini did this in 64a018ce829d1a900692a9e41d2589cbb3518939, thanks a lot!

ptagl (approved):

PR #485, opened 2022-10-26
Ap/sc fees list with pointers

a-petrini:
This pull request updates the previous one dealing with the management of scFees list for v2 non-ceasible sidechains.
The major updates deals with getting rid of the duplicate `scFees_v2` list, in favor of using a unified container for all the sidechains, regardless of their version.
The scFees list now cointains pointers to `Sidechain::ScFeeData` structs, which can be polymorphically instantiated as `Sidechain::ScFeeData_v2` elements at runtime, depending on the sidechain version.

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -1660,8 +1660,8 @@ bool FillScRecordFromInfo(const uint256& scId, const CSidechain& info, CSidechai
             UniValue o(UniValue::VOBJ);
             o.pushKV("forwardTxScFee", ValueFromAmount(entry->forwardTxScFee));
             o.pushKV("mbtrTxScFee", ValueFromAmount(entry->mbtrTxScFee));
-            Sidechain::ScFeeData_v2 *casted_entry = dynamic_cast<Sidechain::ScFeeData_v2*>(entry.get());
-            if (casted_entry != nullptr) {
+            std::shared_ptr<Sidechain::ScFeeData_v2> casted_entry = dynamic_pointer_cast<Sidechain::ScFeeData_v2>(entry);
+            if (casted_entry) {
                 o.pushKV("submissionHeight", casted_entry->submissionHeight);
             }
             sf.push_back(std::move(o));
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -1654,26 +1654,19 @@ bool FillScRecordFromInfo(const uint256& scId, const CSidechain& info, CSidechai
         sc.pushKV("immatureAmounts", ia);
 
         UniValue sf(UniValue::VARR);
-        // v2 non-ceasable sc
-        if (info.isNonCeasing()) {
-            for(const auto& entry: info.scFees_v2)
-            {
-                UniValue o(UniValue::VOBJ);
-                o.pushKV("forwardTxScFee", ValueFromAmount(entry.forwardTxScFee));
-                o.pushKV("mbtrTxScFee", ValueFromAmount(entry.mbtrTxScFee));
-                o.pushKV("submissionHeight", entry.submissionHeight);
-                sf.push_back(std::move(o));
+
+        for(const auto& entry: info.scFees)
+        {
+            UniValue o(UniValue::VOBJ);
+            o.pushKV("forwardTxScFee", ValueFromAmount(entry->forwardTxScFee));
+            o.pushKV("mbtrTxScFee", ValueFromAmount(entry->mbtrTxScFee));
+            Sidechain::ScFeeData_v2 *casted_entry = dynamic_cast<Sidechain::ScFeeData_v2*>(entry.get());
+            if (casted_entry != nullptr) {
+                o.pushKV("submissionHeight", casted_entry->submissionHeight);
             }
+            sf.push_back(std::move(o));
         }
-        else {
-            for(const auto& entry: info.scFees)
-            {
-                UniValue o(UniValue::VOBJ);
-                o.pushKV("forwardTxScFee", ValueFromAmount(entry.forwardTxScFee));
-                o.pushKV("mbtrTxScFee", ValueFromAmount(entry.mbtrTxScFee));
-                sf.push_back(std::move(o));
-            }
-        }
+
         sc.pushKV("scFees", sf);
 
         // get unconfirmed data if any
--- a/src/sc/sidechain.cpp
+++ b/src/sc/sidechain.cpp
@@ -300,7 +300,7 @@ std::string CSidechain::ToString() const
 }
 
 size_t CSidechain::DynamicMemoryUsage() const {
-    return memusage::DynamicUsage(mImmatureAmounts) + memusage::DynamicUsage(scFees) + memusage::DynamicUsage(scFees_v2);
+    return memusage::DynamicUsage(mImmatureAmounts) + memusage::DynamicUsage(scFees);
 }
 
 size_t CSidechainEvents::DynamicMemoryUsage() const {
@@ -748,14 +748,13 @@ void CSidechain::InitScFees()
         LogPrint("sc", "%s():%d - maxSizeOfScFeesContainers set to %d\n", __func__, __LINE__, maxSizeOfScFeesContainers);
 
         assert(scFees.empty());
-        assert(scFees_v2.empty());
 
         if (!isNonCeasing()) {
-            scFees.emplace_back(lastTopQualityCertView.forwardTransferScFee,
-                                lastTopQualityCertView.mainchainBackwardTransferRequestScFee);
+            scFees.emplace_back(new Sidechain::ScFeeData(lastTopQualityCertView.forwardTransferScFee,
+                                lastTopQualityCertView.mainchainBackwardTransferRequestScFee));
         } else {
-            scFees_v2.emplace_back(lastTopQualityCertView.forwardTransferScFee,
-                                   lastTopQualityCertView.mainchainBackwardTransferRequestScFee, lastReferencedHeight);
+            scFees.emplace_back(new Sidechain::ScFeeData_v2(lastTopQualityCertView.forwardTransferScFee,
+                                lastTopQualityCertView.mainchainBackwardTransferRequestScFee, lastReferencedHeight));
         }
     }
 }
@@ -773,7 +772,7 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView, int blockHeigh
 
     // v1 sidechains
     if (!isNonCeasing()) {
-        scFees.emplace_back(ftScFee, mbtrScFee);
+        scFees.emplace_back(new Sidechain::ScFeeData(ftScFee, mbtrScFee));
 
         // remove from the front as many elements are needed to be within the circular buffer size
         // --
@@ -784,85 +783,83 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView, int blockHeigh
         {
             const auto& entry = scFees.front();
             LogPrint("sc", "%s():%d - popping f=%d, m=%d from list, as scFees maxsize has been reached\n",
-                __func__, __LINE__, entry.forwardTxScFee, entry.mbtrTxScFee);
+                __func__, __LINE__, entry->forwardTxScFee, entry->mbtrTxScFee);
             scFees.pop_front();
         }
     }
     // v2 sidechains
     else {
-        auto scFeeIt = std::find_if(scFees_v2.begin(), scFees_v2.end(),
-            [&blockHeight](Sidechain::ScFeeData_v2 scFeeElem) { return scFeeElem.submissionHeight == blockHeight; });
+        auto scFeeIt = std::find_if(scFees.begin(), scFees.end(),
+            [&blockHeight](const std::shared_ptr<Sidechain::ScFeeData> & scFeeElem) {
+                return std::dynamic_pointer_cast<Sidechain::ScFeeData_v2>(scFeeElem)->submissionHeight == blockHeight;
+            });
 
         // We have not found a scFeeData for the current height, so we add a new one and perform all the checks
         // on the container size / element age
-        if (scFeeIt == scFees_v2.end()) {
-            scFees_v2.emplace_back(ftScFee, mbtrScFee, blockHeight);
+        if (scFeeIt->get() == scFees.end()->get()) {
+           scFees.emplace_back(new Sidechain::ScFeeData_v2(ftScFee, mbtrScFee, blockHeight));
 
             // as for v1 sidechains
             const size_t max_size { static_cast<size_t>(maxSizeOfScFeesContainers) };
-            while (scFees_v2.size() > max_size)
+
+            while (scFees.size() > max_size)
             {
-                const auto& entry = scFees_v2.front();
+                const std::shared_ptr<Sidechain::ScFeeData_v2> entry = std::dynamic_pointer_cast<Sidechain::ScFeeData_v2>(scFees.front());
+
                 LogPrint("sc", "%s():%d - popping f=%d, m=%d, h=%d from list, as scFees maxsize has been reached\n",
-                    __func__, __LINE__, entry.forwardTxScFee, entry.mbtrTxScFee, entry.submissionHeight);
-                scFees_v2.pop_front();
+                    __func__, __LINE__, entry->forwardTxScFee, entry->mbtrTxScFee, entry->submissionHeight);
+                scFees.pop_front();
             }
 
             // Also purge all elements submitted within a block that now is too old to be considered
             const auto threshold{blockHeight - this->maxSizeOfScFeesContainers};
-            scFees_v2.remove_if([threshold](Sidechain::ScFeeData_v2 entry) {
-                const bool testCondition = (entry.submissionHeight <= threshold);
+            scFees.remove_if([threshold](std::shared_ptr<Sidechain::ScFeeData> entry) {
+                const std::shared_ptr<Sidechain::ScFeeData_v2> casted_entry = std::dynamic_pointer_cast<Sidechain::ScFeeData_v2>(entry);
+                const bool testCondition = (casted_entry->submissionHeight <= threshold);
                 if (testCondition)
                     LogPrint("sc", "%s():%d - popping f=%d, m=%d, h=%d from list, as entry is too old (threshold height was %d)\n",
-                        __func__, __LINE__, entry.forwardTxScFee, entry.mbtrTxScFee, entry.submissionHeight, threshold);
+                        __func__, __LINE__, casted_entry->forwardTxScFee, casted_entry->mbtrTxScFee,
+                        casted_entry->submissionHeight, threshold);
                 return testCondition;
             });
+
         }
         // On the other hand, if we found a scFeeData element for the current height, we just update it with lower
         // ft and/or mbtr fee rates. All the checks on the container size and element age have been performed
         // at the moment of the element insertion, hence their are not required here
         else {
-            scFeeIt->forwardTxScFee = std::min(scFeeIt->forwardTxScFee, ftScFee);
-            scFeeIt->mbtrTxScFee = std::min(scFeeIt->mbtrTxScFee, mbtrScFee);
+            scFeeIt->get()->forwardTxScFee = std::min(scFeeIt->get()->forwardTxScFee, ftScFee);
+            scFeeIt->get()->mbtrTxScFee    = std::min(scFeeIt->get()->mbtrTxScFee, mbtrScFee);
         }
+
     }
 }
 
 void CSidechain::DumpScFees() const
 {
-    if (!isNonCeasing()) {
-        for (const auto& entry : scFees) {
-            std::cout << "[" << std::setw(2) << entry.forwardTxScFee
-                      << "/" << std::setw(2) << entry.mbtrTxScFee << "]";
+
+    for (const auto& entry : scFees) {
+        std::cout << "[" << std::setw(2) << entry->forwardTxScFee
+                  << "/" << std::setw(2) << entry->mbtrTxScFee;
+        const std::shared_ptr<Sidechain::ScFeeData_v2> casted_entry = std::dynamic_pointer_cast<Sidechain::ScFeeData_v2>(entry);
+        if (casted_entry != nullptr) {
+            std::cout << "/" << std::setw(2) << casted_entry->submissionHeight;
         }
+        std::cout << "]" << std::endl;
     }
-    else {
-        for (const auto& entry : scFees_v2) {
-            std::cout << "[" << std::setw(2) << entry.forwardTxScFee
-                      << "/" << std::setw(2) << entry.mbtrTxScFee
-                      << "/" << std::setw(2) << entry.submissionHeight << "]";
-        }
-    }
-    std::cout << std::endl;
+
 }
 
 CAmount CSidechain::GetMinFtScFee() const
 {
-    CAmount minScFee;
-    if (!isNonCeasing()) {
-        assert(!scFees.empty());
-        minScFee = std::min_element(
-            scFees.begin(), scFees.end(),
-            [] (const Sidechain::ScFeeData& a, const Sidechain::ScFeeData& b) { return a.forwardTxScFee < b.forwardTxScFee; }
-        )->forwardTxScFee;
-    }
-    else {
-        assert(!scFees_v2.empty());
-        minScFee = std::min_element(
-            scFees_v2.begin(), scFees_v2.end(),
-            [] (const Sidechain::ScFeeData_v2& a, const Sidechain::ScFeeData_v2& b) { return a.forwardTxScFee < b.forwardTxScFee; }
-        )->forwardTxScFee;
-    }
+    assert(!scFees.empty());
+
+    CAmount minScFee = std::min_element(
+        scFees.begin(), scFees.end(),
+        [] (const std::shared_ptr<Sidechain::ScFeeData> &a, const std::shared_ptr<Sidechain::ScFeeData> &b) {
+            return a->forwardTxScFee < b->forwardTxScFee;
+        }
+    )->get()->forwardTxScFee;
 
     LogPrint("sc", "%s():%d - returning min=%lld\n", __func__, __LINE__, minScFee);
     return minScFee;
@@ -870,21 +867,14 @@ CAmount CSidechain::GetMinFtScFee() const
 
 CAmount CSidechain::GetMinMbtrScFee() const
 {
-    CAmount minScFee;
-    if (!isNonCeasing()) {
-        assert(!scFees.empty());
-        minScFee = std::min_element(
-            scFees.begin(), scFees.end(),
-            [] (const Sidechain::ScFeeData& a, const Sidechain::ScFeeData& b) { return a.mbtrTxScFee < b.mbtrTxScFee; }
-        )->mbtrTxScFee;
-    }
-    else {
-        assert(!scFees_v2.empty());
-        minScFee = std::min_element(
-            scFees_v2.begin(), scFees_v2.end(),
-            [] (const Sidechain::ScFeeData_v2& a, const Sidechain::ScFeeData_v2& b) { return a.mbtrTxScFee < b.mbtrTxScFee; }
-        )->mbtrTxScFee;
-    }
+    assert(!scFees.empty());
+
+    CAmount minScFee = std::min_element(
+        scFees.begin(), scFees.end(),
+        [] (const std::shared_ptr<Sidechain::ScFeeData> &a, const std::shared_ptr<Sidechain::ScFeeData> &b) {
+            return a->mbtrTxScFee < b->mbtrTxScFee;
+        }
+    )->get()->mbtrTxScFee;
 
     LogPrint("sc", "%s():%d - returning min=%lld\n", __func__, __LINE__, minScFee);
     return minScFee;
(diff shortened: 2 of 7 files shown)


src/sc/sidechain.cpp
```diff
@@ -782,107 +781,98 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView, int blockHeigh
...
         }
     }
     // v2 sidechains
     else {
-        auto scFeeIt = std::find_if(scFees_v2.begin(), scFees_v2.end(),
-            [&blockHeight](Sidechain::ScFeeData_v2 scFeeElem) { return scFeeElem.submissionHeight == blockHeight; });
+        auto scFeeIt = std::find_if(scFees.begin(), scFees.end(),
+            [&blockHeight](const std::shared_ptr<Sidechain::ScFeeData> & scFeeElem) {
+                return std::dynamic_pointer_cast<Sidechain::ScFeeData_v2>(scFeeElem).get()->submissionHeight == blockHeight;
+            });
```
ptagl: If `scFees` is an ordered list (by height) can't we just compare `blockHeight` with the `submissionHeight` of the last element of the list without traversing the whole list with the `find_if?`

src/sc/sidechain.cpp
```diff
@@ -782,107 +781,98 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView, int blockHeigh
...
+            scFees.remove_if([threshold](std::shared_ptr<Sidechain::ScFeeData> entry) {
+                std::shared_ptr<Sidechain::ScFeeData_v2> casted_entry = std::dynamic_pointer_cast<Sidechain::ScFeeData_v2>(entry);
+                const bool testCondition = (casted_entry->submissionHeight <= threshold);
                 if (testCondition)
                     LogPrint("sc", "%s():%d - popping f=%d, m=%d, h=%d from list, as entry is too old\n",
-                        __func__, __LINE__, entry.forwardTxScFee, entry.mbtrTxScFee, entry.submissionHeight);
+                        __func__, __LINE__, casted_entry->forwardTxScFee, casted_entry->mbtrTxScFee,
+                        casted_entry->submissionHeight);
                 return testCondition;
             });
```
ptagl: Using `remove_if()` makes the code very clear, I wonder if that might have an impact on performance: probably that function iterates over all the elements of the list, while by iterating manually we could stop as soon as we find an entry with `submissionHeight` higher than the threshold.

src/sc/sidechain.cpp
```diff
@@ -782,107 +781,98 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView, int blockHeigh
...
+        auto scFeeIt = std::find_if(scFees.begin(), scFees.end(),
+            [&blockHeight](const std::shared_ptr<Sidechain::ScFeeData> & scFeeElem) {
+                return std::dynamic_pointer_cast<Sidechain::ScFeeData_v2>(scFeeElem).get()->submissionHeight == blockHeight;
+            });
 
         // We have not found a scFeeData for the current height, so we add a new one and perform all the checks
         // on the container size / element age
-        if (scFeeIt == scFees_v2.end()) {
-            scFees_v2.emplace_back(ftScFee, mbtrScFee, blockHeight);
+        if (scFeeIt->get() == scFees.end()->get()) {
```
ptagl: Would it be possible to get rid of the call to `get()` and replace the check with something like `scFeeIt == scFees.end()` or `scFeeIt == std::end(scFees)`?

src/sc/sidechain.cpp
```diff
@@ -782,107 +781,98 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView, int blockHeigh
...
-        )->forwardTxScFee;
-    }
+    assert(!scFees.empty());
+
+    CAmount minScFee = std::min_element(
+        scFees.begin(), scFees.end(),
+        [] (const std::shared_ptr<Sidechain::ScFeeData> &a, const std::shared_ptr<Sidechain::ScFeeData> &b) {
+            return a->forwardTxScFee < b->forwardTxScFee;
+        }
+    )->get()->forwardTxScFee;
```
ptagl, line 848:
```suggestion
    )->forwardTxScFee;
```

src/sc/sidechain.cpp
```diff
@@ -782,107 +781,98 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView, int blockHeigh
...
-        )->mbtrTxScFee;
-    }
+    assert(!scFees.empty());
+
+    CAmount minScFee = std::min_element(
+        scFees.begin(), scFees.end(),
+        [] (const std::shared_ptr<Sidechain::ScFeeData> &a, const std::shared_ptr<Sidechain::ScFeeData> &b) {
+            return a->mbtrTxScFee < b->mbtrTxScFee;
+        }
+    )->get()->mbtrTxScFee;
```
ptagl, line 863:
```suggestion
    )->mbtrTxScFee;
```

src/sc/sidechain.cpp
```diff
@@ -300,7 +300,7 @@ std::string CSidechain::ToString() const
 }
 
 size_t CSidechain::DynamicMemoryUsage() const {
-    return memusage::DynamicUsage(mImmatureAmounts) + memusage::DynamicUsage(scFees) + memusage::DynamicUsage(scFees_v2);
+    return memusage::DynamicUsage(mImmatureAmounts) + memusage::DynamicUsage(scFees);
```
JackPiri, line 303: It seems to me `memusage::DynamicUsage(scFees)` would return a `size_t` value computed as `scFees.size()` multiplied by `sizeof(std::shared_ptr<Sidechain::ScFeeData>)`, whilst what we actually want to get is `scFees.size()` multiplied by `sizeof(Sidechain::ScFeeData)`; am I correct?

src/sc/sidechain.cpp
```diff
@@ -782,107 +781,98 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView, int blockHeigh
...
+                __func__, __LINE__, entry->forwardTxScFee, entry->mbtrTxScFee);
             scFees.pop_front();
         }
     }
     // v2 sidechains
     else {
-        auto scFeeIt = std::find_if(scFees_v2.begin(), scFees_v2.end(),
-            [&blockHeight](Sidechain::ScFeeData_v2 scFeeElem) { return scFeeElem.submissionHeight == blockHeight; });
+        auto scFeeIt = std::find_if(scFees.begin(), scFees.end(),
+            [&blockHeight](const std::shared_ptr<Sidechain::ScFeeData> & scFeeElem) {
```
JackPiri, line 793: What about adding a check on the cast result inside the lambda?
Something like:

```
[&blockHeight](const std::shared_ptr<Sidechain::ScFeeData> & scFeeElem) {
    std::shared_ptr<Sidechain::ScFeeData_v2> casted_scFeeElem = std::dynamic_pointer_cast<Sidechain::ScFeeData_v2>(scFeeElem);
    if (casted_entry)
        return casted_entry.get()->submissionHeight == blockHeight;
    return false;
}
```

I see there is a check on the non ceasability of the SC, but maybe this option can be more robust in case of future extensions/developments
An analogous case is at line #815

JackPiri (commented):
Please consider also the comments provided in PR #478 ("added sc v2 support for scFees list")

src/sc/sidechain.cpp
AndreaLanfranchi: @JackPiri why do you think that ?
From what I see scFees is defined as `std::list<Sidechain::ScFeeData>` (no shared ptr) and the template invoked to compute the allocated size does correctly the multiplication of number of elements in list by size of item type.
Can you expand your doubts ?

src/rpc/blockchain.cpp
```diff
@@ -1654,26 +1654,19 @@ bool FillScRecordFromInfo(const uint256& scId, const CSidechain& info, CSidechai
...
-                o.pushKV("forwardTxScFee", ValueFromAmount(entry.forwardTxScFee));
-                o.pushKV("mbtrTxScFee", ValueFromAmount(entry.mbtrTxScFee));
+
+        for(const auto& entry: info.scFees)
+        {
+            UniValue o(UniValue::VOBJ);
+            o.pushKV("forwardTxScFee", ValueFromAmount(entry.forwardTxScFee));
+            o.pushKV("mbtrTxScFee", ValueFromAmount(entry.mbtrTxScFee));
+
+            if (info.fixedParams.version == 2)
```
AndreaLanfranchi: I have a doubt here.
As this test replaces the former `info.isNonCeasing` I don't think `info.fixedParams.version == 2` is enough to define a non-ceasing SC. The full control should be `version == 2 AND withdrawalepochlength == 0`.
Am I missing something ? Is this new version of check intentional ?

See as reference
```
  static bool isNonCeasingSidechain(int version, int withdrawalEpochLength) {
        // SC v2 can be either ceasing (as v0 and v1), or non-ceasing.
        // Non ceasing sidechains are identified by fixedParams.withdrawalEpochLength == 0, which is not allowed for
        // other sidechain versions.
        return version == 2 && withdrawalEpochLength == 0;
    }
```

src/sc/sidechain.cpp
```diff
@@ -746,15 +746,9 @@ void CSidechain::InitScFees()
...
-        assert(scFees_v2.empty());
 
-        if (!isNonCeasing()) {
-            scFees.emplace_back(lastTopQualityCertView.forwardTransferScFee,
-                                lastTopQualityCertView.mainchainBackwardTransferRequestScFee);
-        } else {
-            scFees_v2.emplace_back(lastTopQualityCertView.forwardTransferScFee,
-                                   lastTopQualityCertView.mainchainBackwardTransferRequestScFee, lastReferencedHeight);
-        }
+        scFees.push_back(Sidechain::ScFeeData(lastTopQualityCertView.forwardTransferScFee,
```
AndreaLanfranchi: Prefer `emplace_back` for a direct in-container constructor (avoids a temporary and copy/move ctor).

src/sc/sidechain.cpp
```diff
@@ -771,7 +765,7 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView, int blockHeigh
 
     // v1 sidechains
     if (!isNonCeasing()) {
-        scFees.emplace_back(ftScFee, mbtrScFee);
+        scFees.push_back(Sidechain::ScFeeData(ftScFee, mbtrScFee, blockHeight));
```
AndreaLanfranchi: Same as comment before

src/sc/sidechain.cpp
JackPiri: @AndreaLanfranchi my comment was referring to commit [b3121515ea35d5d6f5ec42dbf58c9cf91adef294](https://github.com/HorizenOfficial/zen/pull/485/commits/b3121515ea35d5d6f5ec42dbf58c9cf91adef294) where member `scFees` was of type `std::list<std::shared_ptr<Sidechain::ScFeeData>>`. In that case the size of item is the size of a `std::shared_ptr`, hence the size of the actual `Sidechain::ScFeeData` is not taken in account (and that seems incorrect to me).
A check is needed to understand if this conversation is automatically resolved by newer commit [ee234e4b836f0a54932d19fac597b7db337355df](https://github.com/HorizenOfficial/zen/pull/485/commits/ee234e4b836f0a54932d19fac597b7db337355df); in addition I know @drgora is working on another version for this feature, so maybe let's just wait for it and later we'll see if this comment gets outdated or not.

AndreaLanfranchi: Ah ok ... so then can we consider this resolved ?

src/rpc/blockchain.cpp
a-petrini: Yes, you are right: we should check also for `withdrawalEpochLength` being 0, as this could create inconsistencies. I will check also the other occurrences of this check over the rest of this PR and the original branch, which this one was derived from.

src/sc/sidechain.cpp
drgora: This is indeed a very good point. I agree with @JackPiri, the last commit uses `shared_ptr`s, and in its current state the code does not compute dynamic memory correctly.

I was about to fix this by adding proper smart pointers handlers in `src/memusage.h`, but I noticed that in our current codebase most of the things are not computed correctly anyway.

Lines 38-44 state:
```
/** Compute the memory used for dynamically allocated but owned data structures.
 *  For generic data types, this is *not* recursive. DynamicUsage(vector<vector<int> >)
 *  will compute the memory used for the vector<int>'s, but not for the ints inside.
 *  This is for efficiency reasons, as these functions are intended to be fast. If
 *  application data structures require more accurate inner accounting, they should
 *  iterate themselves, or use more efficient caching + updating on modification.
 */
```
I'm not sure the claim on performance reason is valid as of today. Also, I'm not even sure that in general the code in `src/memusage.h` is correct and useful. For example, we make empiric assumptions based on glibc 2.19, but we might run `zend` on very different configurations.

All in all, I see 3 options here:
a) leave everything as it is and stick to the approach suggested in `src/memusage.h`, which means that the mem usage is (slightly) underestimated for this `scFees` list
b) implement accurate measurement for this specific list in `src/sc/sidechain.cpp` for this specific list
c) update `src/memusage.h` to compute mem usage accurately for all stl/template data structures

I personally would go for option `a` in this PR, but definitely go for `c` in another PR.

What do you guys think?

src/sc/sidechain.cpp
```diff
@@ -783,108 +782,91 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView, int blockHeigh
...
         {
             const auto& entry = scFees.front();
-            LogPrint("sc", "%s():%d - popping f=%d, m=%d from list, as scFees maxsize has been reached\n",
-                __func__, __LINE__, entry.forwardTxScFee, entry.mbtrTxScFee);
+            LogPrint("sc", "%s():%d - popping %s from list, as scFees maxsize has been reached\n",
+                __func__, __LINE__, entry->ToString().c_str());
             scFees.pop_front();
         }
     }
     // v2 sidechains
```
ptagl: Does this branch only apply to **non-ceasable** v2 sidechains? It should be fixed, and the first comment as well (v1 sidechains).

src/sc/sidechain.cpp
a-petrini: I agree with @drgora observation, and I vote for option `a` for this PR, and update / redesign the memory usage tracking system for its own dedicated task, as it seems that the information collected by the functions in the `memusage::` namespace are, among others, used in verifying the DB, and are part of the logic that decides when cache should flushed to disk.

src/sc/sidechain.cpp
```diff
@@ -783,108 +782,91 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView, int blockHeigh
...
             {
-                const auto& entry = scFees_v2.front();
-                LogPrint("sc", "%s():%d - popping f=%d, m=%d, h=%d from list, as scFees maxsize has been reached\n",
-                    __func__, __LINE__, entry.forwardTxScFee, entry.mbtrTxScFee, entry.submissionHeight);
-                scFees_v2.pop_front();
+                const auto& entry = scFees.front();
+                LogPrint("sc", "%s():%d - popping %s from list, as scFees maxsize has been reached\n",
+                    __func__, __LINE__, entry->ToString().c_str());
+                scFees.pop_front();
             }
```
ptagl: Is this section run also for _non-ceasable_ sidechains? If so, is it wanted?

I mean that the removal of all the old entries through the threshold mechanism should already prevent the list from growing beyond the max size, so the `while (scFees.size() > max_size)` might be useless.

src/undo.h
```diff
@@ -305,6 +311,16 @@ struct CSidechainUndoData
...
+            res += strprintf("scFees.size()=%u\n", scFees.size());
+            for(const auto& entry: scFees) {
+                res += strprintf("scFtFee=%d.%08d - ", entry->forwardTxScFee / COIN, entry->forwardTxScFee % COIN);
+                res += strprintf("scMbtrFee=%d.%08d\n", entry->mbtrTxScFee / COIN, entry->mbtrTxScFee % COIN);
+                // Only for v2 non-ceasable sidechains
+                std::shared_ptr<Sidechain::ScFeeData_v2> casted_entry = std::dynamic_pointer_cast<Sidechain::ScFeeData_v2>(entry);
+                if (casted_entry) {
+                    res += strprintf("submissionHeight=%d\n", casted_entry->submissionHeight);
+                }
+            }
```
JackPiri, line 324: Maybe I am missing something but this logical flow seems quite different to me compared to the previous one: currently, in case the condition `(contentBitMask & AvailableSections::CROSS_EPOCH_CERT_DATA)` is false nothing is printed, while previously there was an if-else logical flow and at least `scFtFee` and `scMbtrFee` were always printed.
Has something changed regarding this?

src/undo.h
```diff
@@ -186,6 +185,8 @@ struct CSidechainUndoData
         if (contentBitMask & AvailableSections::CROSS_EPOCH_CERT_DATA)
         {
             totalSize += ::GetSerializeSize(pastEpochTopQualityCertView, nType, nVersion);
+            // TODO: fix this condition for counting scFees list size
```
JackPiri, line 188: Can this `TODO` be addressed?

src/sc/sidechain.cpp
JackPiri: Seems resonable to me to go for option `a)` in this PR; I agree with @drgora regarding the doubt associated to the performance claim (line 38->44).
Maybe a story/task could be created to track future refactoring of this module (@drgora).

src/serialize.h
```diff
@@ -189,7 +191,8 @@ enum
     SER_GETHASH         = (1 << 2),
 };
 
-#define READWRITE(obj)      (::SerReadWrite(s, (obj), nType, nVersion, ser_action))
+#define READWRITE(obj)                      (::SerReadWrite(s, (obj), nType, nVersion, ser_action))
+#define READWRITE_POLYMORPHIC(obj, b, d)    (::SerReadWrite<decltype(s), b, decltype(obj), d>(s, (obj), nType, nVersion, ser_action))
```
JackPiri: Personally I'd prefer making parameters `b` and `d` names more self-explaining; I guess they stands for `base` and `derived` (correct?), but why not putting some more clear names?

src/serialize.h
```diff
@@ -1046,6 +1106,19 @@ inline void SerReadWrite(Stream& s, T& obj, int nType, int nVersion, CSerActionU
 }
 
 
+template<typename Stream, typename T, typename C, typename D>
```
JackPiri: An analogous consideration regarding naming can be done for `T`, `C` and `D`

src/sc/sidechain.cpp
```diff
@@ -783,108 +782,91 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView, int blockHeigh
         while (scFees.size() > max_size)
         {
             const auto& entry = scFees.front();
-            LogPrint("sc", "%s():%d - popping f=%d, m=%d from list, as scFees maxsize has been reached\n",
-                __func__, __LINE__, entry.forwardTxScFee, entry.mbtrTxScFee);
+            LogPrint("sc", "%s():%d - popping %s from list, as scFees maxsize has been reached\n",
+                __func__, __LINE__, entry->ToString().c_str());
```
AndreaLanfranchi: Here I think `.c_str()` is a bit pleonastic. Strings are already implicitly convertible to a null terminated char*

src/sc/sidechain.cpp
```diff
@@ -783,108 +782,91 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView, int blockHeigh
...
+
+            while (scFees.size() > max_size)
             {
-                const auto& entry = scFees_v2.front();
-                LogPrint("sc", "%s():%d - popping f=%d, m=%d, h=%d from list, as scFees maxsize has been reached\n",
-                    __func__, __LINE__, entry.forwardTxScFee, entry.mbtrTxScFee, entry.submissionHeight);
-                scFees_v2.pop_front();
+                const auto& entry = scFees.front();
+                LogPrint("sc", "%s():%d - popping %s from list, as scFees maxsize has been reached\n",
+                    __func__, __LINE__, entry->ToString().c_str());
```
AndreaLanfranchi: Here I think .c_str() is a bit pleonastic. Strings are already implicitly convertible to a null terminated char*

src/sc/sidechain.cpp
```diff
@@ -783,108 +782,91 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView, int blockHeigh
...
-            scFees_v2.remove_if([threshold](Sidechain::ScFeeData_v2 entry) {
-                const bool testCondition = (entry.submissionHeight <= threshold);
+            scFees.remove_if([threshold](std::shared_ptr<Sidechain::ScFeeData> entry) {
+                const std::shared_ptr<Sidechain::ScFeeData_v2> casted_entry = std::dynamic_pointer_cast<Sidechain::ScFeeData_v2>(entry);
+                const bool testCondition = (casted_entry->submissionHeight <= threshold);
                 if (testCondition)
-                    LogPrint("sc", "%s():%d - popping f=%d, m=%d, h=%d from list, as entry is too old (threshold height was %d)\n",
-                        __func__, __LINE__, entry.forwardTxScFee, entry.mbtrTxScFee, entry.submissionHeight, threshold);
+                    LogPrint("sc", "%s():%d - popping %s from list, as entry is too old (threshold height was %d)\n",
+                        __func__, __LINE__, casted_entry->ToString().c_str(), threshold);
```
AndreaLanfranchi: Here I think .c_str() is a bit pleonastic. Strings are already implicitly convertible to a null terminated char*

src/sc/sidechain.cpp
```diff
@@ -783,108 +782,91 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView, int blockHeigh
...
+
         }
         // On the other hand, if we found a scFeeData element for the current height, we just update it with lower
         // ft and/or mbtr fee rates. All the checks on the container size and element age have been performed
         // at the moment of the element insertion, hence their are not required here
         else {
-            scFeeIt->forwardTxScFee = std::min(scFeeIt->forwardTxScFee, ftScFee);
-            scFeeIt->mbtrTxScFee = std::min(scFeeIt->mbtrTxScFee, mbtrScFee);
+            scFeeIt->get()->forwardTxScFee = std::min(scFeeIt->get()->forwardTxScFee, ftScFee);
+            scFeeIt->get()->mbtrTxScFee    = std::min(scFeeIt->get()->mbtrTxScFee, mbtrScFee);
```
AndreaLanfranchi: I suggest for better readability 
```suggestion
            auto scFeeEntry = scFeeIt->get();
            scFeeEntry->forwardTxScFee = std::min(scFeeEntry->forwardTxScFee, ftScFee);
            scFeeEntry->mbtrTxScFee    = std::min(scFeeEntry->mbtrTxScFee, mbtrScFee);```

File: src/sc/sidechain.cpp [hunk 21]
```diff
@@ -783,108 +782,91 @@ void CSidechain::UpdateScFees(const CScCertificateView& certView, int blockHeigh
...
-            std::cout << "[" << std::setw(2) << entry.forwardTxScFee
-                      << "/" << std::setw(2) << entry.mbtrTxScFee
-                      << "/" << std::setw(2) << entry.submissionHeight << "]";
-        }
+
+    for (const auto& entry : scFees) {
+        std::cout << entry->ToString() << std::endl;
     }
-    std::cout << std::endl;
+
```
AndreaLanfranchi: This is actually very expensive as the stream is flushed at every iteration.
Use this instead
```
    for (const auto& entry : scFees) {
        std::cout << entry->ToString() << "\n";
    }
    std::cout << std::flush;
```
```

File: src/sc/sidechaintypes.h [hunk 22]
```diff
@@ -500,6 +507,15 @@ struct ScFeeData_v2 : public ScFeeData {
     {
         return ScFeeData::operator==(rhs) && submissionHeight == rhs.submissionHeight;
     }
+
+    virtual std::string ToString() const
```
AndreaLanfranchi: Shouldn't this be an `override` instead of `virtual` ?

src/serialize.h
AndreaLanfranchi: +1

src/serialize.h
```diff
@@ -1020,6 +1052,34 @@ void Unserialize(Stream& is, std::list<T, A>& l, int nType, int nVersion)
...
+    }
+}
+
+template<typename Stream, typename T, typename C, typename D>
+std::void_t<typename C::value_type> // enforce that C has member value_type, as it happens for STL collections
+Unserialize(Stream& is, C& l, int nType, int nVersion)
+{
+    l.clear();
+    unsigned int nSize = ReadCompactSize(is);
+    typename C::iterator it = l.begin();
```
AndreaLanfranchi: Is this needed ?

src/undo.h
a-petrini: Not now, as we have a different task for that

src/undo.h
a-petrini: Actually, we forgot a couple of curly braces to enclose the whole section in the context of the `if` true condition... Thanks for spotting this!

ptagl (changes requested):

src/sc/sidechain.cpp
AndreaLanfranchi: `scFees` is a list of shared ptrs and I don't see any enforcement of sorting.
In any case `find_if` returns on the very first occurrence and semantically is better than a normal loop

src/sc/sidechain.cpp
AndreaLanfranchi: There is a difference amongst remove_if() and your proposal: the first removes any item in the list matching the criteria while what you describe would remove the first encountered only. The question is ... is it possible to have multiple entries in the list matching the criteria ?

src/sc/sidechain.cpp
a-petrini: Unfortunately, for this syntax to work we have to dereference the iterator, obtaining a worse syntax:
`(*std::min_element( [...] ))->forwardTxScFee;`

src/sc/sidechain.cpp
a-petrini: [See prev. comment]

src/sc/sidechain.cpp
a-petrini: Done

src/sc/sidechain.cpp
a-petrini: Yes, I agree. I removed the while loop entirely and left only the check on threshold.

src/sc/sidechain.cpp
a-petrini: Done

src/sc/sidechain.cpp
a-petrini: Done

src/sc/sidechain.cpp
a-petrini: Done

src/sc/sidechain.cpp
a-petrini: Done

src/sc/sidechain.cpp
a-petrini: Done

src/sc/sidechain.cpp
a-petrini: I agree with Andrea

src/sc/sidechaintypes.h
drgora: Yes makes sense, thanks for spotting!

src/sc/sidechain.cpp
drgora: Given that this function is definitely not on performance critical path (it would not make sense to print to `cout` otherwise), and it's used only in gtests where it might be better to flush immediately for sake of debugging, I'd rather leave it as it is.

AndreaLanfranchi: This comment doesn't make any sense to me.

src/serialize.h
drgora: No it indeed was a leftover, nice catch!

src/sc/sidechain.cpp
drgora: As you also notice, if we are executing this, then it means it's a non ceasing sidechain, which means we *must* have `ScFeeData_v2`. If something goes wrong/changes in the future, I'd rather not being silent about unexpected conditions, but instead scream out loud immediately.

On the other hand, I agree we should do better than undefined behavior. I propose an `assert(casted_entry)`, is that ok with you guys?

src/serialize.h
drgora: Makes sense, I have renamed `b, d` to `base, derived`.

src/serialize.h
drgora: Makes sense, `T, C, D` -> `Base, Container, Derived`.

src/sc/sidechain.cpp
a-petrini: I agree with @drgora. `DumpScFees()` is used only in two tests, while any other dump of the `scFees` list is done  through `LogPrint()`. I don't think that performance is an issue here, hence I propose to keep the function as it is.

src/sc/sidechain.cpp
a-petrini: Yes, I think it's a good solution.
Returning `false` may not be enough if dynamic casting fails.

src/sc/sidechain.cpp
drgora: @AndreaLanfranchi sorry I was not really clear.
What I meant is that given that this function is not, and will never be in its current form, on any performance critical path, I'd rather keep it more compact. I agree it's not a big difference in size, as surely it won't be a big difference in performance anyway.
About debugging, I was suggesting that not delaying debugging logs might help understand the actual sequence of events, especially in multithreaded applications. Again, not a strong point, I agree, but also no strong points against the current form.

ptagl (approved):

PR #486, opened 2022-10-31
Ap/non ceasing sc quality revert

a-petrini:


src/coins.cpp
```diff
@@ -1185,12 +1185,13 @@ CValidationState::Code CCoinsViewCache::IsCertApplicableToState(const CScCertifi
...
-            __func__, __LINE__, certHash.ToString(), cert.quality);
-        return CValidationState::Code::INVALID;
-    }
+    // REMOVE: 
+    // if (sidechain.isNonCeasing() && cert.quality != 0)
+    // {
+    //     LogPrintf("%s():%d - ERROR: cert %s refers to a non-ceasing sidechain, but has non zero quality (%d)\n",
+    //         __func__, __LINE__, certHash.ToString(), cert.quality);
+    //     return CValidationState::Code::INVALID;
+    // }
```
AndreaLanfranchi: Please remove the commented out code. We can always get it back reading the history of changes of the file.

src/main.cpp
```diff
@@ -1207,7 +1207,7 @@ MempoolReturnValue AcceptCertificateToMemoryPool(CTxMemPool& pool, CValidationSt
 
     if(!CheckCertificate(cert, state))
     {
-        error("%s(): CheckCertificate failed", __func__);
+        error("%s(): CheckCertificate failed\n", __func__);
```
AndreaLanfranchi: Why use `error(` here and immediately before and after use `LogPrintf` ?

src/miner.cpp
```diff
@@ -127,13 +127,13 @@ bool VerifyCertificatesDependencies(const CScCertificate& cert)
         return false;
     }
 
-    if (mempool.mapSidechains.at(cert.GetScId()).mBackwardCertificates.count(std::make_pair(cert.quality, cert.epochNumber)) == 0)
+    if (mempool.mapSidechains.at(cert.GetScId()).mBackwardCertificates.count(std::make_pair(cert.epochNumber, cert.quality)) == 0)
```
AndreaLanfranchi: We do twice (here and immediately below) the creation of a pair and a lookup to `mapSideChains`- Wouldn't it be better to use some temp vars ?

src/txmempool.cpp
```diff
@@ -223,9 +223,9 @@ bool CTxMemPool::addUnchecked(const uint256& hash, const CCertificateMemPoolEntr
 
     if (mapSidechains.count(cert.GetScId())!= 0)
     {
-        assert(mapSidechains.at(cert.GetScId()).mBackwardCertificates.count(std::make_pair(cert.quality, cert.epochNumber)) == 0);
+        assert(mapSidechains.at(cert.GetScId()).mBackwardCertificates.count(std::make_pair(cert.epochNumber, cert.quality)) == 0);
     }
-    mapSidechains[cert.GetScId()].mBackwardCertificates[std::make_pair(cert.quality, cert.epochNumber)] = hash;
+    mapSidechains[cert.GetScId()].mBackwardCertificates[std::make_pair(cert.epochNumber, cert.quality)] = hash;
```
AndreaLanfranchi: This is overly complicated and has an unneded branch.
I suggest this form

```
    auto& sideChain = mapSidechains[cert.GetScId()]; // Creates new element if key does not exist
    const auto certKey = std::make_pair(cert.epochNumber, cert.quality);
    assert(sideChain.mBackwardCertificates.count(certKey) == 0);
    sideChain.mBackwardCertificates[certKey] = hash;

```

src/txmempool.cpp
```diff
@@ -1617,8 +1617,10 @@ bool CTxMemPool::checkReferencedHeight(const CScCertificate& incomingCert) const
 {
     // If quality !=0, then for sure isNonCeasing() sidechain is false, so we do not need
     // to perform any check.
-    if (incomingCert.quality != 0)
-        return true;
+    // ////
+    // if (incomingCert.quality != 0)
+    //     return true;
+    // ////
```
AndreaLanfranchi: Please remove this outdated comment and code

src/sc/sidechain.cpp
```diff
@@ -111,11 +111,11 @@ int CSidechain::getScCoinsMaturity()
 
 bool CSidechain::CheckQuality(const CScCertificate& cert) const
 {
-    // Quality disabled in non-ceasing SC, hence always return true if quality is 0
+    // Quality disabled in non-ceasing SC, hence only check for non negativity of quality
```
AndreaLanfranchi: Actually quality is NOT disabled in non-ceasing SC in fact we check for negativity.
Need to change the comment.
Also ... are we sure this is the only check for non-ceasing SC ?

qa/rpc-tests/sc_cert_getblocktemplate.py
```diff
@@ -183,8 +183,8 @@ def run_test_with_scversion(self, scversion, ceasable = True):
         ### 3: Sending a transaction and a certificate
         # Generate proof before the call of `getblocktemplate`.
         # It is a time consuming operation, so may take more than GET_BLOCK_TEMPLATE_DELAY seconds.
+        quality = quality + 1
```
JackPiri: Can be rewritten with `quality += 1` (coherently with line 279)

AndreaLanfranchi: Wouldn't  `++quality;` be even better ?

qa/rpc-tests/sc_cert_nonceasing.py
```diff
@@ -118,32 +118,48 @@ def run_test(self):
         mark_logs("Node0 confirms sc creation generating 2 blocks", self.nodes, DEBUG_MODE)
         self.nodes[0].generate(3)
 
+        # #------------------------------------------------
```
JackPiri: Are those commented lines (121 to 128) just old copy-pasted code that can be removed?

qa/rpc-tests/sc_cert_nonceasing.py
```diff
@@ -118,32 +118,48 @@ def run_test(self):
...
         #------------------------------------------------
-        mark_logs("## Test nok, trying to overwrite ##", self.nodes, DEBUG_MODE)
+        mark_logs("## Test nok, trying to overwrite with a cert of same quality##", self.nodes, DEBUG_MODE)
+        #------------------------------------------------
+        ref_height = self.nodes[0].getblockcount()-2
+        amount_cert_1 = {"address": addr_node2, "amount": bwt_amount_1}
+        self.try_send_certificate(0, scid, epoch_number, ref_quality, ref_height, MBTR_SC_FEE_HIGH, FT_SC_FEE, amount_cert_1, True, "invalid timing for certificate")
+
+        #------------------------------------------------
+        mark_logs("## Test nok, trying to overwrite with a cert with less quality ##", self.nodes, DEBUG_MODE)
```
drgora: I believe this section of the test might be a bit misleading, suggesting that a certificate could be overwritten by another with higher quality. If you want to keep this structure (even though the certificates are being rejected for ref_height and epoch reasons, not because of the quality), I'd suggest to show it explicitly by adding another section like `Test nok, trying to overwrite with a cert with higher quality`.

src/gtest/test_sidechain_certificate_quality.cpp
```diff
@@ -286,7 +286,7 @@ TEST_F(SidechainsMultipleCertsTestSuite, Cert_LowerQuality_DifferentEpoch_UndoDa
     EXPECT_TRUE(initialScState == revertedSidechain);
 }
 
-TEST_F(SidechainsMultipleCertsTestSuite, V2Cert_HigherQuality_SameEpoch_SidechainIsNotUpdated) {
+TEST_F(SidechainsMultipleCertsTestSuite, V2Cert_SameEpoch_SidechainIsNotUpdated) {
```
JackPiri: I would rename this test from `V2Cert_SameEpoch_SidechainIsNotUpdated` to `V2Cert_SameEpoch_SameQuality_SidechainIsNotUpdated` for better understanding and coherence with other tests names
In addition, the comment `//Insert high quality Certificate` should be corrected to `//Insert same quality Certificate` and `highQualityCert` should be corrected to `sameQualityCert`

src/gtest/test_sidechain_certificate_quality.cpp
```diff
@@ -323,7 +323,44 @@ TEST_F(SidechainsMultipleCertsTestSuite, V2Cert_HigherQuality_SameEpoch_Sidechai
...
+    CMutableScCertificate highQualityCert;
+    highQualityCert.scId        = scId;
+    highQualityCert.epochNumber = initialScState.lastTopQualityCertReferencedEpoch;
+    highQualityCert.quality     = 0;
+    highQualityCert.addBwt(CTxOut(CAmount(90), dummyScriptPubKey));
+    CAmount highQualityCert_TotalBwtAmount = CScCertificate(highQualityCert).GetValueOfBackwardTransfers();
+
+    //test
+    //sidechainsView->UpdateSidechain(highQualityCert, dummyUndo);
+    ASSERT_FALSE(sidechainsView->UpdateSidechain(highQualityCert, dummyUndo, sidechainsView->GetHeight()+1));
```
JackPiri: I guess `highQualityCert` should be corrected to `lowQualityCert`, am I correct?
Similarly, the comment at line 342 should be `//Insert low quality Certificate`

src/coins.cpp
```diff
@@ -1185,12 +1185,13 @@ CValidationState::Code CCoinsViewCache::IsCertApplicableToState(const CScCertifi
         return CValidationState::Code::SCID_NOT_FOUND;
     }
 
-    if (sidechain.isNonCeasing() && cert.quality != 0)
-    {
-        LogPrintf("%s():%d - ERROR: cert %s refers to a non-ceasing sidechain, but has non zero quality (%d)\n",
-            __func__, __LINE__, certHash.ToString(), cert.quality);
-        return CValidationState::Code::INVALID;
-    }
+    // REMOVE:
```
drgora: Just a reminder: please do remove this :)

src/gtest/test_sidechain_certificate_quality.cpp
```diff
@@ -323,7 +323,44 @@ TEST_F(SidechainsMultipleCertsTestSuite, V2Cert_HigherQuality_SameEpoch_Sidechai
...
+    //Insert high quality Certificate
+    CMutableScCertificate highQualityCert;
+    highQualityCert.scId        = scId;
+    highQualityCert.epochNumber = initialScState.lastTopQualityCertReferencedEpoch;
+    highQualityCert.quality     = 0;
+    highQualityCert.addBwt(CTxOut(CAmount(90), dummyScriptPubKey));
+    CAmount highQualityCert_TotalBwtAmount = CScCertificate(highQualityCert).GetValueOfBackwardTransfers();
+
+    //test
+    //sidechainsView->UpdateSidechain(highQualityCert, dummyUndo);
```
drgora: We should remove these commented out lines

src/gtest/test_sidechain_certificate_quality.cpp
```diff
@@ -358,7 +397,115 @@ TEST_F(SidechainsMultipleCertsTestSuite, V2Cert_LowerQuality_SameEpoch_Sidechain
     EXPECT_TRUE(sidechain.balance                     == initialScState.balance);
 }
 
-TEST_F(SidechainsMultipleCertsTestSuite, V2Cert_LowerQuality_DifferentEpoch_SidechainIsUpdated) {
+TEST_F(SidechainsMultipleCertsTestSuite, V2Cert_DifferentEpoch_SameQuality_SidechainIsUpdated) {
```
drgora: I believe that this section of the test (lines 400-542) might be more than we need. We know that, in general, certificates for different epochs have no relations in terms of quality with each other. It's something that we do not care about, so we do not want to assess anything specific.
I'm absolutely open to discussion on this point, but my suggestion would be to remove this part entirely, or at least leave only one test case (e.g. lower quality) and remove the others. In general, my concern is that we end up having a big burden with the maintenance of tests, some of which do not add real value for us.

src/main.cpp
```diff
@@ -1057,7 +1057,7 @@ bool CheckCertificatesOrdering(const std::vector<CScCertificate>& certList, CVal
                 return state.DoS(100, error("%s: incorrect certificate epoch order in block",
                     __func__), CValidationState::Code::INVALID, "bad-cert-epoch-ordering-in-block");
             }
-            if (mBestCertDataByScId.at(scid).second > cert.quality)
+            if ((mBestCertDataByScId.at(scid).first == cert.epochNumber) && (mBestCertDataByScId.at(scid).second > cert.quality))
```
drgora: It would be nice in terms of clarity and performance to limit the number of map accesses, possibly including also line 1053.

JackPiri (approved):
The PR looks good to me, I have just reported the need of some minor refactoring activity

src/coins.cpp
a-petrini: Removed in dfeb52fab30e3b941b44631be6cdf8a95f317a67

src/main.cpp
a-petrini: I am not sure the reason behind this choice. However, the `error(..)` function already appends the new line character, hence I reverted this correction in dfeb52fab30e3b941b44631be6cdf8a95f317a67. Thanks for spotting it.
We could postpone uniforming the` LogPrintf(..)` and the logging system in general in a separate refactoring task.

src/miner.cpp
a-petrini: Improved in dfeb52fab30e3b941b44631be6cdf8a95f317a67

src/txmempool.cpp
a-petrini: Updated in dfeb52fab30e3b941b44631be6cdf8a95f317a67

src/txmempool.cpp
a-petrini: Removed in dfeb52fab30e3b941b44631be6cdf8a95f317a67

src/sc/sidechain.cpp
a-petrini: Yes, the comment was outdated. Corrected in dfeb52fab30e3b941b44631be6cdf8a95f317a67

qa/rpc-tests/sc_cert_getblocktemplate.py
a-petrini: Agree with @JackPiri, updated in dfeb52fab30e3b941b44631be6cdf8a95f317a67.
Python does not have a unary increment operator, hence `++quality` and `quality += 1` are not equivalent.

qa/rpc-tests/sc_cert_nonceasing.py
a-petrini: It tested a condition that was supposed to fail in the past version and has been superseded by test in the lines immediately below them.
Removed in dfeb52fab30e3b941b44631be6cdf8a95f317a67

src/gtest/test_sidechain_certificate_quality.cpp
a-petrini: Updated in dfeb52fab30e3b941b44631be6cdf8a95f317a67.

src/gtest/test_sidechain_certificate_quality.cpp
a-petrini: Updated in dfeb52fab30e3b941b44631be6cdf8a95f317a67.

src/gtest/test_sidechain_certificate_quality.cpp
a-petrini: Yes, I agree with this. I would leave the zero quality submission test as it is ok in non-ceasing v2.

src/coins.cpp
a-petrini: Removed in dfeb52fab30e3b941b44631be6cdf8a95f317a67

src/gtest/test_sidechain_certificate_quality.cpp
a-petrini: Done in c7bb0cc7d304d7a51cc8ee874108797c8a3debe4.

src/gtest/test_sidechain_certificate_quality.cpp
a-petrini: c7bb0cc7d304d7a51cc8ee874108797c8a3debe4 removes all these occurrences from the same file.

qa/rpc-tests/sc_cert_nonceasing.py
a-petrini: c7bb0cc7d304d7a51cc8ee874108797c8a3debe4 updates the info string in `mark_logs()` and adds the suggested test with higher quality.

src/main.cpp
a-petrini: Optimized in c7bb0cc7d304d7a51cc8ee874108797c8a3debe4: three accesses have been replaced with a single one.

src/miner.cpp
AndreaLanfranchi: Still the `at()` lookup is enforced twice.
What about this ?
```
    const auto& side_chain = mempool.mapSidechains[cert.GetScId()];
    const auto epochQualityPair = std::make_pair(cert.epochNumber, cert.quality);
    if (side_chain.mBackwardCertificates.count(epochQualityPair) == 0)
    {
        if (fDebug) assert("cert is in mempool but not duly registered  in mapSidechains." == 0);
        return false;
    }

    if (side_chain.mBackwardCertificates.at(epochQualityPair) != cert.GetHash()) {
        if (fDebug) assert("a different cert with the same scId and quality is in mempool" == 0);
        return false;
    }
```

qa/rpc-tests/sc_cert_getblocktemplate.py
AndreaLanfranchi: Sorry my fault. Didn't notice it was a py file.

src/miner.cpp
a-petrini: Yes, way better, thanks! Fixed in cca710e3afd8f95b38056d2ad44acb08c01bfa8f

src/gtest/test_sidechain_certificate_quality.cpp
AndreaLanfranchi: Structured binding definitely improves readability
I suggest to enforce it by
```
const auto [bestCertEpoch, bestCertQuality] = mBestCertDataByScId[scid];
```
and then refer to variables rather than `first` and `second`

a-petrini: I was not aware of the structured binding feature introduced in C++17, thanks; I will definitely looking into that.
However, in this context this looks like a task more related to refactoring, than to quality revert. For this reason, I prefer to postpone the enforcing of this practice and leave the PR as it is now.

src/gtest/test_sidechain_certificate_quality.cpp
```diff
@@ -771,7 +839,7 @@ TEST(SidechainsMultipleCerts, BlocksWithCertsOfSameOrIncreasingEpochsAndDecreasi
     cert_3.epochNumber = cert_1.epochNumber + 1;
     bBlock.vcert = {cert_1, cert_2, cert_3};
 
-    EXPECT_FALSE(CheckCertificatesOrdering(bBlock.vcert, dummyState));
+    EXPECT_TRUE(CheckCertificatesOrdering(bBlock.vcert, dummyState));   // This is now accepted for v2 sc
```
ptagl: Can we get rid of this comment?
```suggestion
    EXPECT_TRUE(CheckCertificatesOrdering(bBlock.vcert, dummyState));
```

src/gtest/test_sidechain_certificate_quality.cpp
```diff
@@ -771,7 +839,7 @@ TEST(SidechainsMultipleCerts, BlocksWithCertsOfSameOrIncreasingEpochsAndDecreasi
     cert_3.epochNumber = cert_1.epochNumber + 1;
```
ptagl: What about using static values for quality and epoch initialization for increased readability?

src/sc/sidechain.cpp
```diff
@@ -111,11 +111,11 @@ int CSidechain::getScCoinsMaturity()
...
+    // Check for non negativity of quality
     if (isNonCeasing())
     {
-        if (cert.quality != 0) {
-            LogPrint("cert", "%s.%s():%d - NOK, cert %s for non-ceasing sidechain, but quality is non zero (q=%d).\n",
+        if (cert.quality < 0) {
+            LogPrint("cert", "%s.%s():%d - NOK, cert %s for non-ceasing sidechain has negative quality (q=%d).\n",
                 __FILE__, __func__, __LINE__, cert.GetHash().ToString(), cert.quality);
             return false;
         }
```
ptagl: Given the new behavior of non-ceasable sidechains (quality != 0 is allowed) this whole check could be removed.

ptagl (changes requested):

src/gtest/test_sidechain_certificate_quality.cpp
a-petrini: Yes we can.

src/gtest/test_sidechain_certificate_quality.cpp
a-petrini: This is a good idea, as in its current form it's hard to follow.

src/sc/sidechain.cpp
a-petrini: Agreed, I removed this check.

ptagl (approved):

PR #494, opened 2022-11-29
Consider spent tx in p2p

drgora:
This patch makes sure that zend does not ask for tx data to its peers for tx which have already been received and spent.

### DIFF
--- a/src/limitedmap.h
+++ b/src/limitedmap.h
@@ -46,20 +46,22 @@ public:
         }
         return;
     }
-    void erase(const key_type& k)
+    size_t erase(const key_type& k)
     {
         iterator itTarget = map.find(k);
         if (itTarget == map.end())
-            return;
+            return 0;
         std::pair<rmap_iterator, rmap_iterator> itPair = rmap.equal_range(itTarget->second);
-        for (rmap_iterator it = itPair.first; it != itPair.second; ++it)
+        for (rmap_iterator it = itPair.first; it != itPair.second; ++it) {
             if (it->second == itTarget) {
                 rmap.erase(it);
                 map.erase(itTarget);
-                return;
+                return 1; // be consistent with std::map return value
             }
+        }
         // Shouldn't ever get here
         assert(0);
+        return 0;
     }
     void update(const_iterator itIn, const mapped_type& v)
     {
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -6792,6 +6792,7 @@ void ProcessTxBaseMsg(const CTransactionBase& txBase, CNode* pfrom)
 
     pfrom->setAskFor.erase(inv.hash);
     mapAlreadyAskedFor.erase(inv);
+    mapAlreadyReceived.insert(std::make_pair(inv, GetTimeMicros()));
 
     MempoolReturnValue res = MempoolReturnValue::INVALID;
     CValidationState state;
@@ -8150,7 +8151,7 @@ bool SendMessages(CNode* pto, bool fSendTrickle)
         while (!pto->fDisconnect && !pto->mapAskFor.empty() && (*pto->mapAskFor.begin()).first <= nNow)
         {
             const CInv& inv = (*pto->mapAskFor.begin()).second;
-            if (!AlreadyHave(inv))
+            if (!AlreadyHave(inv) && mapAlreadyReceived.find(inv) == mapAlreadyReceived.end())
             {
                 if (fDebug)
                     LogPrint("net", "%s():%d - Requesting %s peer=%d\n", __func__, __LINE__, inv.ToString(), pto->id);
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -99,6 +99,7 @@ map<CInv, CDataStream> mapRelay;
 deque<pair<int64_t, CInv> > vRelayExpiration;
 CCriticalSection cs_mapRelay;
 limitedmap<CInv, int64_t> mapAlreadyAskedFor(MAX_INV_SZ);
+limitedmap<CInv, int64_t> mapAlreadyReceived(MAPRECEIVED_MAX_SZ);
 
 static deque<string> vOneShots;
 CCriticalSection cs_vOneShots;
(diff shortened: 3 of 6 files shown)


src/limitedmap.h
```diff
@@ -46,20 +46,22 @@ class limitedmap
...
             if (it->second == itTarget) {
                 rmap.erase(it);
                 map.erase(itTarget);
-                return;
+                return 1; // be consistent with std::map return value
             }
+        }
         // Shouldn't ever get here
         assert(0);
+        return 0;
```
AndreaLanfranchi, line 64: What's the point of this inserted `return 0` if, as the comment says plus the very strong assertion, we should never get here ?

src/limitedmap.h
```diff
@@ -46,20 +46,22 @@ class limitedmap
...
-    void erase(const key_type& k)
+    size_t erase(const key_type& k)
     {
         iterator itTarget = map.find(k);
         if (itTarget == map.end())
-            return;
+            return 0;
         std::pair<rmap_iterator, rmap_iterator> itPair = rmap.equal_range(itTarget->second);
-        for (rmap_iterator it = itPair.first; it != itPair.second; ++it)
+        for (rmap_iterator it = itPair.first; it != itPair.second; ++it) {
```
AndreaLanfranchi, line 55: There's an extra unneeded copy of the iterator in the for loop initialization.
I'd rather change to

```
        auto [rmap_it, rmap_end] = rmap.equal_range(itTarget->second);
        for (; rmap_it != rmap_end; ++rmap_it) {
            if(rmap_it->second == itTarget) {
                rmap.erase(rmap_it);
                map.erase(itTarget);
                return 1;
            }
        }

```

AndreaLanfranchi (changes requested):

src/limitedmap.h
```diff
@@ -46,20 +46,22 @@ class limitedmap
         }
         return;
```
AndreaLanfranchi, line 47: This return is pleonastic

AndreaLanfranchi (dismissed):

AndreaLanfranchi:
Also according to `limitedmap` description (`STL-like map container that only keeps the N elements with the highest value`) the method `insert` is wrongly implemented cause it may bump off the list higher values.
Picture this:

map contains : [{k1, 2}, {k2, 3}]
nMaxSize == 2

if we try to insert {k3, 1} we end up bumping off k1 and as a result the map contains : [{k3, 1}, {k2, 3}] **which is NOT the list of higher values**

IMHO we should check for to-be-inserted-value to be higher than lowest value **before** insertion.

src/limitedmap.h
drgora: Yes I agree, let's remove it!

src/limitedmap.h
drgora: You're right, I didn't notice, I will fix this... actually, I didn't spend any time refactoring studying/refactoring this class besides the minimal change that I applied, but maybe it's a good chance to also remove the TODO in the code

AndreaLanfranchi: I already examined the TODO and basically I'd remove it. Erase map by iterator removes items in range [begin ... end) (end excluded) hence the comment is misleading (we'd have to advance the iterator).

src/limitedmap.h
drgora: Without this, if we disable assertions, then the compiler would generate warnings for no return value. I'm not sure, but some compilers might even generate warning with assertions enabled?

AndreaLanfranchi: We need to set a standard : as of now I've been told we cannot disable assertions without breaking a whole lot more on the code so I would not see this as an issue.
On the other hand if you mean that we might have a warning for missing return on all code patterns then you might be right but this require either handling exceptions or return codes.

a-petrini (approved):

src/main.cpp
```diff
@@ -6792,6 +6792,7 @@ void ProcessTxBaseMsg(const CTransactionBase& txBase, CNode* pfrom)
 
     pfrom->setAskFor.erase(inv.hash);
     mapAlreadyAskedFor.erase(inv);
+    mapAlreadyReceived.insert(std::make_pair(inv, GetTimeMicros()));
```
JackPiri, line 6795: Is it ok to have just an "insert" here instead of an "insert-or-update"?

src/net.cpp
```diff
@@ -99,6 +99,7 @@ map<CInv, CDataStream> mapRelay;
 deque<pair<int64_t, CInv> > vRelayExpiration;
 CCriticalSection cs_mapRelay;
 limitedmap<CInv, int64_t> mapAlreadyAskedFor(MAX_INV_SZ);
+limitedmap<CInv, int64_t> mapAlreadyReceived(MAPRECEIVED_MAX_SZ);
```
JackPiri, line 102: Is it ok those two `limitedmap`s having different size?

JackPiri (approved):
I have left a couple of comments related to `mapAlreadyReceived`; it should be nothing of critical impact, feel free to ignore.
Hence the PR is fine to me.

src/limitedmap.h
ptagl: Just for reference, there is an updated comment from Bitcoin (even though they removed the file years ago):

https://github.com/bitcoin/bitcoin/blob/cc16fff3e476a9378d2176b3c1b83ad12b1b052a/src/limitedmap.h#L69

src/net.cpp
drgora: It should be ok, the two maps have different purposes and are independent of each other, so it is reasonable they have different sizes.

src/main.cpp
drgora: This is a good question. In principle, once you have and `inv` in the map you should not execute `ProcessTxBaseMsg` anymore. Still, this could happen if a peer's `tx` response is delayed by a lot, so that you first send out multiple requests for the same `inv`, and eventually receive responses for all the requests.

Still, the `mapAlreadyReceived`'s size is thought to accommodate a constant flow of 1000tx/s for 16min, i.e. the maximum expected delay for the execution of an inv request (considering the default parameters of 8 peers).
So, I'd rather not update entries after the initial insertion, if additional requests are actually performed. The update would keep the entry around longer than needed, and use a slot which might be better used by a more recent `inv` request.

src/limitedmap.h
drgora: @AndreaLanfranchi @ptagl  Thank you guys for your comments. All in all, considering also the other comments, I agree with you that currently `src/limitedmap.h` has some problems and could use some refactoring and bugfixing.

I have some pending work already, and I will open another PR shortly dedicated to this, but in the meantime I would rather not pollute this PR and shift its original focus.

Is it ok with you if we go ahead with this merge as it is for now?

ptagl (approved):
LGTM, I like the idea to keep the PR as it is, let's merge and then create a new branch for the refactoring of the limited map.

src/limitedmap.h
AndreaLanfranchi: I'm not on the same page but ok.

src/limitedmap.h
AndreaLanfranchi: Not resolved but "hopefully" will be done i some future

src/limitedmap.h
AndreaLanfranchi: Not removed but "hopefully" will be done in some future

PR #495, opened 2022-11-30
Added new coins selection algorithms

JackPiri:
This pull request brings in improvements in the input (coins/notes) selection phase when wallet needs to create transactions/certificates, in particular:
- new coins/notes selection criteria, aiming to maximize size of selection set*; at the same time change is kept at zero or at low level, through the use of increasing change bands
- validation of transaction size constraint (previously absent); this allows to exclude inputs sets too big that would result in a discarded transaction/certificate (due to oversized limit); furthermore, for z-transactions, in case of impossibility to generate the transaction, an error is immediately signalled during the inputs selection phase instead of waiting for several joinsplits computation (as was previously done)
- improved automatic fee selection, granting that the minimum fee is paid (previously the operation was looped with iterative estimation)

For reviewing this PR, the author suggests to check the changes with the following order:
1. transaction.h/wallet.h/wallet.cpp (focusing on transparent part):
- the new approach of estimating the non-input parts of tx/cert is shown, followed by the actual selection of inputs;
- the fundamentals structure for tx/cert estimation are introduced and major modifications targets methods SelectCoins and SelectCoinsMinConf, new method SelectNotes is provided (but it is better to check later)
2. coinsselectionalgorithm.h/coinsselectionalgorithm.cpp.cpp:
- the new algorithms for coins selection are provided: fast and not optimal (SlidingWindow), slow and optimal (B&B); the new algorithm for notes selection is provided: ForNotes (but it is better to check later)
- check the integration of those new algorithms within SelectCoinsMinConf method
3. sidechainrpc.h/sidechainrpc.cpp:
- check the integration of the new algorithms within addInputs method
4. asyncrpcoperation_sendmany.h/asyncrpcoperation_sendmany.cpp/wallet.h/wallet.cpp/coinsselectionalgorithm.h/coinsselectionalgorithm.cpp.cpp:
- major modifications targets main_impl method
- check the integration of new method SelectNotes and new algorithm ForNotes within main_impl method
5. oversized_tx_cert:
- the creation of oversized tx/cert is tested; the majority of positive tests were failing with previous implementation (but it was a bad failure, because in many cases a solution could be found)
6. all the rest...(very minor sparse refactoring are included in this PR)

*this aspect is crucial in terms of community goal, since the growth of the global set of UTXOs is a critical issue for any UTXO-based model (in contrast with account-based model); hence the reduction of this set should be always encouraged and welcome.


src/coinsselectionalgorithm.hpp
```diff
@@ -0,0 +1,280 @@
...
+//! Abstract class for algorithm of coins selection
+/*!
+  This class provides common fields required by each implementation and utility methods
+*/
+class CCoinsSelectionAlgorithm
+{    
+protected:
+    // auxiliary
+    //! The temporary set of selected elements (true->selected, false->unselected)
+    bool* tempSelection;
```
AndreaLanfranchi: Why this old style (and unsafe) raw pointers to arrays ?
What about a `vector<bool>` or, if you just want an array, a `unique_ptr<bool[]>` ?

src/coinsselectionalgorithm.hpp
```diff
@@ -0,0 +1,280 @@
...
+    const int maxIndex;
+    // auxiliary
+
+    // profiling and control
+    //! Flag identifying if the solving routine is running
+    bool isRunning;
+    //! Flag identifying if a stop of the solving routine has been requested
+    bool stopRequested;
+    //! The thread associated to the solving routine
+    std::thread* solvingThread;
```
AndreaLanfranchi: We have RAII why not use `std::unique_ptr<std::thread>` ?

src/coinsselectionalgorithm.hpp
```diff
@@ -0,0 +1,280 @@
...
+    //! The algorithm type
+    CoinsSelectionAlgorithmType type;
+
+    // input variables
+    //! Number of elements
+    const int problemDimension;
+    //! The array of amounts
+    const CAmount* amounts;
+    //! The array of sizes (in terms of bytes of the associated input)
+    const size_t* sizes;
```
AndreaLanfranchi: See same comment about tmpSelection

src/coinsselectionalgorithm.hpp
```diff
@@ -0,0 +1,280 @@
...
+    const CAmount targetAmount;
+    //! The target amount plus a positive offset (it is an upper-limit constraint)
+    const CAmount targetAmountPlusOffset;
+    //! The available total size (in terms of bytes, it is an upper-limit constraint)
+    const size_t availableTotalSize;
+    // input variables
+
+    // output variables
+    //! The optimal set of selected elements (true->selected, false->unselected)
+    bool* optimalSelection;
```
AndreaLanfranchi: See same comment about tmpSelection

src/coinsselectionalgorithm.cpp
```diff
@@ -0,0 +1,361 @@
...
+    StopSolving();
+    delete[] amounts;
+    delete[] sizes;
+    delete[] tempSelection;
+    delete[] optimalSelection;
+}
+
+CAmount* CCoinsSelectionAlgorithm::PrepareAmounts(std::vector<std::pair<CAmount, size_t>> unsortedAmountsAndSizes)
+{
+    std::vector<std::pair<CAmount, size_t>> sortedAmountsAndSizes = unsortedAmountsAndSizes;
```
AndreaLanfranchi: You already pass `unsortedAmountsAndSizes` by copy then why create a third copy into `sortedAmountsAndSizes` ?

src/coinsselectionalgorithm.cpp
```diff
@@ -0,0 +1,361 @@
...
+    std::sort(sortedAmountsAndSizes.begin(), sortedAmountsAndSizes.end(), [](std::pair<CAmount, size_t> left, std::pair<CAmount, size_t> right) -> bool { return ( left.first > right.first); } );
```
AndreaLanfranchi: If you want to sort in reverse order you simply do `std::sort(sortedAmountsAndSizes.rbegin(), sortedAmountsAndSizes.rend())`
See https://godbolt.org/z/Mv5M4xf6z

src/coinsselectionalgorithm.cpp
```diff
@@ -0,0 +1,361 @@
...
+size_t* CCoinsSelectionAlgorithm::PrepareSizes(std::vector<std::pair<CAmount, size_t>> unsortedAmountsAndSizes)
+{
+    std::vector<std::pair<CAmount, size_t>> sortedAmountsAndSizes = unsortedAmountsAndSizes;
+    std::sort(sortedAmountsAndSizes.begin(), sortedAmountsAndSizes.end(), [](std::pair<CAmount, size_t> left, std::pair<CAmount, size_t> right) -> bool { return ( left.first > right.first); } );
+    size_t* sortedSizes = new size_t[problemDimension];
+    for (int index = 0; index < problemDimension; ++index)
+    {
+        sortedSizes[index] = sortedAmountsAndSizes[index].second;
+    }
+    return sortedSizes;
```
AndreaLanfranchi: As here we do the very same sorting and the very same traversing across the very same collection of items ... isn't it better to merge the two functions into one ?

src/coinsselectionalgorithm.cpp
```diff
@@ -0,0 +1,361 @@
...
+        optimalSelection[index] = false;
+    }
+
+    optimalTotalAmount = 0;
+    optimalTotalSize = 0;
+    optimalTotalSelection= 0;
+
+    isRunning = false;
+    stopRequested = false;
+    solvingThread = nullptr;
```
AndreaLanfranchi: This is quite unsafe. What if the `reset()` method is called while the solving thread is still working ?

src/coinsselectionalgorithm.cpp
```diff
@@ -0,0 +1,361 @@
...
+        CAmount tempTotalAmount = 0;
+        unsigned int tempTotalSelection = 0;
+        int exclusionIndex = maxIndex;
+        int inclusionIndex = maxIndex;
+        for (; inclusionIndex >= 0; --inclusionIndex)   
+        {
+            if (stopRequested)
+            {
+                break;
+            }
```
AndreaLanfranchi: Why not simply `for (; !stopRequested && inclusionIndex >= 0; --inclusionIndex)` ?

src/coinsselectionalgorithm.hpp
```diff
@@ -0,0 +1,280 @@
...
+    BRANCH_AND_BOUND = 2
+};
+
+/* ---------- CCoinsSelectionAlgorithm ---------- */
+
+//! Abstract class for algorithm of coins selection
+/*!
+  This class provides common fields required by each implementation and utility methods
+*/
+class CCoinsSelectionAlgorithm
```
AndreaLanfranchi: I believe this class, and the ones inheriting from this, should not be copyable nor moveable

src/coinsselectionalgorithm.hpp
```diff
@@ -0,0 +1,280 @@
...
+    bool* tempSelection;
+    //! Max index of elements (equal to "problemDimension - 1")
+    const int maxIndex;
+    // auxiliary
+
+    // profiling and control
+    //! Flag identifying if the solving routine is running
+    bool isRunning;
+    //! Flag identifying if a stop of the solving routine has been requested
+    bool stopRequested;
```
AndreaLanfranchi: This var is also used into solving thread to check whether it should stop: should be qualified as atomic to sync memory amongst threads

src/coinsselectionalgorithm.hpp
```diff
@@ -0,0 +1,280 @@
...
+    // auxiliary
+    //! The temporary set of selected elements (true->selected, false->unselected)
+    bool* tempSelection;
+    //! Max index of elements (equal to "problemDimension - 1")
+    const int maxIndex;
+    // auxiliary
+
+    // profiling and control
+    //! Flag identifying if the solving routine is running
+    bool isRunning;
```
AndreaLanfranchi: This var is also used into solving thread to check whether it should stop: should be qualified as atomic to sync memory amongst threads

src/coinsselectionalgorithm.cpp
```diff
@@ -0,0 +1,361 @@
...
+    {
+        if (solvingThread != nullptr)
+        {
+            // this can be the case of a new start on the same algorithm instance,
+            // hence a deletion of previous pointer is performed before reallocation
+            delete solvingThread;
+        }
+        solvingThread = new std::thread(&CCoinsSelectionAlgorithm::Solve, this);
+    }
+}
```
AndreaLanfranchi: Being `isRunning` not atomic and checked with `compareExchangeStrong` we could invoke this multiple times with the effect of deleting the pointer to active thread(s)

AndreaLanfranchi (changes requested):

src/coinsselectionalgorithm.cpp
```diff
@@ -0,0 +1,361 @@
...
+                                                     targetAmountPlusOffset {_targetAmountPlusOffset},
+                                                     availableTotalSize {_availableTotalSize}
+{
+    tempSelection = new bool[problemDimension];    
+    optimalSelection = new bool[problemDimension];
+    for (int index = 0; index < problemDimension; ++index)
+    {
+        tempSelection[index] = false;
+        optimalSelection[index] = false;
+    }
```
AndreaLanfranchi: Since C++11 this loop for initializing the values of arrays is totally unnecessary.
Simply use:
```
    tempSelection = new bool[problemDimension]{};    
    optimalSelection = new bool[problemDimension]{};
```

src/coinsselectionalgorithm.cpp
```diff
@@ -0,0 +1,361 @@
...
+}
+
+CCoinsSelectionAlgorithm::~CCoinsSelectionAlgorithm()
+{
+    // solvingThread stopping must be performed before data destruction
+    StopSolving();
+    delete[] amounts;
+    delete[] sizes;
+    delete[] tempSelection;
+    delete[] optimalSelection;
```
AndreaLanfranchi: With vectors and/or smart pointers we can get rid of this

src/coinsselectionalgorithm.hpp
```diff
@@ -0,0 +1,280 @@
...
+     the temporary optimal solution (ties are handled prioritizing low total amount) then all the branches from the current
+     recursion on are cut; this is done in order to avoid reaching leaves that would certainly not improve the temporary optimal
+     solution.
+*/
+class CCoinsSelectionBranchAndBound : public CCoinsSelectionAlgorithm
+{
+protected:
+    // auxiliary
+    //! The array of cumulative amounts (considered summing amounts from index to end of amounts array)
+    const CAmount* cumulativeAmountsForward;
```
AndreaLanfranchi: Again use STL container or `std::unique_ptr<CAmount[]>` so we don't have to remember to do what RAII already does

src/coinsselectionalgorithm.cpp
```diff
@@ -0,0 +1,361 @@
...
+                if (tempTotalAmountNewBiggestPossible >= targetAmount) // {backtracking}
+                {
+                    unsigned int tempTotalSelectionNew = tempTotalSelection + (value ? 1 : 0);
+                    unsigned int maxTotalSelectionForward = tempTotalSelectionNew + (maxIndex - currentIndex);
+                    if ((maxTotalSelectionForward > optimalTotalSelection) ||
+                        (maxTotalSelectionForward == optimalTotalSelection && tempTotalAmountNewBiggestPossible < optimalTotalAmount)) // {bounding}
+                    {
+                        if (currentIndex < maxIndex)
+                        {
+                            SolveRecursive(nextIndex, tempTotalSizeNew, tempTotalAmountNew, tempTotalSelectionNew);
```
AndreaLanfranchi, line 384: Given the number of items in the provided vector of amounts and sizes I would warmly suggest to revrite this function iteratively instead of recursively to avoid the risk to face stack-overflow limits (which are configurable in OS even at small values).

src/coinsselectionalgorithm.cpp
```diff
@@ -0,0 +1,361 @@
...
+                }
+                break;
+            }
+        }
+        #if COINS_SELECTION_ALGORITHM_PROFILING
+        iterations = (maxIndex + 1 - inclusionIndex) + (maxIndex - exclusionIndex);
+        executionMicroseconds = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::system_clock::now().time_since_epoch()).count() - microsecondsBefore;
+        #endif
+        if (stopRequested == false)
+            hasCompleted = true;
```
AndreaLanfranchi: This is wrong: it changes hasCompleted to true only if not stopRequested.
So this may happen : run one cycle without requesting stop (which will set to hasCompleted to true) then invoke another cycle and this time request stop : as a result hasCompleted is still true.
As a safety measure you should :
- make hasCompleted atomic (is written by one thread and read in another)
- reset hasCompleted to false at the same point where you set isRunning = true
- change this to `hasCompleted = !stopRequested`

AndreaLanfranchi (changes requested):

src/sc/sidechainrpc.cpp
```diff
@@ -13,6 +13,8 @@
 #include <main.h>
 #include <init.h>
 
+#define MAX_LOOP 100
```
AndreaLanfranchi: Please don't use `#define` (MACRO style) to define constants as this pollutes the preprocessor space and basically becomes a literal copied everywhere.
Use `static const` or better `constexpr`

AndreaLanfranchi (changes requested):

src/coinsselectionalgorithm.hpp
JackPiri: Your point makes totally sense to me, but there is a but...
Actually during the initial prototypal phase, all those pointers were `std::vector`, then I moved to performance tuning and I got a significant improvement moving to raw pointers: I am talking about 10%/15% less execution time (which, honestly, was a surprise to me).
Given that the selection algorithm are very performance critical I would keep the raw pointers implementation; for safety I'd ask to the reviewers a double check on this piece of code (@ptagl, @drgora, @a-petrini).

src/coinsselectionalgorithm.hpp
JackPiri: An analogous consideration about `tmpSelection`

src/coinsselectionalgorithm.hpp
JackPiri: An analogous consideration about `tmpSelection`

src/coinsselectionalgorithm.cpp
JackPiri: Agree, copying `unsortedAmountsAndSizes` to `sortedAmountsAndSizes` is resource waist.
Modified.

src/coinsselectionalgorithm.cpp
JackPiri: I agree with your suggestion.
Modified accordingly.

src/sc/sidechainrpc.cpp
JackPiri: I agree with your suggestion.
Modified accordingly.

src/coinsselectionalgorithm.hpp
AndreaLanfranchi: Two considerations here :

- "**Premature optimization is the root of all evils**" (D. Knuth) : first we write safe and maintainable code then we work on it for optimization (if necessary)
- You can use `unique_ptr<boo[]>` at the very least

src/coinsselectionalgorithm.cpp
JackPiri: I agree with your suggestion.
Modified accordingly.

src/coinsselectionalgorithm.hpp
JackPiri: I agree with your suggestion.
Modified accordingly.

src/coinsselectionalgorithm.cpp
JackPiri: I agree with your suggestion.
Modified accordingly.

src/coinsselectionalgorithm.hpp
JackPiri: I agree with your suggestion.
Modified accordingly.

src/coinsselectionalgorithm.hpp
JackPiri: I agree with your suggestion.
Modified accordingly.

src/coinsselectionalgorithm.hpp
JackPiri: I agree with your suggestion.
Modified accordingly.

src/coinsselectionalgorithm.cpp
JackPiri: I agree with your suggestion.
Modified accordingly.

src/coinsselectionalgorithm.cpp
JackPiri: I agree with your suggestion.
Modified so that sorting is done just once..

src/coinsselectionalgorithm.cpp
JackPiri: Member `isRunning` qualified as atomic, moved `solvingThread` from raw pointer to smart pointer.

src/coinsselectionalgorithm.cpp
JackPiri: I was able to think of this iterative code:

```
int selectionSize = 10;
std::vector<int> selection(selectionSize, 0);
int maxIteration = (int)pow(2, selectionSize);
for (int iter = 0; iter < maxIteration; iter++)
    for (int p = 0; p < 32; p++)
        selection[p] = (iter >> p) & 1;
```

but honestly I cannot figure out how to leverage the backtracking and bounding techniques with this iterative way, it would end up listing all the possible combinations with a brute force strategy, something that is not feasible with production data.
By now, what I can do is handle any stackoverflow error with try-catch and possibly reduce the stack frame for increasing recursion depth/capacity (but even right now, considering a stack size of 1 MB, a selection of few more than 16K coins should work fine).

src/coinsselectionalgorithm.cpp
JackPiri: I agree with your suggestion.
Modified accordingly (with additional review of reset method).

src/coinsselectionalgorithm.cpp
JackPiri: Modified `Reset `method for achieveing better synchronization with `solvingThread`.

src/coinsselectionalgorithm.hpp
JackPiri: I have pushed on remote a new temporary branch named *cso_profiling_test*, three commits are available:

- AS IS: current (as of 225568e0caa980a2ddce83c4844cc98ba154d655) implementation
- WITH VECTORS: all raw pointers data replaced by std:vector 
- WITH SMART POINTERS: all raw pointers data replaced by std:_unique_ptr

it's a very quick-and-dirt branch which allow an immediate test of algorithm selection routine at the very start of Zend node (actually aborting the start-up procedure), printing a rough report file with timing performance.

Running on my machine, I confirm the performance degradation using both vectors or smart pointers in place of raw pointers. I know @drgora is playing with that branch and is reporting a degradation with vectors but not with smart pointers; let's wait for a feedback also by the other team members and then we will wrap up all the ideas and take a decision about which implementation to choose.

src/coinsselectionalgorithm.cpp
JackPiri: An analogous consideration about `tmpSelection`

src/coinsselectionalgorithm.hpp
JackPiri: An analogous consideration about `tmpSelection`

src/coinsselectionalgorithm.cpp
JackPiri: Not sure actually if a stackoverflow error can be handled with try-catch; maybe preventing the algorithm running on data sets too big (# > 16K) could be the best approach...

src/wallet/wallet.cpp
```diff
@@ -2946,73 +2946,31 @@ void CWallet::AvailableCoins(vector<COutput>& vCoins, bool fOnlyConfirmed, const
...
+    selectionTotalBytes = 0;
 
     // List of values less than target
-    pair<CAmount, pair<const CWalletTransactionBase*,unsigned int> > coinLowestLarger;
-    coinLowestLarger.first = std::numeric_limits<CAmount>::max();
-    coinLowestLarger.second.first = NULL;
-    vector<pair<CAmount, pair<const CWalletTransactionBase*,unsigned int> > > vValue;
-    CAmount nTotalLower = 0;
+    // <gross value, value, output, size>
+    vector<tuple<CAmount, CAmount, COutput, size_t>> vValue;
```
AndreaLanfranchi: Either use explicitly `std::` namespace or don't but be consistent (see below std::numeric_limits)
Preferrable to use it explicitly

src/wallet/wallet.cpp
```diff
@@ -3024,82 +2982,161 @@ bool CWallet::SelectCoinsMinConf(const CAmount& nTargetValue, int nConfMine, int
...
-                setCoinsRet.insert(vValue[i].second);
-                nValueRet += vValue[i].first;
+                break;
             }
+            std::this_thread::sleep_for(std::chrono::milliseconds(100));
+        }
+        if (!slowOptimalAlgorithm->GetHasCompleted())
+        {
+            slowOptimalAlgorithm->StopSolving();
+        }
```
AndreaLanfranchi: The whole point of using async (better say in this case threads) is to carry out parallel work. There is no point in starting a thread eventually forcing parent thread to sleep and wait for completion. Do the work in current thread with a time out

src/wallet/wallet.cpp
```diff
@@ -3024,82 +2982,161 @@ bool CWallet::SelectCoinsMinConf(const CAmount& nTargetValue, int nConfMine, int
...
+        CCoinsSelectionAlgorithmBase::GetBestAlgorithmBySolution(slowOptimalAlgorithm, fastNotOptimalAlgorithm, bestAlgorithm);
```
AndreaLanfranchi, line 3035: In case `slowOptimalAlgorithm->GetHasCompleted()` returns false ... what is the reliability of its result (`optimalTotalSelection`) is constantly updated in the SolveRecursive routine ?
Are we comparing the result from fastNotOptimalAlgorithm with an **incomplete** slowOptimalAlgorithm ?

AndreaLanfranchi (changes requested):

src/coinsselectionalgorithm.cpp
AndreaLanfranchi: You can change the recursion to iteration

src/coinsselectionalgorithm.hpp
AndreaLanfranchi: Let's consider also the two algos, in this impl, do not run in parallel hence the overall performance is affected.

src/wallet/wallet.cpp
```diff
@@ -3024,82 +2982,161 @@ bool CWallet::SelectCoinsMinConf(const CAmount& nTargetValue, int nConfMine, int
...
+    std::unique_ptr<CCoinsSelectionAlgorithmBase> bestAlgorithm = nullptr;
+
+    std::unique_ptr<CCoinsSelectionAlgorithmBase> fastNotOptimalAlgorithm;
+    for (int i = 0; i < COINS_SELECTION_INTERMEDIATE_CHANGE_LEVELS + 2; ++i)
     {
-        setCoinsRet.insert(coinLowestLarger.second);
-        nValueRet += coinLowestLarger.first;
+        const CAmount targetAmountPlusOffset = targetAmountPlusOffsetNoChange +
+                                               (double)(i) / (COINS_SELECTION_INTERMEDIATE_CHANGE_LEVELS + 1) * (targetAmountPlusOffsetMaxChange - targetAmountPlusOffsetNoChange);
+        fastNotOptimalAlgorithm = std::unique_ptr<CCoinsSelectionAlgorithmBase>(new CCoinsSelectionSlidingWindow(amountsAndSizes, targetAmount, targetAmountPlusOffset, availableTotalSize));
```
AndreaLanfranchi: This is very verbose.
```suggestion
        fastNotOptimalAlgorithm.reset(new CCoinsSelectionSlidingWindow(amountsAndSizes, targetAmount, targetAmountPlusOffset, availableTotalSize));
```
Besides ... destroying and recreating the object on every loop is quite expensive also considering you're passing `amountAndSizes` by copy (which is not necessary)
Consider using the "Reset()" method of the class

src/wallet/wallet.cpp
```diff
@@ -3024,82 +2982,161 @@ bool CWallet::SelectCoinsMinConf(const CAmount& nTargetValue, int nConfMine, int
...
+    
+    const CAmount targetAmount = nTargetValue;
+    const CAmount targetAmountPlusOffsetNoChange = nTargetValue; //no change
+    const CAmount targetAmountPlusOffsetMaxChange = std::get<2>(coinLowestLarger).tx != NULL ?
+                                                    std::min(std::get<1>(coinLowestLarger), nTotalLower) : nTotalLower; //max change
+    const size_t availableTotalSize = availableBytes;
+    std::unique_ptr<CCoinsSelectionAlgorithmBase> bestAlgorithm = nullptr;
+
+    std::unique_ptr<CCoinsSelectionAlgorithmBase> fastNotOptimalAlgorithm;
+    for (int i = 0; i < COINS_SELECTION_INTERMEDIATE_CHANGE_LEVELS + 2; ++i)
```
AndreaLanfranchi: Avoid putting operations in the condition statement. The compiler might be (in most cases) smart enough to understand it evaluates as a constant (such in this case) but this is better
```suggestion
    for (int i = 0, e = COINS_SELECTION_INTERMEDIATE_CHANGE_LEVELS + 2; i < e; ++i)
```

src/wallet/wallet.cpp
```diff
@@ -3024,82 +2982,161 @@ bool CWallet::SelectCoinsMinConf(const CAmount& nTargetValue, int nConfMine, int
...
+        slowOptimalAlgorithm->StartSolvingAsync();
+        for (int wait = 0; wait < GetArg("-coinsselectiontimeout", 4) * 10; ++wait)
+        {
+            if (slowOptimalAlgorithm->GetHasCompleted())
             {
-                setCoinsRet.insert(vValue[i].second);
-                nValueRet += vValue[i].first;
+                break;
             }
+            std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
AndreaLanfranchi: This way of putting the thread to sleep might result being expensive and delay the result by 1/10 of a second. Consider using a signal from the worker thread for completion or timeout

src/wallet/wallet.cpp
```diff
@@ -3024,82 +2982,161 @@ bool CWallet::SelectCoinsMinConf(const CAmount& nTargetValue, int nConfMine, int
...
+        }
+        if (!slowOptimalAlgorithm->GetHasCompleted())
+        {
+            slowOptimalAlgorithm->StopSolving();
+        }
+        CCoinsSelectionAlgorithmBase::GetBestAlgorithmBySolution(slowOptimalAlgorithm, fastNotOptimalAlgorithm, bestAlgorithm);
+        LogPrint("selectcoins", "Best algorithm: %s - %s", std::to_string((int)bestAlgorithm->type).c_str(), bestAlgorithm->ToString().c_str());
```
AndreaLanfranchi: There is no need to redundantly convert to `c_str()` : the placeholder for LogPrint is already %s hence a string.

AndreaLanfranchi (changes requested):

src/wallet/wallet.cpp
JackPiri: I agree with your suggestion.
Modified accordingly.

src/wallet/wallet.cpp
JackPiri: I agree with your suggestion.
Modified accordingly.

src/wallet/wallet.cpp
JackPiri: Inside method `SolveRecursive` the value of `optimalTotalSelection` is constantly updated as soon as new leaves of the tree are explored. Hence, chances are that the incomplete solution (meaning a solution based on the leaves explored up to now) found by `slowOptimalAlgorithm` can be better than the complete one found by `fastNotOptimalAlgorithm`.

src/wallet/wallet.cpp
JackPiri: I agree with your suggestion.
Modified accordingly (refactored all `vector`, `string`, `set`, `map`, `pair`, `tuple`, `make_pair`, `make_tuple` to _std::X_ version inside `wallet.cpp`).

src/wallet/wallet.cpp
AndreaLanfranchi: I am pretty sure you understand how from a logical perspective it does not make any sense.
How a dev (maybe external) is supposed to understand an "aborted" call has produced a reliable result ?

src/coinsselectionalgorithm.hpp
```diff
@@ -0,0 +1,483 @@
...
+    */
+    #if COINS_SELECTION_ALGORITHM_PROFILING
+    uint64_t GetExecutionMicroseconds();
+    #endif
+
+    //! Method for getting the optimal set of selected elements (true->selected, false->unselected)
+    /*!
+      \return the optimal set of selected elements
+    */
+    std::vector<char> GetOptimalSelection();
```
AndreaLanfranchi: Why do you return a copy of the vector instead of a const reference ?
In wallet.cpp this method is called inside a loop which have `problemDimension` iterations and for each of those a copy is created.

src/coinsselectionalgorithm.hpp
JackPiri: Pushed new commit for moving from raw pointers to vectors. This allows safer memory management and better maintainability.

Regarding performances, I need to clarify and explain the root cause of the degradation I experienced using vectors during prototypal phase.
The focus is on members `CCoinsSelectionAlgorithmBase::tempSelection` and `CCoinsSelectionAlgorithmBase::optimalSelection`; when refactoring those members from being `bool*` to being `std::vector<bool>` I didn't consider that `std::vector<bool>` _"is a specialized version of [vector](https://cplusplus.com/vector), which is used for elements of type bool and optimizes for space"_. Indeed, each bool element is in some way compressed/optimized _"so that each value is stored in a single bit"_; hence it seems to me reasonable that accessing elements through the index operator `[]` has to deal with a small overhead, that would eventually result in that ~10% performance degradation. In the end I decided to define those two members as `std::vector<char>`, preferring to favor processing over memory optimization in this situation.
Thanks to @drgora for pointing this out and orienting me on the right way.
Additional reference: https://cplusplus.com/reference/vector/vector-bool/.

src/coinsselectionalgorithm.hpp
JackPiri: Moved from raw pointers to vectors.

src/coinsselectionalgorithm.hpp
JackPiri: Moved from raw pointers to vectors.

src/coinsselectionalgorithm.cpp
JackPiri: Moved from raw pointers to vectors.

src/coinsselectionalgorithm.hpp
JackPiri: Moved from raw pointers to vectors.

src/coinsselectionalgorithm.cpp
```diff
@@ -0,0 +1,596 @@
...
+        bool bestAdmissibleFound = false; // "best" for this specific algorithm implementation
+        bool breakOuterLoop = false;
+
+        for (; !stopRequested && !breakOuterLoop && windowBackIndex >= 0; --windowBackIndex)
+        {
+            // quick reset before restarting with sliding window back index decreased by one position
+            for (int index = 0; index < problemDimension; ++index)
+            {
+                tempSelection[index] = 0;
+            }
```
AndreaLanfranchi: As you've changed from array to vector you have plenty of options to make the code more concise, more idiomatic ... instead of using this verbose loops with random accessor

Option 1
`tmpSelection.assign(problemDimension, 0);`

Option 2
`memset(&tmpSelection[0], 0, problemDimension * sizeof(tmpSelection[0]));`

Option 3 (which is the best describing the intent)
`std::fill(tmpSelection.begin(), tmpSelection.end(), 0)`

src/coinsselectionalgorithm.cpp
```diff
@@ -0,0 +1,596 @@
...
+    {
+        if (isRunning)
+        {
+            StopSolving();
+        }
+        for (int index = 0; index < problemDimension; ++index)
+        {
+            tempSelection[index] = 0;
+            optimalSelection[index] = 0;
+        }
```
AndreaLanfranchi: What about ?

```
tempSelection.assign(problemDimension, 0);
optimalSelection.assign(problemDimension, 0);
```

src/coinsselectionalgorithm.cpp
```diff
@@ -0,0 +1,596 @@
...
+                if (bestAdmissibleFound || windowFrontIndex == 0)
+                {
+                    optimalTotalSize = tempTotalSize;
+                    optimalTotalAmount = tempTotalAmount;
+                    optimalTotalSelection = tempTotalSelection;
+                    for (int index = 0; index < problemDimension; ++index)
+                    {
+                        optimalSelection[index] = tempSelection[index];
+                    }
+                    break;
```
AndreaLanfranchi: As you're doing a copy isn't it better to do this ?
```
optimalSelection.assign(tempSelection.begin(), tempSelection.end()); 
// or as option (less elegant)
// memcpy(optimalSelection.data(), tempSelection.data(), problemDimension); // note I omitted the conversion in bytes as we already know is a vector of bytes
                                                                            // but for the sake of completeness we should multiply for sizeof(elem)
```

src/coinsselectionalgorithm.cpp
```diff
@@ -0,0 +1,596 @@
...
+                            #if COINS_SELECTION_ALGORITHM_PROFILING
+                            ++reachedLeaves;
+                            #endif
+                            optimalTotalSize = tempTotalSizeNew;
+                            optimalTotalAmount = tempTotalAmountNew;
+                            optimalTotalSelection = tempTotalSelectionNew;
+                            for (int index = 0; index < problemDimension; ++index)
+                            {
+                                optimalSelection[index] = tempSelection[index];
+                            }
```
AndreaLanfranchi: See comments above about copying vector data

src/coinsselectionalgorithm.cpp
```diff
@@ -0,0 +1,596 @@
...
+                    // if best admissible solution already found or reached array end, set optimal solution
+                    if (bestAdmissibleFound || windowFrontIndex == 0)
+                    {
+                        optimalTotalSize = tempTotalSize + (numberOfJoinsplitsOutputsAmounts - joinsplitsOutputAmountIndex) * sizes[0]; // first element size is used (but actually all sizes are equal)
+                        optimalTotalAmount = tempTotalAmount;
+                        optimalTotalSelection = tempTotalSelection;
+                        for (int index = 0; index < problemDimension; ++index)
+                        {
+                            optimalSelection[index] = tempSelection[index];
+                        }
```
AndreaLanfranchi: See comments above for copying vector data

AndreaLanfranchi (changes requested):

src/coinsselectionalgorithm.cpp
```diff
@@ -0,0 +1,596 @@
...
+    optimalSelection = std::vector<char>(problemDimension, 0);
+
+    optimalTotalAmount = 0;
+    optimalTotalSize = 0;
+    optimalTotalSelection= 0;
+
+    isRunning = false;
+    stopRequested = false;
+    solvingThread = nullptr;
+    hasCompleted = false;
```
AndreaLanfranchi: As of C++11 you can (and best practices say you SHOULD) set default values in the declaration of non-static member variables

AndreaLanfranchi (changes requested):

src/coinsselectionalgorithm.cpp
```diff
@@ -0,0 +1,596 @@
...
+{
+    if (!stopRequested)
+    {
+        stopRequested = true;
+        if (solvingThread != nullptr)
+        {
+            solvingThread->join();
+            solvingThread = nullptr;
+        }
+    }
```
AndreaLanfranchi: Here stopRequested is set to true regardless whether there's an active thread which makes no sense.
The correct implementation should be

```
    bool expected = false;
    if (solvingThread && stopRequested.compare_exchange_strong(&expected, true)) {
        solvingThread->join(); // Wait for completion
        solvingThread.reset(); // Reset to nullptr
        stopRequested = false; // We don't need it anymore
    }
```

src/coinsselectionalgorithm.cpp
```diff
@@ -0,0 +1,596 @@
...
+                                  ",optimalTotalSize=" + std::to_string(optimalTotalSize) +
+                                  ",optimalTotalSelection=" + std::to_string(optimalTotalSelection) + "}\n";
+}
+
+void CCoinsSelectionAlgorithmBase::StartSolvingAsync()
+{
+    if (!isRunning)
+    {
+        solvingThread = std::unique_ptr<std::thread>(new std::thread(&CCoinsSelectionAlgorithmBase::Solve, this));
+    }
```
AndreaLanfranchi: isRunning MUST be set before starting a new thread.
If not this may happen :
- StartSolvingAsync is invoked when isRunning == false
- The scheduler takes some time to prepare the thread and synchronize the memory: also Solve() sets the atomic var to true AFTER having done Reset() so it takes time
- In the mean time another StartSolvingAsync is invoked : isRunning is still false and solvingThread is reset to a new thread making the original thread orphan.

AndreaLanfranchi (changes requested):

src/coinsselectionalgorithm.cpp
AndreaLanfranchi: In the context the var `isRunning` is totally useless: is enough to check whether solvingThread has a valid pointer to a thread

src/coinsselectionalgorithm.hpp
JackPiri: I agree with your suggestion.
Modified accordingly.

src/coinsselectionalgorithm.cpp
JackPiri: I agree with your suggestion.
Modified accordingly.

src/coinsselectionalgorithm.cpp
JackPiri: I agree with your suggestion.
Modified accordingly.

src/coinsselectionalgorithm.cpp
JackPiri: I agree with your suggestion.
Modified accordingly.

src/coinsselectionalgorithm.cpp
JackPiri: I agree with your suggestion.
Modified accordingly.

src/coinsselectionalgorithm.cpp
JackPiri: I agree with your suggestion.
Modified accordingly.

src/coinsselectionalgorithm.cpp
JackPiri: I agree with your suggestion.
Modified accordingly.

src/coinsselectionalgorithm.cpp
JackPiri: Based on current implementation, the scenario of a call to `CCoinsSelectionAlgorithmBase::StopSolving` without an active thread can happen if there's a need to stop (from another thread) a (synchronous) solving routine after it has started through `CCoinsSelectionAlgorithmBase::Solve` (please note that currently that's not the case, but I'd prefer to handle that scenario for possible future use).

In addition, the value of `stopRequested` should remain `true` after the actual stop beacuase it is used at the end of `CCoinsSelectionAlgorithmBase::Solve` method (`hasCompleted = !stopRequested;`) to distinguish between a solving routine that reached end due to stop request and a solving routine that reached end due to actual finding completion.

Putting together your suggestion and what I have written above, maybe the best implementation could be:
``
void CCoinsSelectionAlgorithmBase::StopSolving()
{
    if (!stopRequested && isRunning)
    {
        stopRequested = true;
        if (solvingThread != nullptr)
        {
            solvingThread->join();
            solvingThread.reset();
        }
    }
}
``

JackPiri: Based on current implementation, the scenario of a call to `CCoinsSelectionAlgorithmBase::StopSolving` without an active thread can happen if there's a need to stop (from another thread) a (synchronous) solving routine after it has started through `CCoinsSelectionAlgorithmBase::Solve` (please note that currently that's not the case, but I'd prefer to handle that scenario for possible future use).

In addition, the value of `stopRequested` should remain `true` after the actual stop beacuase it is used at the end of `CCoinsSelectionAlgorithmBase::Solve` method (`hasCompleted = !stopRequested;`) to distinguish between a solving routine that reached end due to stop request and a solving routine that reached end due to actual completion of optimal searching.

Putting together your suggestion and what I have written above, maybe the best implementation could be:

```
void CCoinsSelectionAlgorithmBase::StopSolving()
{
    if (!stopRequested && isRunning)
    {
        stopRequested = true;
        if (solvingThread != nullptr)
        {
            solvingThread->join();
            solvingThread.reset();
        }
    }
}
```

src/coinsselectionalgorithm.cpp
JackPiri: I agree with the criticality described in your scenario.
I cannot get rid of the variable `isRunning` because it is used to check if the sync solving is in progress; this means that checking for `solvingThread` is pointless if a sync solution has been requested through a direct call to `Solve` method. I have added another control variable `asyncStartRequested` that should handle properly the case you highlighted.

src/coinsselectionalgorithm.cpp
JackPiri: Simplified a little bit with:

```
void CCoinsSelectionAlgorithmBase::StopSolving()
{
    stopRequested = true;
    if (solvingThread != nullptr)
    {
        solvingThread->join();
        solvingThread.reset();
    }
}
```

src/coinsselectionalgorithm.hpp
AndreaLanfranchi: That is generally correct with one hint : actually memory is not optimized nor compressed. Is simply how memory access works.  

The problem is a bool is actually 1 bit (0 or 1) but  `sizeof(bool)` is implementation defined generally to 1 byte as computers can't access memory in denomination units less than that.  Hence std::vector<bool> as well as std::bitset allocate a minimum of 1 byte.

Hance to retrieve the boolean in a position of the vector implies :
- a division to find the byte where the bit we want to examine is stored (BN)
- a modulo operation to find the bit within the byte (bN)
- a bitwise operation to determine if the bit is set or not : `BN & (1 << bN)`

This is the overhead. Also endianness plays a role as we need to identify is the bit in question is RtL or LtR

src/coinsselectionalgorithm.cpp
AndreaLanfranchi: > Based on current implementation, the scenario of a call to CCoinsSelectionAlgorithmBase::StopSolving without an active thread can happen if there's a need to stop (from another thread) a (synchronous) solving routine after it has started through CCoinsSelectionAlgorithmBase::Solve (please note that currently that's not the case, but I'd prefer to handle that scenario for possible future use).

Should this be the case then the whole design is even more entangled and critical :
- either the Solve() method is threaded in both derived classes or it isn't
- if you're thinking "in the future" about possibly stopping `CCoinsSelectionSlidingWindow::Solve()` from a thread which has not started it then you need to redesign the whole base class (and derived ones to be thread-safe: all members exposed with getters must be protected with mutexes and locks on reads and writes otherwise you end up requesting ASYNCRONOUSLY a stop and immediately after you could call `GetOptimalSelection` which returns a vector whose content is totally undefined in lack of memory syncronization.

src/wallet/wallet.cpp
JackPiri: The async call followed by parent thread sleep has been replaced with sync call controlled through execution timeout (provided as constructor parameter).

src/wallet/wallet.cpp
JackPiri: The async call followed by parent thread sleep has been replaced with sync call controlled through execution timeout (provided as constructor parameter).

src/wallet/wallet.cpp
JackPiri: Added a comment in order to clarify the intent:
```
// the solution found by slowOptimalAlgorithm is checked even if the algorithm was unable to complete its full exploration
// due to timeout; since optimal solution is updated continuously, chances are that it is better than solution found by
// fastNotOptimalAlgorithm nevertheless
```

src/wallet/wallet.cpp
JackPiri: Modified using `std::unique_ptr::reset`.
`Reset()` method of the class cannot be used in this case, since it is designed for resetting internal variables and counters, input variables (like `targetAmountPlusOffset`) cannot be updated because they are enforced to be `const`.

JackPiri:
Moved to PR #576 after refactoring.

PR #497, opened 2022-12-13
Finalize sc_v2

drgora:
This PR updates sidechains v2 to prevent undeserved bans for honest nodes.

In particular, we add a new constraint on certificates for non ceasing sidechains: each certificate must reference a block whose height is greater or equal to the height of the block containing the previous certificate.

This also implies that it is never possible to send more than one certificate per block (hence the certificate rate is limited). Accordingly, we enforce that at any time no more than 1 certificate is present in the mempool for a non ceasing sidechain.

Finally, Python tests are updated to account for the new rules.


src/sc/sidechain.cpp
```diff
@@ -128,7 +128,7 @@ bool CSidechain::CheckQuality(const CScCertificate& cert) const
 bool CSidechain::CheckCertTiming(int certEpoch, int referencedHeight, const CCoinsViewCache& view) const {
     if (GetState(view) != State::ALIVE)
     {
-        return error("%s():%d - ERROR: certificate cannot be accepted, sidechain already ceased\n",
+        return error("%s():%d - ERROR: certificate cannot be accepted, sidechain not alive\n",
             __func__, __LINE__);
     }
```
a-petrini, line 134: Does the following comment still hold for ceasable sidechains?

qa/rpc-tests/sc_cert_nonceasing.py
```diff
@@ -219,94 +254,130 @@ def run_test(self):
...
+        assert_equal(bwt_amount_1, scinfo['items'][0]['lastCertificateAmount'])
+        assert_equal(cert_1, scinfo['items'][0]['lastCertificateHash'])
+        assert_equal(2, scinfo['items'][0]['lastCertificateQuality'])
+
+        cpool = self.nodes[0].getrawmempool()
+        if len(cpool) > 0:
+            assert(cert_2 in cpool)
+        if len(cpool) > 1:
+            assert(fwd_tx in cpool)
+        assert(len(cpool) <= 2)
```
ptagl: As far as I understand, asserts here are conditional because there might be the possibility that `fwt_tx` may have a dependency on cert_2, depending on the algorithm that selects available coins when calling `sc_send()` (an algorithm that is being refactored and improved on another branch).

 I think that an alternative approach would be to "force" `fwt_tx` to not have any dependency on the last certificate, for instance calling `sc_send()` with the optional parameter `minconf` set to a high value.
 
It would be great to discuss the two approaches and find all the pros and cons.

qa/rpc-tests/sc_cert_nonceasing.py
```diff
@@ -317,24 +388,30 @@ def run_test(self):
...
-        assert(cert_2 in cpool)
-        self.nodes[0].reconsiderblock(inv_bl_hash)
-        assert_equal(0, len(self.nodes[0].getrawmempool())) # all the certificates and fwt are in the blockchain again
+        # certificates 2 and 3 have no reference anymore, so they are gone
+        # fwd_tx may or may not depend on Cert2, so it may or may not be still around
+        if len(cpool) > 0:
+            assert(cert_1 in cpool)
+        if len(cpool) > 1:
+            assert(fwd_tx in cpool)
+        assert(len(cpool) <= 2)
```
ptagl: Same situation as for line 365 described in the previous comment.

src/sc/sidechain.cpp
```diff
@@ -142,11 +142,12 @@ bool CSidechain::CheckCertTiming(int certEpoch, int referencedHeight, const CCoi
                     __func__, __LINE__, certEpoch, lastTopQualityCertReferencedEpoch + 1);
         }
 
-        // Check ordering of references
-        if (referencedHeight <= lastReferencedHeight)
+        // Check that every certificate references a block whose commitment tree includes the previous certificate.
+        // This also implies referencedHeight > lastReferencedHeight.
```
a-petrini:
```suggestion
        // This also implies referencedHeight > lastInclusionHeight.
```

qa/rpc-tests/walletbackup.py
```diff
@@ -76,9 +76,9 @@ def setup_network(self, split=False):
 
         # nodes 1, 2,3 are spenders, let's give them a keypool=100
         extra_args = [
-           ['-debug=zrpc', "-keypool=100", ed0],
-           ['-debug=zrpc', "-keypool=100", ed1],
-           ['-debug=zrpc', "-keypool=100", ed2],
+           ['-debug=1', "-keypool=100", ed0],
+           ['-debug=1', "-keypool=100", ed1],
+           ['-debug=1', "-keypool=100", ed2],
```
ptagl: Is `-debug=1` a change left for debugging reasons or is it strictly needed?

I think we could take the opportunity to think more in general about how we use debug logs in Python tests, in particular, I would like to understand what is the impact on the execution time of those logs.

Examples of questions I have in mind:
- how long does it take to run all the Python tests without any debugging category set?
- how long does it take to run all the Python tests with a few essentials categories?
- how long does it take to run all the Python tests with `debug=1`?

Of course, it's not related to this PR, so I would do that later as a separate activity, I just wanted to leave a reminder here.

src/main.cpp
```diff
@@ -1301,17 +1290,24 @@ MempoolReturnValue AcceptCertificateToMemoryPool(CTxMemPool& pool, CValidationSt
...
+                LogPrint("mempool", "%s():%d - ERROR: cert[%s] refers to a non existing sidechain[%s]\n", __func__, __LINE__, certHash.ToString(), cert.GetScId().ToString());
+                return MempoolReturnValue::INVALID;
+            }
+
+            if (sc.isNonCeasing() && pool.certificateExists(cert.GetScId()))
+            {
+                state.Invalid(
+                    error("%s():%d - Dropping cert %s : conflicting with another cert in mempool for non ceasing SC\n",
+                        __func__, __LINE__, certHash.ToString()),
+                    CValidationState::Code::INVALID, "bad-sc-cert-conflict"); // TODO(dr): is this msg ok?
```
ptagl: The error message seems fine to me, the only alternative I could think of is something like `bad-non-ceasable-sc-cert-conflict`, but I don't know if you like it or if anyone else has a better idea.

src/txmempool.cpp
```diff
@@ -2023,16 +1973,12 @@ bool CCoinsViewMemPool::GetSidechain(const uint256& scId, CSidechain& info) cons
...
-            else
-            {
-                info.lastReferencedHeight = map_it->second;
+                LogPrint("mempool", "%s():%d - could not find referenced block for certTopQual %s. This is a problem.\n", __func__, __LINE__, certTopQual.GetHash().ToString());
+                assert(false);
             }
 
             info.lastTopQualityCertReferencedEpoch = certTopQual.epochNumber;
-            // info.lastInclusionHeight cannot be affected by mempool
+            info.lastInclusionHeight += 1; // assume this certificate would be included in the next block
```
ptagl: By increasing `lastInclusionHeight` do we really get the next block to be mined? Shouldn't we use the height of the `pcoinstip` instead or using INT_MAX (depending on the checks that are affected by `lastInclusionHeight`)?

src/undo.h
```diff
@@ -166,7 +166,6 @@ struct CSidechainUndoData
     // NONCEASING_CERT_DATA
     int prevInclusionHeight;
     int prevReferencedHeight;
```
ptagl: Since we removed `lastReferencedHeight` in the CSidechain class, do we still need the `prevReferencedHeight` here?

ptagl (changes requested):
The PR seems in general fine to me, I've only reported a few minor things that shouldn't affect the overall expected behavior of both ceasable and non-ceasable sidechains.

The main doubt I have is about the increment of `lastInclusionHeight` inside the `GetSidechain()` function (`txmempool.cpp`), but it shouldn't lead to any issue, also considering that we have an explicit check to avoid several certificates in the mempool or in a block for non-ceasable sidechains.

src/main.cpp
a-petrini: Fine for me!

qa/rpc-tests/sc_cert_getblocktemplate.py
```diff
@@ -246,7 +246,7 @@ def run_test_with_scversion(self, scversion, ceasable = True):
 
         ### 4: Check certificates ordering (by epoch number for non-ceasable sidechains, by quality for ceasable ones)
 
-        num_certificates = 10
+        num_certificates = 10 if ceasable else 1
         from_addresses = []
```
a-petrini: I wonder if it makes sense to perform the entire followin section just for ceasable sidechains, as for non-ceasable we create and test the ordering of one certificate

qa/rpc-tests/getblockexpanded.py
```diff
@@ -409,10 +409,19 @@ def run_test_non_ceasable(self, scversion):
...
-        # Add to mempool cert2 referring to (current block - 2)
+        # Mine 3 blocks, the first one containing cert1
+        mark_logs("########### Mine 3 blocks ##################", self.nodes, DEBUG_MODE)
+        cert1_block = self.nodes[0].generate(1)[0]
+        self.sync_all()
+        cert1_maturityHeight = self.nodes[0].getblockcount()
+        self.nodes[0].generate(2)
+        self.sync_all()
+
+        # Add to mempool cert3 referring to (current block - 1)
```
a-petrini:
```suggestion
        # Add to mempool cert2 referring to (current block - 1)
```

src/sc/sidechain.cpp
```diff
@@ -142,11 +142,12 @@ bool CSidechain::CheckCertTiming(int certEpoch, int referencedHeight, const CCoi
...
+        if (referencedHeight < lastInclusionHeight)
         {
-            return error("%s():%d - ERROR: certificate cannot be accepted, cert height (%d) not greater than last (%d)\n",
-                __func__, __LINE__, referencedHeight, lastReferencedHeight);
+            return error("%s():%d - ERROR: certificate cannot be accepted, cert height (%d) not greater than last certificate inclusion height (%d)\n",
```
JackPiri: I think `not greater than` should be restated as `lesser than`

qa/rpc-tests/sc_cert_nonceasing.py
```diff
@@ -81,19 +97,39 @@ def try_send_certificate(self, node_idx, scid, epoch_number, quality, ref_height
...
+            - can    send Cert N referencing a block higher than the inclusion of Cert N-1
+
+        - Dynamic interactions:
+            - last referenced and inclusion heights are still considered after nodes restart (tests serialization)
+            - can send tx with a fee not ok with the mempool one, but ok with the blockchain one
+            - on block disconnection certificates which remain valid are added to the mempool
+            - on block disconnection if a reference for a certificate is removed, such certificate is removed from the mempool
+            - on block connection, if a certificate for epoch N is included in the block for a given sc, any certificate in the mempool for such sidechain is removed
+            - disconnecting multiple blocks at once yields the expected result (oldest still valid certificate in mempool, others removed)
         '''
```
a-petrini, line 123: Great explanation!

a-petrini (approved):

JackPiri (approved):
The PR is ok to me, nothing special to report.
Thanks.

qa/rpc-tests/walletbackup.py
drgora: This was a leftover from my local dev view. I have restored the previous debugging flags now.

I also agree we should investigate the impact of debugging logs in future, and in case reconsider our current usage.

src/main.cpp
drgora: Thanks for your input, as agreed I left the more compact `bad-sc-cert-conflict`.

src/txmempool.cpp
drgora: You are right, this was indeed not the height of the blockchain, thanks for catching!
As agreed, I removed this line entirely as this check is now superfluous.

src/undo.h
drgora: Good point, I removed `prevReferencedHeight` from `CSidechainUndoData` entirely.

src/sc/sidechain.cpp
drgora: Yes, makes sense, thanks.

src/sc/sidechain.cpp
drgora: I updated the comments to make them more clear, thanks.

src/sc/sidechain.cpp
drgora: With `lastReferencedHeight` I really meant "the height referenced by the last certificate". I updated the comment to try make it less ambiguous, thanks.

src/main.cpp
drgora: Great, thanks, I removed the `TODO` comment

qa/rpc-tests/sc_cert_getblocktemplate.py
drgora: Yes I agree it makes sense, I removed the section altogether for non ceasing sidechains.

qa/rpc-tests/getblockexpanded.py
drgora: Good catch, fixed!

qa/rpc-tests/sc_cert_nonceasing.py
drgora: Thanks :)

qa/rpc-tests/sc_cert_nonceasing.py
drgora: Good idea, let's use `minconf` to make sure there are no dependencies and remove the switch.

qa/rpc-tests/sc_cert_nonceasing.py
drgora: Yes, same as above

ptagl (approved):

PR #498, opened 2022-12-15
Cache built dependencies for faster CI builds

AndreaLanfranchi:
Before
image

After
image

Also amended README about how to run `pip install`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -268,7 +268,8 @@ os: linux
         - depends/sources
         - depends/work
         - depends/x86_64-linux-gnu
-        - depends/x86_64-unknown-linux.gnu
+        - depends/x86_64-unknown-linux-gnu
+        - depends/x86_64-w64-mingw32
     script:
       - bash -c "${TRAVIS_BUILD_DIR}/contrib/ci-horizen/scripts/build/build_${TRAVIS_CPU_ARCH}_${DOCKER_TARGET_OS}_${DOCKER_FROM}.sh"
       - bash -c "${TRAVIS_BUILD_DIR}/contrib/ci-horizen/scripts/common/push_archive.sh ${B2_UL_COMPRESS_FOLDER} ${B2_UL_FILENAME}"
--- a/.travis.yml
+++ b/.travis.yml
@@ -264,6 +264,11 @@ os: linux
     cache:
       directories:
         - "$HOME/.ccache"
+        - depends/built
+        - depends/sources
+        - depends/work
+        - depends/x86_64-linux-gnu
+        - depends/x86_64-unknown-linux.gnu
     script:
       - bash -c "${TRAVIS_BUILD_DIR}/contrib/ci-horizen/scripts/build/build_${TRAVIS_CPU_ARCH}_${DOCKER_TARGET_OS}_${DOCKER_FROM}.sh"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -264,6 +264,11 @@ os: linux
     cache:
       directories:
         - "$HOME/.ccache"
+        - depends/built
+        - depends/sources
+        - depends/work
+        - depends/x86_64-linux-gnu
+        - depends/x86_64-unknown-linux.gnu
```
a-petrini: Can you check the correctness of this path (line 271)?
My local build makes a folder called
`depends/x86_64-unknown-linux-gnu`
and our CI drops the `unknown-` from the folder name, hence the path of line 270 is the one actually used in the three Ubuntu build for linux.

drgora: Is there any reason for this specific set of paths? Why not also `depends/x86_64-w64-mingw32`, for example?

AndreaLanfranchi: Good point. 2 things here :
1) I bad copied the path
2) I believe the "unknown" derives from builds on Windows' WSL

AndreaLanfranchi: Yes skimmed over it

.travis.yml
```diff
@@ -264,6 +264,12 @@ os: linux
     cache:
       directories:
         - "$HOME/.ccache"
+        - depends/built
+        - depends/sources
+        - depends/work
+        - depends/x86_64-linux-gnu
+        - depends/x86_64-unknown-linux-gnu
+        - depends/x86_64-w64-mingw32
```
ptagl: Should we also add a line for MacOS? Looking at one of the last Travis builds, it seems to be using the folder `x86_64-apple-darwin17.7.0`.

And another question: does it make sense to temporarily enable the OSX build on this branch to check if these changes have a significant impact on its execution time?

cronicc: For the sake of completeness it would be great to also include the paths for MacOS. I'm not 100% sure that caching the depends folder is safe though, there are circumstances which might require a rebuild from scratch:
- Changes to OS libraries in the docker images that we dynamically link against, e.g. libc6/glibc
- Changes to package versions in depends/packages/*.mk, the build system should detect this. But a manual test to verify this would be great.

cronicc (approved):
Approved but see comment.

.travis.yml
AndreaLanfranchi: No. This caches directives are for linux builds only (and cross built windows images).
OSX builds rely on a totally different OS image hence the caching directives must be placed elsewhere and need a real env to determine where the dependencies builds get into play.

AndreaLanfranchi: This caching directives are for linux OS ... OSX images need different directives and I wouldn' focus too much for the moment on a build which relies on outdated Xcode

ptagl: I made a rapid test on a new branch based on this one and the MacOS cache seems to be working fine, the build time is reduced from [45](https://app.travis-ci.com/github/HorizenOfficial/zen/builds/256088249) to [15 minutes](https://app.travis-ci.com/github/HorizenOfficial/zen/builds/259004966).

Anyway, if you think it's not worth including such changes, feel free to ignore them, the PR is ok for me!

ptagl (approved):

.travis.yml
AndreaLanfranchi: Cherry picked

PR #499, opened 2023-01-02
Remove ENABLE_ADDRESS_INDEXING pre-processor directive

AndreaLanfranchi:
The title says all.


src/init.cpp
```diff
@@ -1642,7 +1632,6 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
                 pblocktree->ReadString("indexVersion", indexVersionStr);
                 LogPrintf("%s: indexVersion %s\n", __func__, indexVersionStr);  
    
-#ifdef ENABLE_ADDRESS_INDEXING
                 // Check for changed -addressindex state
                 if (fAddressIndex != GetBoolArg("-addressindex", false)) {
```
cronicc: As we're including addressindexing in the general release with this PR we need to be sure that all conditions that would require a reindex on startup are detected. This code only covers detecting changed `-addressindex`,  changes to `-timestampindex` and `-spentindex` should also be detected here. I'm not sure if `-timestampindex` and `-spentindex` imply running with `-addressindex`, if that's the case we also need to check for it.

cronicc (changes requested):

src/init.cpp
AndreaLanfranchi: Good point ! It should have been done when new indexes have been introduced though cause this PR doesn't change a thing in logic. Anyway ... this is spit milk under the bridge.

The indexes are now : `-txindex -spentindex -timespentindex -addressindex -maturityheightindex`
Will try to address all of them.

AndreaLanfranchi: @cronicc please verify correctness for a8f2819 and eventually resolve the conversation.
Thank you

pawel8282 (commented):
> ``

pawel8282 (approved):

src/init.cpp
```diff
@@ -1553,19 +1547,16 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
...
     int64_t nBlockTreeDBCache = nTotalCache / 8;
 
-#ifdef ENABLE_ADDRESS_INDEXING
+
     if (GetBoolArg("-addressindex", DEFAULT_ADDRESSINDEX) || GetBoolArg("-spentindex", DEFAULT_SPENTINDEX)) {
         // enable 3/4 of the cache if addressindex and/or spentindex is enabled
         nBlockTreeDBCache = nTotalCache * 3 / 4;
     } else {
-#endif // ENABLE_ADDRESS_INDEXING
         if (nBlockTreeDBCache > (1 << 21) && !GetBoolArg("-txindex", false)) {
```
JackPiri: The default value is missing, this should be `GetBoolArg("-txindex", DEFAULT_TXINDEX)`.
Similar comment also at line 1029.

src/init.cpp
```diff
@@ -378,14 +378,12 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-txindex", strprintf(_("Maintain a full transaction index, used by the getrawtransaction rpc call (default: %u)"), 0));
```
JackPiri: The default value is missing, this should be `strprintf(_("Maintain a full transaction index, used by the getrawtransaction rpc call (default: %u)"), DEFAULT_TXINDEX)`.

src/init.cpp
```diff
@@ -378,14 +378,12 @@ std::string HelpMessage(HelpMessageMode mode)
...
     strUsage += HelpMessageOpt("-maturityheightindex", strprintf(_("Maintain a maturity height index that stores for every height the cerficates that became mature, used by the getblockexpanded rpc call. It requires -txindex (default: %u)"), 0));
```
JackPiri: The default value is missing, this should be `strprintf(_("Maintain a maturity height index that stores for every height the cerficates that became mature, used by the getblockexpanded rpc call. It requires -txindex (default: %u)"), DEFAULT_MATURITYHEIGHTINDEX)`.

src/main.h
```diff
@@ -109,16 +106,13 @@ static const unsigned int MAX_REJECT_MESSAGE_LENGTH = 111;
...
 
-#ifdef ENABLE_ADDRESS_INDEXING
+static const bool DEFAULT_TXINDEX = false;
+static const bool DEFAULT_MATURITYHEIGHTINDEX = false;
 static const bool DEFAULT_ADDRESSINDEX = false;
 static const bool DEFAULT_TIMESTAMPINDEX = false;
 static const bool DEFAULT_SPENTINDEX = false;
 static const unsigned int DEFAULT_DB_MAX_OPEN_FILES = 1000;
 static const bool DEFAULT_DB_COMPRESSION = true;
-#else
```
JackPiri, line 115: In my opinion the values for constants `DEFAULT_DB_MAX_OPEN_FILES` and `DEFAULT_DB_COMPRESSION` should stay respectively `64` and `false` as per previous non-ENABLE_ADDRESS_INDEXING scenario; maybe two additional constants can be provided (for example `DEFAULT_DB_MAX_OPEN_FILES_ADDRESS_INDEXING` and `DEFAULT_DB_COMPRESSION_ADDRESS_INDEXING`), their values (`1000` and `true` respectively) being used if `"-addressindex"` arg is enabled and the associated args (`"-blocktreedbmaxopenfiles"` and `"-blocktreedbcompression"`) are not provided by user.

contrib/ci-horizen/scripts/common/setup_environment.sh
```diff
@@ -12,8 +12,7 @@ NEED_PGP_SIGN_CREDS="false"
 export B2_DOWNLOAD_URL="https://f001.backblazeb2.com/file/${B2_BUCKET_NAME}/"
 
 if [[ $TRAVIS_TAG =~ ^.*-bitcore$ ]] || [[ $TRAVIS_BRANCH =~ ^AddressIndexing.*$ ]]; then
-  export ENABLE_ADDRESS_INDEX=1
-  export MAKEFLAGS="${MAKEFLAGS:-} --enable-address-indexing"
+  export ENABLE_ADDRESS_INDEX=1 # This is kept only as trigger for RPC tests (see rpc-tests.sh)
```
drgora: Wouldn't it make sense to enable these tests for all branches? Or maybe it would incur a significant overhead?

src/main.h
AndreaLanfranchi: On the matter please see https://github.com/bitcoin/bitcoin/pull/12495
The 64 value for MAX_DB_OPEN_FILES we enforced has a serious impact on performances and gives negligible advantage in memory savings.

Agree on enabling compression only for explicit user choice.

src/init.cpp
AndreaLanfranchi: Addressed

src/init.cpp
AndreaLanfranchi: Addressed

src/init.cpp
AndreaLanfranchi: Addressed

src/init.cpp
JackPiri: Same modification is needed at line 1029.

AndreaLanfranchi: This makes me question the whole logic behind `GetBoolArg()` and the defaults associated like DEFAULT_TXINDEX
All command line arguments parsed by GetBoolArg() are **flags** which mean `true` when explicitly defined and `false` when missing. Hence setting a default constant for those flags has no meaning and only adds unneded verbosity to code.

As those flags are options, should we change for instance at a certain point DEFAULT_TXINDEX to true it would mean the user not wanting that index should be forced to set `-txindex=false` or `-txindex=0`  which means forcing him to explictly deny the enforcement of an option (which is against the logic of how an option is perceived).

In case txindex should become enabled by default then we should change also command line arguments to `-no-txindex` to optionally disable the feature.

AndreaLanfranchi: Looking at the implementation of GetBoolArg() (util.cpp ~line 397) you see default is returned **only** if the command argument is not present and, as a result, being an **option** the default cannot be anything else than `false`

src/clientversion.cpp
```diff
@@ -19,11 +19,7 @@
  * for both bitcoind and bitcoin-core, to make it harder for attackers to
  * target servers or GUI users specifically.
  */
-#ifdef ENABLE_ADDRESS_INDEXING
-const std::string CLIENT_NAME("zenai");
```
a-petrini, line 23: We are now unifying the name of the client, could this create issues for users that has used zend with address indexing enabled?

src/rpc/server.cpp
```diff
@@ -370,14 +368,12 @@ static const CRPCCommand vRPCCommands[] =
     { "rawtransactions",    "fundrawtransaction",     &fundrawtransaction,     false },
 #endif
 
-#ifdef ENABLE_ADDRESS_INDEXING
     /* Address index */
     { "addressindex",       "getaddressmempool",      &getaddressmempool,      true  },
```
drgora, line 372: If used on a node that is not started with `-addressindex`, I believe `getaddressmempool` would produce no error, but return an empty response.
I'd suggest to be more explicit and throw an error, as it happens for the other RPCs. To be consistent with other RPCs, a good place to add a check and throw an error might be `CTxMemPool::getAddressIndex()`.

src/init.cpp
AndreaLanfranchi: Reading further things are even worse.
While parsing arguments (from CLI or Config file) at the very end the so called negative settings (i.e. a setting name prefixed by `-no` .e.g `-notxindex`) are parsed too but in a very clumsy and ineffective way.
The negation arg is effective only if the positive flag is **NOT** set which leads to the uncomfortable situation where the user can enter something like `./zend -txindex -notxindex` maybe expecting the indexing is disabled which instead is not true. 🤕

src/rpc/server.cpp
```diff
@@ -370,14 +368,12 @@ static const CRPCCommand vRPCCommands[] =
...
     { "addressindex",       "getaddressutxos",        &getaddressutxos,        false },
     { "addressindex",       "getaddressdeltas",       &getaddressdeltas,       false },
     { "addressindex",       "getaddresstxids",        &getaddresstxids,        false },
     { "addressindex",       "getaddressbalance",      &getaddressbalance,      false },
```
drgora, line 376: This is a minor comment, feel free to ignore.
If we want to keep the behavior consistent with what we have now, we should filter these RPCs form the list produced in `CRPCTable::help`, as it happens for the `hidden` category.

But maybe the new behavior is preferable, if we have appropriate warnings in place in the help messages, as suggested above.

drgora (changes requested):

src/main.h
AndreaLanfranchi: > Notice that bitcoin currently is still using 64 on non-windows 32bit systems.
Given that we also support linux 32bit, at least in theory, maybe it makes sense to do something similar?

I wonder nowadays who dares to rely on a 32 bit OS ... however the suggestion makes sense.

> On the compression, I'm wondering what happens if we disable compression on a local db that was created with compression? Maybe it makes sense to keep the defaults as they were before this patch, that is enable compression if a node is started with -addressindex?

I've tested that and the answer is : nothing happens. I'll explain:
Compression involves only  ldb files in `.zen/blocks/index` data directory and the parameter is relevant **only on flushing file to disk**. The compression info is written in the header of the file so when ldb have to reopen it it firstly read the meta info of the file and decide whether the content is compressed.
So basically this happens : start an instance with indexing and compression enabled causes newly generated files to be written to disk as compressed. Then stop the instance and restart with indexing but this time with compression disabled : newly generated files are not compressed but existing one still are and when ldb has to open/modify those it reads the compression flag from file's metadata not from the setting in the environment.

src/clientversion.cpp
AndreaLanfranchi: As far as I can see this value is advertised in p2p methods `getnetworkinfo` and `pushversion` but I don't see in the code wheter there's different treatment of the message for the nodes receiving it.

Nevertheless it "might" have an impact outside (maybe for Secure/Super nodes ? @cronicc ) even if I highly doubt that. Being not part of the protocol consensus anybody can customize their own code setting even a fantasy or vanity name.

src/rpc/server.cpp
AndreaLanfranchi: getaddressmempool returns an error

```
    if (!mempool.getAddressIndex(addresses, indexes)) {
        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "No information available for address");
    }
```

src/rpc/server.cpp
AndreaLanfranchi: As far as I can see all the rpc methods in the "addressindex" category return an error if related indexes are not enabled

src/rpc/server.cpp
drgora: True, but `CTxMemPool::getAddressIndex()` always returns `true`. Indeed, my suggestion was to make `CTxMemPool::getAddressIndex()` return `false` if `-addressindex` is not enabled.

AndreaLanfranchi: Sorry @drgora if `CTxMemPool::getAddressIndex()` always return true (which I can confirm) what is the point of having this ?

```
    if (!mempool.getAddressIndex(addresses, indexes)) {
        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "No information available for address");
    }
```
This gets never executed then

drgora: I agree with you, this is never executed as of today.
One way of fixing this might be removing these code lines altogether, but I think that it would be better to actually throw this error if addressindex is not enabled. In other words, I believe that throwing an explicit RPC error is way more clear than returning an empty result.

AndreaLanfranchi: The quickest way I believe is to check for fAddressIndex in CTxMemPool::getAddressIndex and return false if the first is false.
And secondly to change the check in RPC call as 

```
    if (!mempool.getAddressIndex(addresses, indexes)) {
        throw JSONRPCError(RPC_INTERNAL_ERROR, "Address indexing not enabled");
    } else if (indexes.empty()) {
        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "No information available for address");
    }
```
WDYT ?

drgora: I fully agree on the first point, that's actually what I meant to suggest, sorry if it was not clear.

On the second, I'm not sure. I'd say that an empty result is not an error per se (if addressindex is enabled), so we might even avoid throwing an error in that case. But I don't have strong opinions on this, your suggestion is also ok for me.

AndreaLanfranchi: @drgora ok ... give a second look at df63ccf

drgora: Lgtm, thanks

src/rpc/server.cpp
drgora: Yes I agree. I'm suggesting to update the help msg for the followings RPCs to state it clearly that some flags must be enabled for them to function correctly:
- getspentinfo (`spentindex`)
- getblockhashes (`timestampindex`)
- getblockdeltas (which misses the help msg completely, `spentindex`)

src/clientversion.cpp
cronicc: No impact expected, safe to unify.

AndreaLanfranchi: Thank you so much @cronicc

src/init.cpp
cronicc: @AndreaLanfranchi Resolving this conversation, but don't have the bandwidth to re-review fully at the moment. But @drgora and @a-petrini seem to be on it and have left some great feedback. To summarize, none of the defaults should change. Users running an old version of zend without or with provided switches that impact indexing, be it on addressindexing version of zend or normal zend should not experience breaking changes or have to reindex with the new unified version of zend.

Addressindexing RPC tests can be enabled by default, they don't add that much time to the test jobs. CI logic detection based on branch/tag name can be removed.

AndreaLanfranchi: Great thank you.

src/main.h
AndreaLanfranchi: Addressed in [27fab1e](https://github.com/HorizenOfficial/zen/pull/499/commits/27fab1e6bb4ec7a48ad65b587034d90159229c29)

src/rpc/server.cpp
AndreaLanfranchi: Addressed in [537f382](https://github.com/HorizenOfficial/zen/pull/499/commits/537f382619e2a423626689b6f3d5c4b60cf6a2e6)

Please complete help description for getblockdeltas

contrib/ci-horizen/scripts/common/setup_environment.sh
AndreaLanfranchi: Addressed in a031669

src/rpc/server.cpp
drgora: Pfa the description for getblockdeltas. Notice that this is mostly borrowed from zcash, and checked for consistency with our codebase.

[getblockdeltas.txt](https://github.com/HorizenOfficial/zen/files/10448602/getblockdeltas.txt)

AndreaLanfranchi: Thank you. Done in 1db7a1b

src/rpc/misc.cpp
```diff
@@ -501,50 +499,46 @@ bool getAddressFromIndex(const AddressType type, const uint160 &hash, std::strin
...
-    return true;
+    std::vector<std::pair<uint160, AddressType>> ret;
+    ret.reserve(value.size());
+    for (const auto& item : value.getValues()) {
+        ret.push_back(addressHashAndTypeFromValue(item));
+    }
+    return ret;
 }
 
-#ifdef ENABLE_ADDRESS_INDEXING
```
JackPiri, line 540: I think the specification _"(requires addressindex to be enabled)"_ should be dropped from the help-documentation of the following methods:

- `getaddressmempool`
- `getaddressutxos`
- `getaddressdeltas`
- `getaddressbalance`
- `getaddresstxids`

src/main.h
drgora: @AndreaLanfranchi please correct me if I'm wrong, but from what I can see I believe that this PR is changing the defaults in terms of compression.
Before this PR we have compression enabled by default for `addressindex`, and disabled otherwise. After this PR, we have compression disabled in all cases by default.

If the premise is correct, I think it would be better to restore the previous defaults in terms of compression, based on the runtime switch `-addressindex` (and probably also `-spentindex` and `-txindex`?).

src/rpc/server.cpp
drgora: Thank you

src/main.h
AndreaLanfranchi: This is not completely correct
It was the preprocessor directive `ENABLE_ADDRESS_INDEXING` which was setting compression to true, not the CLI argument `addressindex` hence compression was enabled regardless the user choice. (i.e. even if no -addressindex was set in instance launch)

drgora: Good point, I agree.

Still, my assumption is that the main use of the addressindex branch (i.e. the bitcore binary packages) is indeed with the `-addressindex` flag enabled at runtime. If that is correct, I am suggesting to minimize the number of users with different compression settings after this PR.

Also, I'm not sure about the rationale behind the choice of defaults for the compression, but I guess that the main reason is that addressindexing nodes store a greater amount of data, hence they want compression to be enabled to limit the db size on the storage. If this is correct, the impact of this PR for current bitcore nodes running without `-addressindex` (limited db size increase) would probably be smaller than the impact of disabling compression for actual addressindexing nodes (bigger db size increase).

With that being said, the ideal thing to do would probably be measuring the actual impact of compression on the db size and on access speed, and decide accordingly.

AndreaLanfranchi: I honestly don't see the point here.

Default db compression is set to false which adds no compression/decompression overhead (useful in most cases when users DO NOT use indexing features).
I have a fully synced node (at the time of this writing with this PR codebase) with all indexes enabled and no compression and the size of  `.zen/blocks/index` (the only one affected by compression) is 17 GiBi (total data size is roughly 82GiBi)

If that amount of space becomes an issue for some usecases the owner of the instance can freely enable the compression by adding `-blocktreedbcompression` argument to command line.

This IMHO adheres more to the concept of "optional feature"

AndreaLanfranchi: Prior of this PR instead the interpretation was ambiguous as the compression was ON by default for builds with the directive set (even if the user didn't want to enable indexing cli arguments) and if user did not want the index files to be compressed had to add `-noblocktreedbcompression` which is to me a bit of nonsense (you have to explicitly state you don't want what is meant to be an optional feature)

AndreaLanfranchi: As additional information I issued a reindex of the sync'ed node (block 1.27M roughly) with compression enabled as :

```
./src/zend -txindex -spentindex -timespentindex -addressindex -maturityheightindex -reindex -blocktreedbcompression 
```

After 15 hours the reindexing is still at block 390K with less than 10M total transactions indexed (at the tip total transactions are slightly more than 90M) and the total usage `.zen/blocks/index` is 2.8 GiBi.
As a comparative metric consider (on same hardware)  the full sync from scratch (hence including blocks download from peers) with compression disabled and **all indexes enabled** took about 18 hours.

ptagl:
Closing as the discussed changes have been merged with PR https://github.com/HorizenOfficial/zen/pull/531.

PR #500, opened 2023-01-05
Refactor limitedmap

drgora:
This PR updates limitedmap to fix some bugs, and at the same time refactor the code to remove duplication and improve efficiency and style.

### DIFF
--- a/src/limitedmap.h
+++ b/src/limitedmap.h
@@ -10,7 +10,7 @@
 
 /** STL-like map container that only keeps the N elements with the highest value. */
 template <typename K, typename V>
-class limitedmap
+class LimitedMap
 {
 public:
     typedef K key_type;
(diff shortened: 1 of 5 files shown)


src/limitedmap.h
```diff
@@ -26,68 +26,76 @@ class limitedmap
...
+        }
         std::pair<iterator, bool> ret = map.insert(x);
         if (ret.second) {
-            if (nMaxSize && map.size() == nMaxSize) {
-                map.erase(rmap.begin()->second);
-                rmap.erase(rmap.begin());
-            }
-            rmap.insert(make_pair(x.second, ret.first));
+            // in zend, we expect to insert increasing timestamps most of the times, hence the hint.
+            rmap.insert(rmap.end(), make_pair(x.second, ret.first));
```
AndreaLanfranchi: I see a problem here. The deletion of an element due to exceeded capacity is done **before** the actual insertion takes place (which might not if the key already exists).
Sample :
```
    LimitedMap<char,int> lmap(2);
    lmap.insert(std::pair<char,int>('a', 7));
    lmap.insert(std::pair<char,int>('b', 5));
    lmap.insert(std::pair<char,int>('a', 6));
    assert(lmap.find('b') != lmap.end());
```
The key 'b' is gone but we don't have exceeded capacity cause the new 'a' has not been inserted cause it already exist. size() now returns 1 but it should return 2

Also I would change the return type of `insert()` from void to bool to signal true -> the insertion has taken place; false -> no insertion performed as they key already exists

Also don't understand the comment
`// in zend, we expect to insert increasing timestamps most of the times, hence the hint.`

Apparently has nothing to do with `ret.second` nor with increasing values.

src/limitedmap.h
```diff
@@ -26,68 +26,76 @@ class limitedmap
...
+            if (it->second == map_it) {
+                return it;
+            }
+        }
+        return rmap.end();
+    }
+
 public:
-    limitedmap(size_type nMaxSizeIn = 0) { nMaxSize = nMaxSizeIn; }
+    LimitedMap(size_type nMaxSizeIn = 0) { nMaxSize = nMaxSizeIn; }
```
AndreaLanfranchi: I honestly don't understand the rational for letting this object be created with a max_size==0
It would prevent insertion of any element.
IMHO the default shold be removed and drop an assertion for values <1

AndreaLanfranchi (changes requested):

src/limitedmap.h
```diff
@@ -26,68 +26,76 @@ class limitedmap
...
     size_type max_size() const { return nMaxSize; }
-    size_type max_size(size_type s)
-    {
-        if (s)
-            while (map.size() > s) {
-                map.erase(rmap.begin()->second);
-                rmap.erase(rmap.begin());
-            }
+
+    size_type max_size(size_type s) {
```
AndreaLanfranchi: As for the ctor I'd drop an assertion for `s < 1`

AndreaLanfranchi:
Also would implementing a `clear()` method be helpful for reuse in other contexts ?

src/limitedmap.h
drgora: That's a great catch, thanks! I've fixed the problem with the last commit. I also modified the signature of `insert()` as suggested, and rephrased the comment. I hope it's more clear now.

src/limitedmap.h
drgora: Not sure what the rationale was, but I agree we don't have any use case for that. I applied your suggestion.

src/limitedmap.h
drgora: Yep

drgora:
Thanks for your feedback @AndreaLanfranchi. I'm not sure that adding `clear()` would be the best thing right now.
I believe that if it ever becomes necessary, then adding it would be real quick and easy. But until then, I'd rather leave the code compact, and avoid adding things that we do not use.

AndreaLanfranchi (approved):
Ok for me. Thank you for fast reaction.

AndreaLanfranchi:
Just a petty comment: since c++11 you can get rid of `make_pair` almost everywhere
e.g. 
```
    - rmap.insert(rmap.end(), make_pair(x.second, ret.first));
    + rmap.insert(rmap.end(), {x.second, ret.first}); 
```

src/limitedmap.h
```diff
@@ -26,68 +26,82 @@ class limitedmap
...
```
a-petrini: Do you think that `resize` would be a more appropriate name for this function? This would be more aligned with the naming convention of the STL containers which provide the same functionality (e.g. `vector`); also, it will differentiate its behavior from equally named `max_size()` get function.
However, this function is never called in the whole codebase, hence it does not impact anything else, at the moment.

a-petrini (approved):

src/limitedmap.h
```diff
@@ -26,68 +26,82 @@ class limitedmap
...
+        rmap.erase(rit);
+        map.erase(itTarget);
+        return 1; // be consistent with std::map return value
     }
-    void update(const_iterator itIn, const mapped_type& v)
-    {
-        // TODO: When we switch to C++11, use map.erase(itIn, itIn) to get the non-const iterator.
-        iterator itTarget = map.find(itIn->first);
+
+    void update(const_iterator itIn, const mapped_type& v) {
```
JackPiri: Possibly `update` method can have a `bool` type return value (similarly to `insert` method) in order to inform if an update has been actually performed or not; btw, this return value wouldn't be considered in current implementation, so feel free to ignore this comment.

JackPiri (approved):
This PR is fine to me.
Thanks

src/limitedmap.h
drgora: Makes sense, renamed to `resize()`, thanks.

src/limitedmap.h
drgora: Yeah makes sense, updated accordingly, thanks.

src/limitedmap.h
AndreaLanfranchi: This is imho arguable. In an STL container (which limitedmap pretends to emulate) typically a vector the `resize()` has the utility to add or remove elements in the container.
Actually this is not the case and this resize() behaves like an hybrid : has the same effect of resize() if the value passed is LT actual internal vector size while it behaves like `reserve()` when the value is GE

a-petrini: Thanks for clarifying that. I was concerned that having two functions with the same name - albeit with different signatures - may cause confusion, and `resize()` was a more appropriate name.
Can you suggest a name that can better represent the hybrid behaviour you pointed out?
Also, I believe I do not follow your comment about `reserve()`, since as far as I know, in STL containers this may trigger a reallocation, which never happens in `Limitedmap`.

AndreaLanfranchi: @a-petrini actually reserve() (for vector) might imply reallocations **but does not imply constructing/destructing instances of T** i.e. reserve() does not change the size() of a vector while a resize(), on the other hand, destroys as much elements needed to get size() equal to the new size or, if the new size is greater, creates new elements to reach the new count of elements specified by the value.

Here instead the resize() method behaves like vector's resize (for new values below actual size of the map) and like a reserve() for higher values. Hence an hybrid.

On the other hand IMHO the change of the max amount of items to keep once this limitedmap is created is a bit of a nonsense **and is not used anywhere in the code beside in tests**. Therefore my strong suggestion is to remove entirely the method and change nMaxSize to const (eventually a more convenient naming would be maxCapacity - for the getter also) and remove the associated unit tests. Same as for the (right) comment about clear() by @drgora 

>I believe that if it ever becomes necessary, then adding it would be real quick and easy. But until then, I'd rather leave the code compact, and avoid adding things that we do not use.

a-petrini: Clear, thanks for explaining! Also, I agree with the solution you are proposing.

drgora: Thank you guys, I agree, let's get rid of unused code. Done with the last update.

PR #504, opened 2023-01-26
Gp/zend233_zend234

JackPiri:
PR bringing in minor refactoring.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -687,34 +687,37 @@ void ThreadImport(std::vector<boost::filesystem::path> vImportFiles)
     if (fReindex || fReindexFast)
     {
         CImportingNow imp;
-        int nFile = 0;
+        CDiskBlockPos pos(0, 0);
         while (fReindexFast)
         {
-            CDiskBlockPos pos(nFile, 0);
             if (!boost::filesystem::exists(GetBlockPosFilename(pos, "blk")))
                 break; // No block files left to reindex
             FILE *file = OpenBlockFile(pos, true);
-            if (!file) break; // This error is logged in OpenBlockFile
-            uiInterface.InitMessage(_(tfm::format("Reindexing block headers from files... (blk%05u.dat)", (unsigned int)nFile).c_str()));
-            LogPrintf("Reindexing block file blk%05u.dat, headers-only...\n", (unsigned int)nFile);
+            assert(file); // This error is logged in OpenBlockFile
+            uiInterface.InitMessage(tfm::format(_("Reindexing block headers from files...") + "(blk%05i.dat)", pos.nFile));
+            LogPrintf("Reindexing block file blk%05i.dat, headers-only...\n", pos.nFile);
             LoadBlocksFromExternalFile(file, &pos, /*loadHeadersOnly*/true);
-            nFile++;
+            ++pos.nFile;
+            pos.nPos = 0;
         }
         if (fReindexFast)
+        {
             LogPrintf("Headers-only reindexing finished. Going on with blocks\n");
+            pos.nFile = 0;
+            pos.nPos = 0;
+        }
 
-        nFile = 0;
         while (true)
         {
-            CDiskBlockPos pos(nFile, 0);
             if (!boost::filesystem::exists(GetBlockPosFilename(pos, "blk")))
                 break; // No block files left to reindex
             FILE *file = OpenBlockFile(pos, true);
-            if (!file) break; // This error is logged in OpenBlockFile
-            uiInterface.InitMessageAfterLoading(_(tfm::format("Reindexing block from files... (blk%05u.dat)", (unsigned int)nFile).c_str()));
-            LogPrintf("Reindexing block file blk%05u.dat...\n", (unsigned int)nFile);
+            assert(file); // This error is logged in OpenBlockFile
+            uiInterface.InitMessageAfterLoading(tfm::format(_("Reindexing block from files...") + "(blk%05i.dat)", pos.nFile));
+            LogPrintf("Reindexing block file blk%05i.dat...\n", pos.nFile);
             LoadBlocksFromExternalFile(file, &pos, /*loadHeadersOnly*/false);
-            nFile++;
+            ++pos.nFile;
+            pos.nPos = 0;
         }
 
         pblocktree->WriteReindexing(false);
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -711,7 +711,7 @@ void ThreadImport(std::vector<boost::filesystem::path> vImportFiles)
                 break; // No block files left to reindex
             FILE *file = OpenBlockFile(pos, true);
             if (!file) break; // This error is logged in OpenBlockFile
-            uiInterface.InitMessage(_(tfm::format("Reindexing block from files... (blk%05u.dat)", (unsigned int)nFile).c_str()));
+            uiInterface.InitMessageAfterLoading(_(tfm::format("Reindexing block from files... (blk%05u.dat)", (unsigned int)nFile).c_str()));
             LogPrintf("Reindexing block file blk%05u.dat...\n", (unsigned int)nFile);
             LoadBlocksFromExternalFile(file, &pos, /*loadHeadersOnly*/false);
             nFile++;
@@ -2004,7 +2004,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     // ********************************************************* Step 11: finished
 
     SetRPCWarmupFinished();
-    uiInterface.InitMessage(_("Done loading"));
+    SetDoneLoadingUI();
 
 #ifdef ENABLE_WALLET
     if (pwalletMain) {
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -688,7 +688,6 @@ void ThreadImport(std::vector<boost::filesystem::path> vImportFiles)
     {
         CImportingNow imp;
         int nFile = 0;
-        if (fReindexFast) uiInterface.InitMessage(_("Reindexing block headers from files..."));
         while (fReindexFast)
         {
             CDiskBlockPos pos(nFile, 0);
@@ -696,6 +695,7 @@ void ThreadImport(std::vector<boost::filesystem::path> vImportFiles)
                 break; // No block files left to reindex
             FILE *file = OpenBlockFile(pos, true);
             if (!file) break; // This error is logged in OpenBlockFile
+            uiInterface.InitMessage(_(tfm::format("Reindexing block headers from files... (blk%05u.dat)", (unsigned int)nFile).c_str()));
             LogPrintf("Reindexing block file blk%05u.dat, headers-only...\n", (unsigned int)nFile);
             LoadBlocksFromExternalFile(file, &pos, /*loadHeadersOnly*/true);
             nFile++;
@@ -704,7 +704,6 @@ void ThreadImport(std::vector<boost::filesystem::path> vImportFiles)
             LogPrintf("Headers-only reindexing finished. Going on with blocks\n");
 
         nFile = 0;
-        uiInterface.InitMessage(_("Reindexing block from files..."));
         while (true)
         {
             CDiskBlockPos pos(nFile, 0);
@@ -712,6 +711,7 @@ void ThreadImport(std::vector<boost::filesystem::path> vImportFiles)
                 break; // No block files left to reindex
             FILE *file = OpenBlockFile(pos, true);
             if (!file) break; // This error is logged in OpenBlockFile
+            uiInterface.InitMessage(_(tfm::format("Reindexing block from files... (blk%05u.dat)", (unsigned int)nFile).c_str()));
             LogPrintf("Reindexing block file blk%05u.dat...\n", (unsigned int)nFile);
             LoadBlocksFromExternalFile(file, &pos, /*loadHeadersOnly*/false);
             nFile++;
--- a/src/metrics.cpp
+++ b/src/metrics.cpp
@@ -86,6 +86,7 @@ boost::synchronized_value<std::list<uint256>> trackedBlocks;
 
 boost::synchronized_value<std::list<std::string>> messageBox;
 boost::synchronized_value<std::string> initMessage;
+boost::synchronized_value<std::string> initMessageAfterLoading;
 bool loaded = false;
 
 extern int64_t GetNetworkHashPS(int lookup, int height);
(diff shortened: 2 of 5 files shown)


src/init.cpp
```diff
@@ -704,14 +704,14 @@ void ThreadImport(std::vector<boost::filesystem::path> vImportFiles)
...
         nFile = 0;
-        uiInterface.InitMessage(_("Reindexing block from files..."));
         while (true)
         {
             CDiskBlockPos pos(nFile, 0);
             if (!boost::filesystem::exists(GetBlockPosFilename(pos, "blk")))
                 break; // No block files left to reindex
             FILE *file = OpenBlockFile(pos, true);
             if (!file) break; // This error is logged in OpenBlockFile
+            uiInterface.InitMessage(_(tfm::format("Reindexing block from files... (blk%05u.dat)", (unsigned int)nFile).c_str()));
```
ptagl: Due to the internal logic of the `metrics thread` (responsible for the UI) this reindexing message is never shown because it gets continuously overwritten by the `printStats()` function.

I wouldn't remove the `InitMessage()` here at the moment, but it might be the case to investigate in the future how we can improve the update logic of the UI so that all the important information is shown. For my personal preference, I would like to see the normal block height counter increasing and the "reindex" message as well, so that the user can clearly understand that the blocks are not being downloaded from other peers.

src/sc/proofverifier.cpp
```diff
@@ -485,29 +485,28 @@ ProofVerificationResult CScProofVerifier::NormalVerifyCertificate(CCertProofVeri
...
+    if (!ret)
     {
         LogPrintf("ERROR: %s():%d - certificate proof with ID [%d] does not verify: code [0x%x]\n",
-            __func__, __LINE__, input.proofId, code);
+                  __func__, __LINE__, input.proofId, code);
+        return ProofVerificationResult::Failed;
     }
-
-    return ret ? ProofVerificationResult::Passed : ProofVerificationResult::Failed;
+    return ProofVerificationResult::Passed;
```
ptagl, line 530: Given that it's never possible to end up in a state in which the verifier returned `true` and the error code is different from `CctpErrorCode::OK`, what about adding an assert?

```suggestion
    assert(code == CctpErrorCode::OK);
    return ProofVerificationResult::Passed;
```

In case, let's do the same for the function `NormalVerifyCsw()`.

ptagl (changes requested):

src/sc/proofverifier.cpp
JackPiri: Makes sense, addressed in 222155ba23849d56ca62d84e830bb301e9c88186.

a-petrini (approved):

src/init.cpp
JackPiri: Addressed in 0d6cab92c1103b9fb2791a767c1599387ce3bff9.

src/metrics.cpp
```diff
@@ -86,6 +86,7 @@ boost::synchronized_value<std::list<uint256>> trackedBlocks;
 
 boost::synchronized_value<std::list<std::string>> messageBox;
 boost::synchronized_value<std::string> initMessage;
+boost::synchronized_value<std::string> initMessageAfterLoading;
```
a-petrini, line 89: Why creating another `boost::synchronized_value` object, and not using the same `initMessage`?
ThreadImport() sets its value sequentially (at first when doing header reindex, and then when doing a proper reindex). Outside the thread, `AppInit2()` also sets its value sequentially without the risk of being overwritten by other threads (`ThreadImport()` and `StartNode()`, which are started after the last access to `initMessage` by `AppInit2()`)

JackPiri: We have two different threads acting (in the sense of setting status and messages, a third thread running `ThreadShowMetricsScreen` is responsible of actually printing to GUI) here:

a->`AppInit2` (main thread)
1a) "Verifying wallet..."
2a) "Loading block index..."
3a) "Verifying blocks..."
4a) "Loading wallet..."
5a) "Activating best chain..."
5.5a) ThreadImport starts on another thread
6a) "Done loading" (via `SetDoneLoadingUI`)
7a) no printing if not importing, otherwise print importing info

b->`ThreadImport` (another thread)
1b) "Reindexing block headers from files...(blk%05u.dat)"
2b) "Reindexing block from files...(blk%05u.dat)"

Now, due to the synchronization mechanism:
```
while (!fRequestShutdown && chainActive.Tip() == NULL)
    MilliSleep(10);
```
the actual sequence of events is (considering for example 3 blkXXXXX.dat available):

...
4a)
5a)
1b) blk00000.dat, blk00001.dat, blk00002.dat
2b*) blk00000.dat
6a)
...) some minutes pass
2b**) blk00001.dat,
...) some minutes pass
2b***) blk00002.dat

Hence, if `initMessageAfterLoading` would not be present (having only `initMessage`):
1. the step 2b* would be immediately overwritten, with "Done Loading"
2. after printing "Done Loading" the GUI would not print `initMessage` (resulting in the user feeling all the loading has been completed and no importing is underway, guessing the possible blockchain progress due to p2p download)
3. after some minutes (this is the time required for importing a blkXXXXX.dat file; full importing, not headers only), step 2b** would trigger an update of `initMessage`

Conversely, introducing `initMessageAfterLoading`, the messages are stored in two different strings and the decision related to which string should be printed is based on the loading/importing status.

src/metrics.cpp
```diff
@@ -436,20 +444,33 @@ int printMessageBox(size_t cols)
...
+        msg = static_cast<std::string>(_("Init message: Done loading"));
+        doneLoadingPrinted = true;
+    }
+    else if (fImporting)
+    {
+        msg = static_cast<std::string>(_("Init message (node already loaded): ")) + static_cast<std::string>(*initMessageAfterLoading);
     }
 
+    std::cout << msg << std::endl;
+    std::cout << std::endl;
```
AndreaLanfranchi: std::endl means "append new line char and flush stream"
This should be written as 
```suggestion
    std::cout << msg << "\n" << std::endl;
```

src/init.cpp
```diff
@@ -688,14 +688,14 @@ void ThreadImport(std::vector<boost::filesystem::path> vImportFiles)
...
         int nFile = 0;
-        if (fReindexFast) uiInterface.InitMessage(_("Reindexing block headers from files..."));
         while (fReindexFast)
         {
             CDiskBlockPos pos(nFile, 0);
             if (!boost::filesystem::exists(GetBlockPosFilename(pos, "blk")))
                 break; // No block files left to reindex
             FILE *file = OpenBlockFile(pos, true);
             if (!file) break; // This error is logged in OpenBlockFile
+            uiInterface.InitMessage(_(tfm::format("Reindexing block headers from files... (blk%05u.dat)", (unsigned int)nFile).c_str()));
```
AndreaLanfranchi: Why casting nFile to unsigned int when it's simpler to change the format argument from 05u to 05i ?
Also conversion to c_str() is pleonastic and causes a double conversion.

The operator `_` tries translate the string produced by tfm::format. The way it is used makes no sense as :
- we don't actually provide any translation
- the way it is written would impose we translate all possible strings resulting from all possible combinations of blkxxxxx.dat

If we really want to keep the translation it should be written as 
uiInterface.InitMessage(tfm::format(_("Reindexing block headers from files...") + "(blk%05u.dat)", (unsigned int)nFile));

src/init.cpp
```diff
@@ -688,14 +688,14 @@ void ThreadImport(std::vector<boost::filesystem::path> vImportFiles)
     {
         CImportingNow imp;
         int nFile = 0;
-        if (fReindexFast) uiInterface.InitMessage(_("Reindexing block headers from files..."));
         while (fReindexFast)
         {
             CDiskBlockPos pos(nFile, 0);
```
AndreaLanfranchi: There is no point in initializing struct pos before we actually check for file existence.
Move this after the existence check

src/init.cpp
```diff
@@ -704,14 +704,14 @@ void ThreadImport(std::vector<boost::filesystem::path> vImportFiles)
...
 
         nFile = 0;
-        uiInterface.InitMessage(_("Reindexing block from files..."));
         while (true)
         {
             CDiskBlockPos pos(nFile, 0);
             if (!boost::filesystem::exists(GetBlockPosFilename(pos, "blk")))
                 break; // No block files left to reindex
             FILE *file = OpenBlockFile(pos, true);
             if (!file) break; // This error is logged in OpenBlockFile
```
AndreaLanfranchi: This logic is broken.
If OpenBlockFile returns NULL (considering we've already done existence check - which only verifies the name of the file exists in file system but it's location has not been reached yet) it means we likely have an IO error. In such case the loop breaks as if it has traversed all the files and considers he task done which is not true.

AndreaLanfranchi (changes requested):

src/init.cpp
JackPiri: Modified with:
`uiInterface.InitMessage(tfm::format(_("Reindexing block headers from files...") + "(blk%05i.dat)", nFile));`
and:
`uiInterface.InitMessageAfterLoading(tfm::format(_("Reindexing block from files...") + "(blk%05i.dat)", nFile));`.

src/metrics.cpp
JackPiri: Agree with your suggestion, modified accordingly.

src/init.cpp
JackPiri: Cannot move after because `pos `is required by `GetBlockPosFilename`.

src/init.cpp
JackPiri: Do you think adding a retry and if failing again printing an error message on screen (at log level this is already the case) can be ok?
In first place I opted for inserting an assert but maybe this is a little too severe; indeed if there is some error when importing, the node would keep the blockchain up to current importing level and proceed downloading the remaining blocks from p2p network.
Do you have any other proposal?

src/init.cpp
AndreaLanfranchi: What about an overload of function GetBlockPosFilename() which receives nFile instead of pos considering in that function the only relevant payload of pos is nfile ?

AndreaLanfranchi: Or even better ... Initialize only once outside the loop and for every iteration reset nfile and pos.
So we stop allocating and destructing on every loop.

src/init.cpp
AndreaLanfranchi: The problem here is that this thread is labelled "Importing" but actually serves the reindex purpose too.
If something goes south during reindex (or fast reindex) the data on disk is, as you say, to last importing level **but indexes** are inconsistent and we have wrongly stated that the reindex has completed correctly.
Which means all RPC calls which rely on correct indexes might return inconsistent data.

src/init.cpp
JackPiri: Yes, I have opted for constructing once outside loop and then resetting at the end of each iteration.

src/init.cpp
AndreaLanfranchi: I would say in this case an assert would be more appropriate

src/init.cpp
```diff
@@ -687,34 +687,35 @@ void ThreadImport(std::vector<boost::filesystem::path> vImportFiles)
...
-            LogPrintf("Reindexing block file blk%05u.dat, headers-only...\n", (unsigned int)nFile);
+            uiInterface.InitMessage(tfm::format(_("Reindexing block headers from files...") + "(blk%05i.dat)", pos.nFile));
+            LogPrintf("Reindexing block file blk%05i.dat, headers-only...\n", pos.nFile);
             LoadBlocksFromExternalFile(file, &pos, /*loadHeadersOnly*/true);
-            nFile++;
+            pos.nFile++;
```
AndreaLanfranchi: Prefer pre increment rather than post increment

src/init.cpp
```diff
@@ -687,34 +687,35 @@ void ThreadImport(std::vector<boost::filesystem::path> vImportFiles)
...
+            pos.nPos = 0;
         }
         if (fReindexFast)
             LogPrintf("Headers-only reindexing finished. Going on with blocks\n");
 
-        nFile = 0;
-        uiInterface.InitMessage(_("Reindexing block from files..."));
+        pos.nFile = 0;
+        pos.nPos = 0;
```
AndreaLanfranchi: Probably these two lines should stay in the `if (freindexFast)` branched code: in fact if that was false we never incremented pos and therefore no need to reset it to 0.0

AndreaLanfranchi (changes requested):

src/metrics.cpp
```diff
@@ -86,6 +86,7 @@ boost::synchronized_value<std::list<uint256>> trackedBlocks;
 
 boost::synchronized_value<std::list<std::string>> messageBox;
 boost::synchronized_value<std::string> initMessage;
+boost::synchronized_value<std::string> initMessageAfterLoading;
 bool loaded = false;
```
drgora: This is unlikely, but now that this variable is detached from a specific string, we should make sure that we do not have unexpected memory issues with memory order. I'd suggest to make this an atomic variable:
```suggestion
std::atomic<bool> loaded = false;
```

src/init.cpp
AndreaLanfranchi: I don't agree with Paolo's analysis.
As shown in Jira ticket the message of reindexing was printed and stuck to the very same value till the reindexing finishes.
I believe several additions here introduced were not really necessary

AndreaLanfranchi: The message "Done loading" (which sets loaded == true in metrics thread) is not shown until chainActive.Tip() != NULL: and this condition happens only after reindex has completed (main thread stays in sleep wait every 10 milliseconds)

src/metrics.cpp
AndreaLanfranchi: I agree with @a-petrini : all additions are not needed here.
This
```
while (!fRequestShutdown && chainActive.Tip() == NULL)
    MilliSleep(10);
```
actually suspends main thread until all reindexing finishes i.e. after InitBlockIndex is completely finalized (which instead does nothing if either fReindex or fReindexFast are true)

src/metrics.cpp
JackPiri: Agree with your suggestion, modified accordingly.

drgora (approved):
This is ok for me, thanks.

src/init.cpp
JackPiri: Modified accordingly.

src/init.cpp
JackPiri: Modified accordingly.

src/init.cpp
JackPiri: Modified accordingly.

src/init.cpp
ptagl: Unfortunately `chainActive.Tip()` is NULL only during the processing of the headers, as soon as the first block is read from file and processed it's not NULL anymore, that's why we need such change.

ptagl (approved):

ptagl:
Looks good to me, thanks!

@AndreaLanfranchi waiting for your approval to proceed with the merge.

AndreaLanfranchi (approved):

PR #513, opened 2023-02-15
Update boost to v1.81 and fix warnings

drgora:
This PR has the main goal of fixing compilation warnings generated when compiling with recent versions of GCC (v12) and clang (v15).

To fix all boost warnings, boost is updated to the latest stable version upstream (1.81).

Finally, we remove the project-wide `-Wno-nonnull`, and let boost use it only where actually needed.


JackPiri (approved):
The PR is fine to me, thanks.

ptagl (approved):

PR #515, opened 2023-02-27
Ap/zend 304

a-petrini:
This pull request adds several checks when building a sidechain tx commitment tree, managing the errors returned by CCTPcrypto lib.
In particular, the validity check is performed:

- before accepting a tx/cert in the mempool
- creating a block via getblocktemplate()
- validating an incoming block

### DIFF
--- a/qa/rpc-tests/sc_big_commitment_tree.py
+++ b/qa/rpc-tests/sc_big_commitment_tree.py
@@ -1,0 +1,252 @@
+#!/usr/bin/env python3
+# Copyright (c) 2014 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+from decimal import Decimal
+import time
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import assert_equal, initialize_chain_clean, \
+     start_nodes, connect_nodes_bi, assert_true, assert_false, mark_logs
+from test_framework.test_framework import ForkHeights, JSONRPCException
+from test_framework.blockchainhelper import BlockchainHelper
+
+NUMB_OF_NODES = 2
+DEBUG_MODE = 1
+EPOCH_LENGTH = 10
+CERT_FEE = Decimal('0.0001')
+
+'''
+This test excercises the txs commitment builder.
+CCTPlib stores the txs commitment tree as follows:
+- information for alive sidechains and ceased sidechains are stored on different subtrees
+- the txs commitment tree can contain up to 4096 subtrees (alive + ceased), each one for a different sidechain
+- each alive sidechain subtree has 3 subtrees, each one containing Forward Transfers, Backward Transfer Requests
+  and Certificates hashes in their leaves
+- each ceased sidechain subtree has 1 subtree containing Ceased Sidechain Withdrawal hashed in its leaves
+- FT / BWTR / CERT / CSW can contain up to 4095 leaves
+- Sidechain Creation transaction is NOT stored in a leaf, but in a separate field in the Rust structure that models
+  the txs commitment tree
+
+CCTPlib txs commitment builder fails when:
+- adding FTs / BWTRs / CERTs / CSWs belonging to more than 4096 different sidechains
+- adding more than 4095 FTs for a SC
+- adding more than 4095 BWTRs for a SC
+- adding more than 4095 CERTs for a SC
+- adding more than 4095 CSWs for a SC
+- adding a FT / BWTR / CERT for a SC after a CSW for the same SC has been added
+- adding a CSW for a SC after a FT / BWTR / CERT for the same SC has been added
+- internal hashing issues
+
+Currently, CommitmentBuilderGuard manages all these failure reasons but the last one.
+
+In this test, we call getblocktemplate() so that a txs commitment tree is built from the txs that are stored in the mempool of a node
+[CreateNewBlock(..) and related function in zend]
+1. we create a set of 4 transactions having in total 4094 FTs for a single SC; getblocktemplate() will not fail and returns a valid block
+     mempool <- {tx1[1024FT], tx2[1024FT], tx3[1024FT], tx4[1022FT]}
+     getblocktemplate() returns a block with [tx1[1024FT], tx2[1024FT], tx3[1024FT], tx4[1022FT]]
+2. adding another tx having 2 FTs for the same sc to the mempool will make getblocktemplate() ignore the newly added tx, as its addiction
+   makes the commitmentTreeBuilder fail. getblocktemplate returns the same block
+     mempool <- {tx1[1024FT], tx2[1024FT], tx3[1024FT], tx4[1022FT], tx5[2FT]}
+     getblocktemplate() returns a block with [tx1[1024FT], tx2[1024FT], tx3[1024FT], tx4[1022FT]]
+3. adding a single tx having a single FT for the same sc to the mempool does not make the commitmentTreeBuilder fail and the block
+   returned by getblocktemplate() will include the 4 txs of 1. and this newly added txs. tx of 2. is still rejected
+     mempool <- {tx1[1024FT], tx2[1024FT], tx3[1024FT], tx4[1022FT], tx5[2FT], tx6[1FT]}
+     getblocktemplate() returns a block with [tx1[1024FT], tx2[1024FT], tx3[1024FT], tx4[1022FT], tx6[1FT]]
+4. adding a single tx having a single FT for the same sc to the mempool will now make the commitmentTreeBuilder fail.
+     mempool <- {tx1[1024FT], tx2[1024FT], tx3[1024FT], tx4[1022FT], tx5[2FT], tx6[1FT], tx7[1FT]}
+     getblocktemplate() returns a block with [tx1[1024FT], tx2[1024FT], tx3[1024FT], tx4[1022FT], tx6[1FT]]
+5. We mine a block WITHOUT calling the getblocktemplate and confirm that only tx5 and tx7 are still in mempool
+     mempool <- {tx5[2FT], tx7[1FT]}
+'''
+
+class BigCommitmentTree(BitcoinTestFramework):
+
+    def setup_chain(self):
+        print("Initializing test directory " + self.options.tmpdir)
+        initialize_chain_clean(self.options.tmpdir, NUMB_OF_NODES)
+
+    def setup_network(self, split=False):
+        self.nodes = start_nodes(NUMB_OF_NODES, self.options.tmpdir,
+                                 extra_args=[['-logtimemicros=1', '-debug=sc', '-debug=py', '-debug=mempool', '-debug=cert']] * NUMB_OF_NODES)
+
+        connect_nodes_bi(self.nodes, 0, 1)
+        self.is_network_split = split
+        self.sync_all()
+
+    # This function waits until the getblocktemplate() command returns the given number of transactions
+    # within a specified time frame (default = 10 seconds, set to 0 seconds to wait indefinitely)
+    def wait_for_block_template(self, node_id, expected_num_transactions, timeout = 10):
+        delay = 1
+        iteration = 0
+
+        while timeout > 0 and iteration < timeout // delay:
+            current_block_template = self.nodes[node_id].getblocktemplate({}, True)
+
+            if len(current_block_template['transactions']) == expected_num_transactions:
+                return current_block_template
+
+            time.sleep(delay)
+            iteration = iteration + 1
+
+        return None
+
+    def run_test(self):
+        # Reach SC v2 fork height
+        mark_logs("Generating enough blocks to reach non-ceasing SC fork", self.nodes, DEBUG_MODE, 'e')
+        self.nodes[0].generate(ForkHeights['NON_CEASING_SC'])
+        self.sync_all()
+
+        test_helper = BlockchainHelper(self)
+
+        mark_logs("Node 0 creates a v2 sidechain (ceasing)", self.nodes, DEBUG_MODE, 'e')
+        test_helper.create_sidechain("test_sidechain", 2)
+        sc_id = test_helper.get_sidechain_id("test_sidechain")
+        mark_logs(f"Sidechain {sc_id} has been created", self.nodes, DEBUG_MODE, 'e')
+
+        self.nodes[0].generate(1)
+        self.sync_all()
+
+        # Send some coins to Node1
+        # By using a new address for Node1, we are sure that any transaction created by Node1
+        # will use the change of the previous one, forcing the getblocktemplate() function
+        # to sort transaction based on submission order (because any transaction will depend
+        # on the previous one)
+        node1_address = self.nodes[1].getnewaddress()
+        self.nodes[0].sendtoaddress(node1_address, 100)
+        self.sync_all()
+        self.nodes[0].generate(1)
+        self.sync_all()
+
+        current_sc_txs_commitment_tree_root = ""
+
+        forwardTransferOuts = [{'toaddress': "aabb", 'amount': Decimal(0.001), "scid": sc_id, "mcReturnAddress": node1_address}]
+
+        # Add to mempool 4 transactions, each containing 1024, 1024, 1024 and 1022 forward transfers (total 4094,
+        # which is almost the limit size of ScTxsCommTree)
+        mark_logs(f"Adding 4094 FT for sc {sc_id} to the mempool (4 transactions)", self.nodes, DEBUG_MODE, 'g')
+        multipll = [1024, 1024, 1024, 1022]
+        for i in range(4):
+            self.nodes[1].sc_send(forwardTransferOuts * multipll[i], {"fee": Decimal(0.0001 * (i + 1)), "minconf": 0})
+            mark_logs(f"Sent transaction #{i + 1}", self.nodes, DEBUG_MODE, 'c')
+            self.sync_all()
+
+            block_template = self.wait_for_block_template(0, i + 1)
+            assert(block_template is not None)
+            mark_logs(f"getblocktemplate() ->", self.nodes, DEBUG_MODE, 'e')
+            for k, transaction in enumerate(block_template['transactions']):
+                mark_logs(f"  {k} - {transaction['hash']}", self.nodes, DEBUG_MODE, 'e')
+            mark_logs(f"    ScTxsCommitmentTree root: {block_template['scTxsCommitment']}", self.nodes, DEBUG_MODE, 'e')
+
+            assert_false(current_sc_txs_commitment_tree_root == block_template['scTxsCommitment'])
+            current_sc_txs_commitment_tree_root = block_template['scTxsCommitment']
+
+
+        # Let's try to add two more forward transfers to the mempool to exceed ScTxsCommTree capacity
+        mark_logs(f"Adding 2 FT for sc {sc_id} to the mempool", self.nodes, DEBUG_MODE, 'y')
+        try:
+            exceeding_tx = self.nodes[1].sc_send(forwardTransferOuts * 2, {"fee": Decimal(0.0001 * 5), "minconf": 0})
+        except JSONRPCException as e:
+            print(e.error['message'])
+            assert(False)
+        mark_logs(f"Sent transaction FT #4095 and #4096", self.nodes, DEBUG_MODE, 'c')
+        mark_logs(f"  4 - {exceeding_tx}", self.nodes, DEBUG_MODE, 'e')
+        self.sync_all()
+
+        # Ask for a new block template
+        block_template = self.wait_for_block_template(0, i + 1)
+        assert(block_template is not None)
+        mark_logs(f"getblocktemplate() ->", self.nodes, DEBUG_MODE, 'e')
+        for k, transaction in enumerate(block_template['transactions']):
+            mark_logs(f"  {k} - {transaction['hash']}", self.nodes, DEBUG_MODE, 'e')
+        mark_logs(f"    ScTxsCommitmentTree root: {block_template['scTxsCommitment']}", self.nodes, DEBUG_MODE, 'e')
+
+        # The commitment tree should be the same as in the previous step because we rejected the last TX.
+        assert_equal(current_sc_txs_commitment_tree_root, block_template['scTxsCommitment'])
+        # Also, exceeding_tx should not have been included in the block
+        assert_false(exceeding_tx in [tx['hash'] for tx in block_template['transactions']], "The last TX should not be included in the block candidate!!!")
+
+
+        # Try again, but this time we only send the 4095th FT, so that it will be included in the block candidate
+        # We use createrawtransaction to avoid dependencies to the previous tx
+        mark_logs(f"Adding 1 FT for sc {sc_id} to the mempool", self.nodes, DEBUG_MODE, 'g')
+        try:
+            tx_data = {"address": "aabb", "amount": 0.003, "scid": sc_id, "mcReturnAddress": node1_address}
+            rawtx = self.nodes[0].createrawtransaction([],{},[], [], [tx_data])
+            fundedtx = self.nodes[0].fundrawtransaction(rawtx)
+            sigRawtx = self.nodes[0].signrawtransaction(fundedtx['hex'])
+            last_tx = self.nodes[0].sendrawtransaction(sigRawtx['hex'])
+        except JSONRPCException as e:
+            print(e.error['message'])
+            assert(False)
+        mark_logs(f"Sent transaction FT #4095", self.nodes, DEBUG_MODE, 'c')
+        mark_logs(f"  5 - {last_tx}", self.nodes, DEBUG_MODE, 'e')
+        self.sync_all()
+
+        # Ask for a new block template
+        block_template = self.wait_for_block_template(0, i + 2)
+        assert(block_template is not None)
+        current_sc_txs_commitment_tree_root = block_template['scTxsCommitment']
+        # Check it has been included in the block candidate
+        assert_true(last_tx in [tx['hash'] for tx in block_template['transactions']], "The last TX should be included in the block candidate!!!")
+        mark_logs(f"getblocktemplate() ->", self.nodes, DEBUG_MODE, 'e')
+        for k, transaction in enumerate(block_template['transactions']):
+            mark_logs(f"  {k} - {transaction['hash']}", self.nodes, DEBUG_MODE, 'e')
+        mark_logs(f"    ScTxsCommitmentTree root: {block_template['scTxsCommitment']}", self.nodes, DEBUG_MODE, 'e')
+
+
+        # Try again adding an exceeding tx to the block candidate
+        mark_logs(f"Adding 1 FT for sc {sc_id} to the mempool", self.nodes, DEBUG_MODE, 'y')
+        try:
+            exceeding_tx2 = self.nodes[1].sc_send(forwardTransferOuts, {"fee": Decimal(0.0001 * 7), "minconf": 0})
+        except JSONRPCException as e:
+            print(e.error['message'])
+            assert(False)
+        mark_logs(f"Sent transaction FT #4096", self.nodes, DEBUG_MODE, 'c')
+        mark_logs(f"  6 - {exceeding_tx2}", self.nodes, DEBUG_MODE, 'e')
+        self.sync_all()
+
+        # Ask for a new block template
+        block_template = self.wait_for_block_template(0, i + 2)
+        assert(block_template is not None)
+        mark_logs(f"getblocktemplate() ->", self.nodes, DEBUG_MODE, 'e')
+        for k, transaction in enumerate(block_template['transactions']):
+            mark_logs(f"  {k} - {transaction['hash']}", self.nodes, DEBUG_MODE, 'e')
+        mark_logs(f"    ScTxsCommitmentTree root: {block_template['scTxsCommitment']}", self.nodes, DEBUG_MODE, 'e')
+
+        # The commitment tree should be the same as in the previous step because we rejected the last TX.
+        assert_equal(current_sc_txs_commitment_tree_root, block_template['scTxsCommitment'])
+        # Also, exceeding_tx2 should not have been included in the block
+        assert_false(exceeding_tx2 in [tx['hash'] for tx in block_template['transactions']], "The last TX should not be included in the block candidate!!!")
+
+
+        # Now we mine the block that includes the 5 valid txs, then we make sure that exceeding_tx and exciding_tx2 is still in mempool
+        mark_logs(f"Mine a block", self.nodes, DEBUG_MODE, 'c')
+        self.nodes[0].generate(1)
+        self.sync_all()
+        mark_logs(f"Confirm that only the exceeding txs are still in mempool", self.nodes, DEBUG_MODE, 'g')
+        a = self.nodes[1].getrawmempool(True)
+        assert_equal(len(a), 2)
+        assert_true(exceeding_tx  in a, "tx {exceeding_tx} should still be in node 1 mempool")
+        assert_true(exceeding_tx2 in a, "tx {exceeding_tx2} should still be in node 1 mempool")
+
+        # Mine a block to clean the mempool (and make sure that exceeding_tx and exceeding_tx2 have been included in the block)
+        mark_logs(f"Mine a block", self.nodes, DEBUG_MODE, 'c')
+        self.nodes[0].generate(1)
+        self.sync_all()
+        mark_logs(f"Confirm that the exceeding txs have been included in a block", self.nodes, DEBUG_MODE, 'g')
+        a = self.nodes[1].getrawmempool(True)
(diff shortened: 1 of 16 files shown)


src/miner.cpp
```diff
@@ -809,9 +845,15 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn,  unsigned int nBlo
         // Fill in header
         pblock->hashPrevBlock  = pindexPrev->GetBlockHash();
 
-        if (pblock->nVersion == BLOCK_VERSION_SC_SUPPORT )
+        if (pblock->nVersion == BLOCK_VERSION_SC_SUPPORT)
         {
-            pblock->hashScTxsCommitment = pblock->BuildScTxsCommitment(view);
+            pblock->hashScTxsCommitment.SetNull();
+            pblock->BuildScTxsCommitment(view, pblock->hashScTxsCommitment);
```
drgora: We might assert that result is `true` here?

qa/rpc-tests/sc_big_commitment_tree.py
```diff
@@ -0,0 +1,254 @@
...
+            exceeding_tx = self.nodes[1].sc_send(forwardTransferOuts * 2, {"fee": Decimal(0.0001 * 5), "minconf": 0})
+        except JSONRPCException as e:
+            print(e.error['message'])
+            assert(False)
+        mark_logs(f"Sent transaction FT #4095 and #4096", self.nodes, DEBUG_MODE, 'c')
+        mark_logs(f"  4 - {exceeding_tx}", self.nodes, DEBUG_MODE, 'e')
+        self.sync_all()
+
+        # Ask for a new block template
+        time.sleep(5)
```
ptagl: Can we get rid of this `sleep` now that we have the `wait_for_block_template()` function?

```suggestion
```

qa/rpc-tests/sc_big_commitment_tree.py
```diff
@@ -0,0 +1,254 @@
...
+            exceeding_tx2 = self.nodes[1].sc_send(forwardTransferOuts, {"fee": Decimal(0.0001 * 7), "minconf": 0})
+        except JSONRPCException as e:
+            print(e.error['message'])
+            assert(False)
+        mark_logs(f"Sent transaction FT #4096", self.nodes, DEBUG_MODE, 'c')
+        mark_logs(f"  6 - {exceeding_tx2}", self.nodes, DEBUG_MODE, 'e')
+        self.sync_all()
+
+        # Ask for a new block template
+        time.sleep(5)
```
ptagl: Can we get rid of this sleep now that we have the wait_for_block_template() function?

```suggestion
```

src/gtest/test_libzendoo.cpp
```diff
@@ -2491,3 +2500,409 @@ TEST(CctpLibrary, TestCustomFieldsValidationFix)
...
+    // ceased sidechain withdrawals
+    mtx3.vcsw_ccin.push_back(csw);
+    CTransaction tx3(mtx3);
+
+    // tx with CSW after tx with FT
+    SidechainTxsCommitmentBuilder cmtObj2;
+    ASSERT_TRUE(cmtObj2.add(tx2));
+    ASSERT_FALSE(cmtObj2.add(tx3));
+
+    // tx with CSW after tx with FT
```
ptagl: Is it possible that the comment here should be reversed?

```suggestion
    // tx with FT after tx with CSW
```

src/main.cpp
```diff
@@ -1142,6 +1142,13 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
         }
     }
 
+    // checking txs commitment tree validity
+    SidechainTxsCommitmentBuilder scCommitmentBuilder;
+    if (!scCommitmentBuilder.add(tx)) {
+        return state.DoS(100, error("%s(): SidechainTxsCommitmentBuilder failed\n", __func__),
+                        CValidationState::Code::INVALID, "bad-txns-txscommitmenttree-fail");
+    }
+
```
ptagl: I see here that the check prevents a single transaction from exceeding the capacity of the commitment tree, and that in case the sender node is banned, it's a very important check.

I wonder if we could find another place for this, for instance at the end of `Sidechain::checkTxSemanticValidity()`, so that it is performed both for the mempool and the block validation.

src/sc/sidechainTxsCommitmentBuilder.h
```diff
@@ -42,6 +42,28 @@ class SidechainTxsCommitmentBuilder
...
+        uint32_t cert = 0;
+    };
+    struct CommitmentBuilderGuardCeasedCounter {
+        uint32_t  csw = 0;
+    };
+    struct CommitmentBuilderGuard {
+        std::map<uint256, CommitmentBuilderGuardAliveCounter>  cbgaMap;
+        std::map<uint256, CommitmentBuilderGuardCeasedCounter> cbgcMap;
+        // The following values MUST be aligned with those specified in CCTPlib!
+        const int   SC_LIMIT = 4096;
```
drgora: I think it would be good to take these values directly from CCTPlib. I understand this requires updating external libraries, so it is ok to leave it for another PR.

Also, I believe it makes sense to make these constants `static` to save some memory.

src/main.cpp
```diff
@@ -1317,6 +1324,15 @@ MempoolReturnValue AcceptCertificateToMemoryPool(CTxMemPool& pool, CValidationSt
...
 
+            // checking txs commitment tree validity
+            SidechainTxsCommitmentBuilder scCommitmentBuilder;
+            if (!scCommitmentBuilder.add(cert, view)) {
+                state.Invalid(
+                        error("%s(): SidechainTxsCommitmentBuilder failed\n", __func__),
+                                CValidationState::Code::INVALID, "bad-sc-cert-txscommitmenttree-fail");
+                return MempoolReturnValue::INVALID;
+            }
+
```
ptagl: Similar to what I mentioned in the previous comment, would it be possible to move this check to `Sidechain::checkCertSemanticValidity()`?

src/rpc/mining.cpp
```diff
@@ -731,7 +731,8 @@ UniValue getblocktemplate(const UniValue& params, bool fHelp)
 
         if (certSupported) {
             CCoinsViewCache view(pcoinsTip);
-            pblock->hashScTxsCommitment = pblock->BuildScTxsCommitment(view);
+            pblock->hashScTxsCommitment.SetNull();
+            pblock->BuildScTxsCommitment(view, pblock->hashScTxsCommitment);
```
ptagl: If I understand correctly, if we reach this line the commitment tree root should be computed without errors.

What about adding an `assert` on the return value just to be sure that we don't return a valid block with an invalid commitment tree in case of malfunctions?

src/sc/sidechainTxsCommitmentBuilder.cpp
```diff
@@ -151,15 +151,49 @@ bool SidechainTxsCommitmentBuilder::add_fwt(const CTxForwardTransferOut& ccout,
...
-         &bws_tx_hash,
-         out_idx,
-         &ret_code
-    );
+    // Check against the number of sidechains currently in the commitment tree
+    if ((cbg.cbgaMap.size() + cbg.cbgcMap.size()) == cbg.SC_LIMIT) {
+        LogPrint("sc", "%s():%d - scTxsCommitment building failed: too many sidechains, when adding scId[%s].\n",
+            __func__, __LINE__, ccout.GetScId().ToString());
+        return false;
+    }
```
ptagl: Would it be possible that this check fails in the following scenario?
- A transaction with `SC_LIMIT` sidechains
- One _Forward Transfer Output_ for each sidechain
- Then we add a new _Forward Transfer Output_ for an **existing sidechain**

If that's the case, should we perform the check only if the sidechain ID is not already present in the tree?

src/sc/sidechainTxsCommitmentBuilder.cpp
```diff
@@ -276,19 +381,54 @@ bool SidechainTxsCommitmentBuilder::add_cert(const CScCertificate& cert, const S
...
+    // The sidechain id should not be already in the ceased tree
+    if (cbg.cbgcMap.count(cert.GetScId()) > 0) {
+        LogPrint("sc", "%s():%d - scTxsCommitment building failed: adding cert for scId[%s] on alive subtree, but already added on ceased subtree.\n",
+            __func__, __LINE__, cert.GetScId().ToString());
+        return false;
+    }
+
+    CommitmentBuilderGuardAliveCounter& counterRef = cbg.cbgaMap[cert.GetScId()];
+
+    // Check against the number of BWT already present in the sc subtree.
```
ptagl:
```suggestion
    // Check against the number of certificates already present in the sc subtree.
```

ptagl (changes requested):
In general, the pull request seems fine to me, I've only left a few comments for possible improvements, thanks!

As we talked about offline, we still need to bind the additional checks to the fork manager for a controlled introduction.

Additionally, I was wondering if it could be useful to add a couple of more tests (or extending the existing ones):
-  a test that adds 4095 FTs, one for each sidechain, and then adds a second FT for any sidechain (this test should fail now but should pass after fixing the checks in the Commitment Builder)
- a test that tries to completely fill a commitment tree with 4095 sidechain, 4095 certificates, 4095 forward transfers, etc. If it takes too long to generate proofs, we could try to generate fake ones.

qa/rpc-tests/sc_big_commitment_tree.py
```diff
@@ -0,0 +1,254 @@
...
+        mark_logs(f"Sent transaction FT #4095", self.nodes, DEBUG_MODE, 'c')
+        mark_logs(f"  5 - {last_tx}", self.nodes, DEBUG_MODE, 'e')
+        self.sync_all()
+
+        # Ask for a new block template
+        block_template = self.wait_for_block_template(0, i + 2)
+        assert(block_template is not None)
+        current_sc_txs_commitment_tree_root = block_template['scTxsCommitment']
+        # Check it has been included in the block candidate
+        assert_true(last_tx in [tx['hash'] for tx in block_template['transactions']], "The last TX should not be included in the block candidate!!!")
```
drgora: I believe this should be "The last TX should be included in the block candidate!!!"

src/sc/sidechainTxsCommitmentBuilder.cpp
JackPiri: I'd also consider moving from equality check to inequality check (even if I understand fwt are added one by one).

Possibly a check like this:

```
if ((cbg.cbgaMap.size() + cbg.cbgcMap.size()) + (1 - cbg.cbgaMap.count(ccout.GetScId())) > cbg.SC_LIMIT)
```

could handle the scenario described above?

Similar comment applies to `add_bwtr` and `add_cert` (`cbg.cbgaMap`) and `add_csw` (`cbg.cbgcMap`).

src/primitives/block.h
```diff
@@ -183,7 +183,7 @@ class CBlock : public CBlockHeader
 
     // return the sc txs commitment calculated as described in zendoo paper. It is based on contribution from
```
JackPiri: Maybe the description of this method can be adapted like:

```
// Builds the sc txs commitment calculated as described in zendoo paper. It is based on contribution from
// sidechains-related txes and certificates contained in this block. Returns the building result
```

src/miner.cpp
```diff
@@ -528,6 +529,15 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn,  unsigned int nBlo
         return NULL;
     CBlock *pblock = &pblocktemplate->block; // pointer for convenience
 
+    // We create a temporary sidechain tx commitment builder and populate it
+    // with the txs and certs that are progressively added to the candidate block.
+    // Each time this happens we check for failures that may be caused by:
+    // - reaching the SC subtree limit (currently max 4096 sidechains)
+    // - reaching the FT / BWT / CERT / CSW limit in each subtree (currently max 4096 in each category subtree per sc)
```
JackPiri: Very minor point: for `FT / BWT / CERT / CSW` the limit currently is `4095`; I wonder if it is maybe more maintainable to drop the numeric values and maybe just reference the associated constant variables names.

drgora (commented):
I left few comments, none of which is fundamental.
This looks good for merge, once [this](https://github.com/HorizenOfficial/zen/pull/515/files#r1119076368) comment is solved. Thanks.

src/main.cpp
JackPiri: @ptagl It seems to me this is already available for both `AcceptTxToMemoryPool` and `CheckBlock`.

src/main.cpp
```diff
@@ -1317,6 +1324,15 @@ MempoolReturnValue AcceptCertificateToMemoryPool(CTxMemPool& pool, CValidationSt
                 }
             }
 
+            // checking txs commitment tree validity
+            SidechainTxsCommitmentBuilder scCommitmentBuilder;
+            if (!scCommitmentBuilder.add(cert, view)) {
+                state.Invalid(
```
JackPiri: I have two questions here:

1. shouldn't this piece of code be inside `CheckCertificate` rather than in `AcceptCertificateToMemoryPool`? in this way the code would be more aligned with transaction analogous (`CheckTransactionWithoutProofVerification`) and even more important the check would involve both mempool-validation as well as block-validation

2. why in this case failure results in a call to `state.Invalid` while on the transaction analogous (`CheckTransactionWithoutProofVerification`) failure results in a call to `state.DoS`?

src/gtest/test_libzendoo.cpp
```diff
@@ -25,6 +25,13 @@ extern unsigned char ReverseBitsInByte(unsigned char input);
 
 using namespace blockchain_test_utils;
 
+// Keep them aligned with those defined in CCTPlib!
```
JackPiri, line 29: Can we leverage `getCBG()` method to insert a static assert in order to make sure constant values in `test_libzendoo.cpp` are aligned with those in` sidechainTxsCommitmentBuilder.h`? Or at least add a check within the test execution?

src/gtest/test_libzendoo.cpp
```diff
@@ -2491,3 +2500,409 @@ TEST(CctpLibrary, TestCustomFieldsValidationFix)
...
+    ASSERT_EQ(cbgRef.cbgaMap[sidechainId].cert, 0);
+}
+
+TEST(CctpLibrary, CommitmentBuilder_toomanyCERT) {
+    SelectParams(CBaseChainParams::REGTEST);
+    const BlockchainTestManager& testManager = BlockchainTestManager::GetInstance();
+
+    SidechainTxsCommitmentBuilder cmtObj;
+
+    CTransaction scCreationTx = txCreationUtils::createNewSidechainTxWith(CAmount(10), /*height*/10);
```
JackPiri: Very low important and coming from previous commit; I think the comment `/*height*/` should be removed (the value `10` being `epochLength`); similar at lines 584 and 2824.

qa/rpc-tests/sc_big_commitment_tree.py
```diff
@@ -0,0 +1,254 @@
+#!/usr/bin/env python3
```
JackPiri, line 1: I see in this file (line 248) and in `sc_big_commitment_tree_getblockmerkleroot.py` (lines 49, 51, 57) the critical values 4095/4096 being used as numeric constant rather than constant variables.
Maybe a quick change can be to make them global/shared within Python modules (like for example `ForkHeights`).
As a wider consideration, I see those values being duplicated in cpp production code, cpp test code (gtest) and python test code, in addition to rust code (cctplib); this can be a valid point supporting a change to cctplib and mccryptolib in order to avoid duplications and possible silent bugs (something related is under internal evaluation).

JackPiri (changes requested):
The PR is overall fine to me, but I have left some comments (some to be addressed, other with minor importance)

qa/rpc-tests/sc_big_commitment_tree.py
a-petrini: Addressed in e6db47d33

qa/rpc-tests/sc_big_commitment_tree.py
a-petrini: Addressed in e6db47d33

src/gtest/test_libzendoo.cpp
a-petrini: Correct. Addressed in e6db47d33

src/main.cpp
a-petrini: Checks on the commitment tree are performed in different places, depending on which entity we are currently checking, covering all the possibilities:
- for incoming blocks, we test the correctness of the txs commitment `CheckBlock()`, by building a `SidechainTxsCommitmentBuilder` and progressively adding all the txs an certs contained within
- we do the same thing in `CreateNewBlock()` for blocks we are generating i.e. through `getblocktemplate()`
- for an incoming tx to be added to the mempool, we do this check in `AcceptTxToMemoryPool()->CheckTransaction()->CheckTransactionWithoutProofVerification()`
- for an incoming cert to be added to the mempool, we do this check directly in `AcceptCertificateToMemoryPool()`, as for this check we need a valid view to extract the sidechain from `CSidechainMap`, given its sidechain id; so although this check being theoretically pure semantic, we cannot move it to `Sidechain::checkTxSemanticValidity()`

src/main.cpp
a-petrini: See previous answer

src/rpc/mining.cpp
a-petrini: Agree, it could be quite useful. Addressed in e6db47d33

src/sc/sidechainTxsCommitmentBuilder.cpp
a-petrini: Yes, the `if` condition was not correct. Addressed in e6db47d33. 
@JackPiri I prefer to state it in this way:
`if ((cbg.cbgaMap.count(ccout.GetScId()) == 0) && ((cbg.cbgaMap.size() + cbg.cbgcMap.size()) == cbg.SC_LIMIT))`
which in my opinion is more clear.

src/sc/sidechainTxsCommitmentBuilder.cpp
a-petrini: Corrected in e6db47d33

src/primitives/block.h
a-petrini: Reworded in e6db47d33

src/miner.cpp
a-petrini: Reworded in e6db47d33

src/miner.cpp
a-petrini: Yes, addressed in e6db47d33

src/sc/sidechainTxsCommitmentBuilder.h
a-petrini: I totally agree with this one, as I am aware that manually keep all the values aligned is error-prone.
However I think that these symbols are not exposed in zendoo_CCTPlib.a, hence we need to modify the behaviour of that lib.
Agree on moving this task to another PR.

Also, e6db47d33 changed variable declarations to `static`

qa/rpc-tests/sc_big_commitment_tree.py
a-petrini: Yes, addressed in e6db47d33

src/main.cpp
a-petrini: 1. See [this answer](https://github.com/HorizenOfficial/zen/pull/515#discussion_r1121465652)
2. Yes, agree. Changed back to `DoS` in e6db47d33

src/gtest/test_libzendoo.cpp
a-petrini: Yes, corrected in e6db47d33

qa/rpc-tests/sc_big_commitment_tree.py
a-petrini: Corrected in e6db47d33001

src/rpc/mining.cpp
ptagl: Thanks! Sorry for nitpicking here, but given that we may want in the future to completely get rid of all the asserts in the code, what do you think about keeping the call to `BuildScTxsCommitment()` outside of the `assert` and just checking the return code?

src/sc/sidechainTxsCommitmentBuilder.cpp
ptagl: Thanks, the current checks seem to be working properly now.

I consider this comment as resolved, I just wanted to propose an additional improvement: given that this check is performed at least 3 or 4 times in different parts of the commitment builder, can we encapsulate it somehow in the class (for instance having functions like `ceasedTreeContains()`, `aliveTreeContains()`, `totalSize()`, `isFull()`, etc.)?

JackPiri: I consider this comment as resolved.

JackPiri (approved):
PR approved, thanks.

src/miner.cpp
```diff
@@ -809,9 +871,16 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn,  unsigned int nBlo
         // Fill in header
         pblock->hashPrevBlock  = pindexPrev->GetBlockHash();
 
-        if (pblock->nVersion == BLOCK_VERSION_SC_SUPPORT )
+        if (pblock->nVersion == BLOCK_VERSION_SC_SUPPORT)
         {
-            pblock->hashScTxsCommitment = pblock->BuildScTxsCommitment(view);
+            pblock->hashScTxsCommitment.SetNull();
+            bool retValtxsComm = pblock->BuildScTxsCommitment(view, pblock->hashScTxsCommitment);
```
ptagl: Looking at the call to `BuildScTxsCommitment()`, would it make sense to improve encapsulation by making the function immediately update the CBlock member `hashScTxsCommitment` without passing it as a parameter?

src/sc/sidechainTxsCommitmentBuilder.cpp
```diff
@@ -161,6 +161,14 @@ bool SidechainTxsCommitmentBuilder::add_fwt(const CTxForwardTransferOut& ccout,
...
     );
 
+    if (ret) {
+    LogPrint("sc", "%s():%d - scTxsCommitment builder: successfully added FT for sidechain scId[%s].\n",
+            __func__, __LINE__, ccout.GetScId().ToString());
+    } else {
+        LogPrint("sc", "%s():%d - scTxsCommitment building failed when adding FT for sidechain scId[%s] (CCTPlib error).\n",
+            __func__, __LINE__, ccout.GetScId().ToString());
+    }
+
```
ptagl: We saw in the past some performance issues related to the printing of debug logs when dealing with the commitment tree. This is mainly because additions to the tree can happen quite frequently.

What do you think about printing the line only in case of error and doing nothing if the operation succeeds?

ptagl (changes requested):

src/miner.cpp
a-petrini: Agree. `BuildScTxsCommitment` is now `UpdateScTxsCommitment` which directly modifies `hashScTxsCommitment` member variable

src/sc/sidechainTxsCommitmentBuilder.cpp
a-petrini: Agree. Removed from both `SidechainTxsCommitmentBuilder` and `SidechainTxsCommitmentGuard`

src/rpc/mining.cpp
```diff
@@ -731,7 +731,10 @@ UniValue getblocktemplate(const UniValue& params, bool fHelp)
 
         if (certSupported) {
             CCoinsViewCache view(pcoinsTip);
-            pblock->hashScTxsCommitment = pblock->BuildScTxsCommitment(view);
+            pblock->hashScTxsCommitment.SetNull();
```
JackPiri: Is it ok no call to `BuildScTxsCommitmentGuard` is performed here, compared for example to analogous code in `getblockmerkleroots`?

src/miner.cpp
```diff
@@ -707,6 +727,48 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn,  unsigned int nBlo
...
+                bool scCommitmentBuilderResult;
+                if (tx.IsCertificate()) {
+                    scCommitmentBuilderResult = scCommBuilder->add(dynamic_cast<const CScCertificate&>(tx), view);
+                }
+                else {
+                    scCommitmentBuilderResult = scCommBuilder->add(dynamic_cast<const CTransaction&>(tx));
+                }
+                if (!scCommitmentBuilderResult)
+                {
+                    LogPrint("sc", "%s():%d - Skipping [%s] because we cannot add that to the sc commitment tree\n",
```
JackPiri, line 757: I'd say also `scCommGuard` has to rebuilt (or its counters be updated) in case there is a failure on `scCommBuilder`.

src/rpc/mining.cpp
a-petrini: Yes, it is safe, as we are building a commitment out a set of tx/cert that we know are safe, as they have been previously chosen by `CreateNewBlockWithKey()`, where a similar - actually more in-depth - check has been performed.

src/sc/sidechainTxsCommitmentGuard.cpp
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    CommitmentBuilderStatsAliveCounter& counterRef = cbs.cbsaMap[ccout.GetScId()];
+
+    // Check against the number of BWT already present in the sc subtree.
+    // Only try to add if under the limit; increase the counter only if add was successful,
+    // as adding may fail for reasons different than having a full subtree.
+    if (counterRef.bwtr < cbs.BWTR_LIMIT) {
+        ++counterRef.bwtr;
+        LogPrint("sc", "%s():%d - scTxsCommitment guard: successfully added BWTR for sidechain scId[%s].\n",
```
JackPiri: (very minor importance) Maybe this info can be added  to the log: `- current size: %d.\n",`; at least to get aligned with `add_fwt`. Similar comment applies for `add_csw` and `add_cert`.

src/gtest/test_libzendoo.cpp
```diff
@@ -2491,3 +2501,578 @@ TEST(CctpLibrary, TestCustomFieldsValidationFix)
...
+    CTransaction tx(mtx);
+    CTransaction tx2(mtx2);
+    ASSERT_TRUE(guardObj.add(tx, true));
+    // Transaction must be rejected and cbs state should be restored
+    ASSERT_FALSE(guardObj.add(tx2, true));
+
+    // Guard object must not contain anything about sidechainId2
+    ASSERT_EQ(guardObj.getCBS().cbsaMap.size(), 1);
+    ASSERT_EQ(guardObj.getCBS().cbsaMap.count(sidechainId),  1);
+    ASSERT_EQ(guardObj.getCBS().cbsaMap.count(sidechainId2), 0);
```
JackPiri, line 3118: I'd insert just this additional check to ensure internal counters have been rewound correctly:

```
auto cbsRef  = guardObj.getCBS();
ASSERT_EQ(cbsRef.cbsaMap[sidechainId].ft, 10);
```

src/sc/sidechainTxsCommitmentGuard.cpp
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    // Restore CBS to a valid state if we were not able to add any of the FT / BWTR / CSW
+    if (!addOK && autoRewind) {
+        rewind(tx, addedFt, addedBwtr, addedCsw);
+        
+        // Remove from the maps all the sidechains without entities
+        std::vector<uint256> aliveToBeRemoved;
+        for (auto [scid, cbsa]: cbs.cbsaMap) {
+            if ((cbsa.bwt == 0) && (cbsa.bwtr == 0) && (cbsa.cert == 0) && (cbsa.ft == 0))
```
JackPiri: (very minor importance) Just a matter a taste, would you consider this:

```
for (auto it = cbs.cbsaMap.cbegin(); it != cbs.cbsaMap.cend(); )
{
    if ((it->second.bwt == 0) && (it->second.bwtr == 0) && (it->second.cert == 0) && (it->second.ft == 0))
    {
        cbs.cbsaMap.erase(it++);
    }
    else
        ++it;
}
```

in order to avoid creating a vector for keeping track of the items to be removed?

Feel free to ignore if you prefer current style.

qa/rpc-tests/sc_big_commitment_tree_getblockmerkleroot.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def run_test(self):
+        '''
+        1. Mine blocks until reaching the sidechain fork; initialize tx_data with valid transaction data
+        2. Create a raw tx
+        3. Generate merkleTree and scTxsCommitmentTree for (FT_LIMIT - 1), FT_LIMIT and (FT_LIMIT + 1) txs (current FT_LIMIT is 4095)
+        4. Check if the ScTxsCommitmentTree stops updating due to limited height
+        '''
+
+        # Reach SC v2 fork height
```
JackPiri: I guess this comment should be `# Reach SC fork height`.

src/sc/sidechainTxsCommitmentGuard.cpp
a-petrini: Thanks, but as suggested by @ptagl  we removed all the logs concerning successful adds to the maps

JackPiri (commented):
The PR looks very good and clean to me; I'm going to approve but please resolve this comment https://github.com/HorizenOfficial/zen/pull/515/files#r1124332139 before.
Thanks

src/sc/sidechainTxsCommitmentGuard.cpp
a-petrini: I reimplemented this section in a very similar way to avoid the allocation of the vectors.

src/miner.cpp
a-petrini: Yes, you are right! I modified this section to keep them aligned.

qa/rpc-tests/sc_big_commitment_tree_getblockmerkleroot.py
a-petrini: Yes, corrected.

src/gtest/test_libzendoo.cpp
a-petrini: Check added.

src/sc/sidechainTxsCommitmentGuard.h
```diff
@@ -0,0 +1,90 @@
...
+    struct CommitmentBuilderStats {
+        std::map<uint256, CommitmentBuilderStatsAliveCounter>  cbsaMap;
+        std::map<uint256, CommitmentBuilderStatsCeasedCounter> cbscMap;
+        // The following values MUST be aligned with those specified in CCTPlib!
+        static const int   SC_LIMIT = 4096;
+        static const int   FT_LIMIT = 4095;
+        static const int BWTR_LIMIT = 4095;
+        static const int CERT_LIMIT = 4095;
+        static const int  CSW_LIMIT = 4095;
+        static const int  BWT_LIMIT = 4096;
```
i-Alex: I think we need to keep BWT_LIMIT = 4095 as well because of the way how it is handled by CCTP lib (limit errors are ignored there)
Beside that it seems the limit is considered as a total number of BWT per sidechain, but must be per certificate.
For example, cert A (sc1, bt_number = 3000), cert B(sc1, bt_number 2000) are allowed to be put inside the same block. It is not an issue for CommTree.

JackPiri (approved):
PR approved, thanks.

src/sc/sidechainTxsCommitmentGuard.h
```diff
@@ -0,0 +1,94 @@
...
+        uint32_t  bwt = 0;
+    };
+    struct CommitmentBuilderStatsCeasedCounter {
+        uint32_t  csw = 0;
+    };
+    struct CommitmentBuilderStats {
+        std::map<uint256, CommitmentBuilderStatsAliveCounter>  cbsaMap;
+        std::map<uint256, CommitmentBuilderStatsCeasedCounter> cbscMap;
+        // The following values MUST be aligned with those specified in CCTPlib!
+        static const int   SC_LIMIT = 4096;
```
i-Alex, line 52: Just a note: I've carefully double-checked the value for this limit, 4096 seems safe.
But in general, can we get all these constants from `zendoo-mc-cryptolib`?

i-Alex (changes requested):

src/sc/sidechainTxsCommitmentGuard.h
a-petrini: Yes, you are right. I changed the logic so that we check the limit per certificate. Also, limit has been decreased as suggested.

src/sc/sidechainTxsCommitmentGuard.h
a-petrini: Thanks for checking! Getting the limits directly from mc-crypto or cctplib was the original idea, but unfortunately they require a rework on both libraries: their symbols are defined in cctplib, but not exposed to mc-crypto.
We plan to do this in another PR, however, to simplify the alignment of these limits.

i-Alex (approved):
LGTM!

ptagl (approved):

PR #517, opened 2023-03-08
Ap/codacy ci intergration

a-petrini:
This PR includes all the code modification to support Codacy static code analysis and code coverage in our CI system.
It includes:
- new / updated Travis tasks in `.travis.yml`
- update of code coverage generation / filtering in the build system
- support for code coverage in test scripts
- updated docker files


contrib/ci-horizen/scripts/test/clang-tidy-launcher.sh
```diff
@@ -0,0 +1,45 @@
...
+# Get the current commit hash
+export COMMIT=`git log -1 --format="%H"`
+# Get the names of the files touched by this commit
+export FILELIST=`git diff-tree --no-commit-id --name-only -r $COMMIT`
+
+# Filter out all files which do not end with .cpp, .h and .cc
+FILTEREDFILELIST=()
+for i in $FILELIST;
+do
+    if [[ $i =~ .*\.cpp$ ]] || [[ $i =~ .*\.h$ ]] || [[ $i =~ .*\.tcc$ ]];
```
JackPiri: Is `[[ $i =~ .*\.tcc$ ]]` a typo for `[[ $i =~ .*\.cc$ ]]` (based on above comment)? Or do we want to include both `.cc` and `.tcc` files (in this case both the comment and the filter need to be updated)?

contrib/ci-horizen/scripts/test/clang-tidy-launcher.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+
+# Get the current commit hash
+export COMMIT=`git log -1 --format="%H"`
+# Get the names of the files touched by this commit
+export FILELIST=`git diff-tree --no-commit-id --name-only -r $COMMIT`
+
+# Filter out all files which do not end with .cpp, .h and .cc
```
JackPiri: Probably we should include also `.hpp` files, being the current extension of some headers.

contrib/ci-horizen/scripts/test/clang-tidy-launcher.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+
+# Get the current commit hash
+export COMMIT=`git log -1 --format="%H"`
```
JackPiri: Some considerations can be made on extracting all the commits with respect to remote branch and then iterating the clan-tidy analysis on each of them; for sure this would require more time for the analysis but the result would be more appropriate, targeting all the commits and not just the last one.

Makefile.am
```diff
@@ -161,6 +161,10 @@ baseline_filtered.info: baseline.info
                       "$(abs_builddir)/depends/x86_64-unknown-linux-gnu/include/boost/*" \
                       "$(abs_builddir)/depends/x86_64-unknown-linux-gnu/include/gmock/*" \
                       "$(abs_builddir)/depends/x86_64-unknown-linux-gnu/include/gtest/*" \
+                      "$(abs_builddir)/depends/x86_64-linux-gnu/include/*.h" \
```
drgora, line 164: I'd suggest the following to get rid of some duplication:

```suggestion
                      "$(abs_builddir)/depends/$(host_triplet)/include/*.h" \
```

contrib/ci-horizen/scripts/test/clang-tidy-launcher.sh
a-petrini: I actually wanted to include only the `*.tcc` files, as all the `*.cc` included in the repository belong to leveldb. Instead, I would add the filter for `*.hpp` files, as you stated in another comment.

contrib/ci-horizen/scripts/test/clang-tidy-launcher.sh
a-petrini: Yes, agree!

contrib/ci-horizen/scripts/test/clang-tidy-launcher.sh
a-petrini: Yes, you are right.
However, keeping track of which files have been modified in all the commits is not very convenient and may break in case of force push. Instead, I am experimenting with `run-clang-tidy.py` [works in parallel] to analyze all the codebase, and modify this script accordingly.

Makefile.am
```diff
@@ -246,6 +262,16 @@ total.coverage/.dirstamp: total_coverage.info
...
+COMMIT = `git log -1 --format="%H"`
+
+cov_ci: test_bitcoin.coverage/.dirstamp cov-zen total.coverage/.dirstamp
+	bash <(curl -Ls https://coverage.codacy.com/get.sh) report --partial -l CPP --commit-uuid ${COMMIT} \
+    --api-token ${CODACY_API_TOKEN_COVERAGE} --organization-provider "gh" \
+    --username "HorizenOfficial" --project-name "zen" -r total_coverage.info 
+	bash <(curl -Ls https://coverage.codacy.com/get.sh) report --partial -l C   --commit-uuid ${COMMIT} \
+    --api-token ${CODACY_API_TOKEN_COVERAGE} --organization-provider "gh" \
+    --username "HorizenOfficial" --project-name "zen" -r total_coverage.info
+
```
drgora: I'd suggest the following to get rid of some duplication

```suggestion
cov_ci_%:
      bash <(curl -Ls https://coverage.codacy.com/get.sh) report --partial -l $* --commit-uuid ${COMMIT} \
    --api-token ${CODACY_API_TOKEN_COVERAGE} --organization-provider "gh" \
    --username "HorizenOfficial" --project-name "zen" -r total_coverage.info 

cov_ci: test_bitcoin.coverage/.dirstamp cov-zen total.coverage/.dirstamp
	cov_ci_CPP
	cov_ci_C
```

src/Makefile.am
```diff
@@ -363,6 +363,7 @@ crypto_libbitcoin_crypto_a_SOURCES = \
   crypto/sha256.h \
   crypto/sha512.cpp \
   crypto/sha512.h
+crypto_libbitcoin_crypto_a_LDFLAGS = $(AM_CXXFLAGS) $(PIE_FLAGS) $(COVERAGE_FLAGS)
```
drgora: I'm not sure what is the idea here, but this seems weird to me. Given the file name ending with `a`, I guess we are building a static library here. This means that there should be no linker involved at all. So, why are we adding `LDFLAGS`?

Makefile.am
a-petrini: agree, modified!

src/Makefile.am
a-petrini: Yes, you are right. I revised the entire file and left `$(COVERAGE_FLAGS)` only when it is really needed.

contrib/ci-horizen/scripts/test/clang-tidy-launcher.sh
a-petrini: The whole section has been rewritten to take into account the history of non-analyzed commits. Now the script `clang-tidy_getFileList.py` scans the branch history and considers all the code files in the commits between the last one submitted and the last one analyzed by Codacy in the branch history.
Also, clang-tidy now runs in parallel to further reduce analysis time.

Makefile.am
a-petrini: Thanks, I saw it has been defined along with `build_triplet`, but it is never used elsewhere. Also, I would like tokeep the same `lcov` filters that have been defined in `rpc-tests.sh`.

Makefile.am
a-petrini: Actually, this has been reverted to the original version

contrib/ci-horizen/scripts/test/clang-tidy-launcher.sh
a-petrini: the last comments have been outdated by the last version, as the CI now examines all the files in a commit. It still may be necessary to manually trigger commit analysis on the Codacy dashboard when submitting more than a commit at once.

contrib/ci-horizen/scripts/build/build_amd64_linux_ubuntu_jammy.sh
```diff
@@ -5,4 +5,4 @@ set -euo pipefail
 docker run --rm -v "${TRAVIS_BUILD_DIR}:/mnt/build" -v "$HOME/.ccache:/mnt/.ccache" -e LOCAL_USER_ID="$(id -u)" \
   -e LOCAL_GRP_ID="$(id -g)" --env-file <(env | grep 'DOCKER_\|MAKEFLAGS') "${IMAGE_NAME}:${IMAGE_TAG}" \
   bash -c 'set -xeuo pipefail && export HOST="$(gcc -dumpmachine)" && export MAKEFLAGS="${MAKEFLAGS:-} -j $(($(nproc)+1))" \
-    && cd "${DOCKER_HOME}" && /usr/bin/time -v ./zcutil/build.sh $MAKEFLAGS'
+    && cd "${DOCKER_HOME}" && bear -- ./zcutil/build.sh $MAKEFLAGS'
```
ptagl: A couple of questions here:
- Can we keep `/usr/bin/time -v` before `bear` in order to have statistics about the build process?
- What is the purpose of `--`? I couldn't find any mention in the help.

a-petrini: For the first remark, I updated the script as suggested.
As for the `--` option of `bear`, it is required for versions > 2.4, as those installed by default on Ubuntu Jammy through `apt` package manager (see https://github.com/rizsotto/Bear#how-to-use )

ptagl: I was using version 2.4, that's why I couldn't find it. Thanks a lot!

ptagl (approved):

JackPiri (approved):
The PR is fine to me.
Thanks

drgora (approved):

PR #518, opened 2023-03-14
[CI] Always run MacOS jobs when building a tag

ptagl:
After some recent adjustments to Travis CI, we started skipping MacOS jobs by default on any branch with the exception of:
- `master`
- `development`

As an unwanted side effect, we also started skipping MacOS when pushing tags, and this is an issue as we typically tag new releases and need all the binaries to be generated.

This PR adds an exception to the `travis.yml` file so that MacOS jobs are always included when building a tag.

I've also included a small refactoring to make (hopefully) some `if` statements more readable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -294,7 +294,7 @@ os: linux
       - bash -c "${TRAVIS_BUILD_DIR}/contrib/ci-horizen/scripts/common/travis_stay_alive.sh"
       - bash -c '${TRAVIS_BUILD_DIR}/contrib/ci-horizen/scripts/test/run_test.sh "${TEST_CMD}" "${TEST_ARGS}"'
   - &test_osx_xcode9_4
-    if: (NOT env(SKIP_TESTS)) AND (NOT env(SKIP_OSX)) OR tag IS present # Never skip MacOS when running the CI for a tag
+    if: NOT env(SKIP_TESTS) AND NOT env(SKIP_OSX) OR tag IS present # Never skip MacOS when running the CI for a tag
     os: osx
     osx_image: xcode9.4
     cache:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -432,50 +432,50 @@ jobs:
...
       <<: *test_osx_xcode9_4
       <<: *amd64_osx_xcode9_4_rpc-tests_9
     - stage: Test
-      if: env(SKIP_OSX) != true
+      if: NOT env(SKIP_OSX) OR tag IS present
       <<: *test_osx_xcode9_4
       <<: *amd64_osx_xcode9_4_rpc-tests_10
     - stage: Test
-      if: env(SKIP_OSX) != true
+      if: NOT env(SKIP_OSX) OR tag IS present
```
cronicc: All of these `if: NOT env(SKIP_OSX) OR tag IS present` lines should not be necessary, as the `<<: *test_osx_xcode9_4` yml anchor pulls in line 297 `if: (NOT env(SKIP_TESTS)) AND (NOT env(SKIP_OSX)) OR tag IS present # Never skip MacOS when running the CI for a tag`

cronicc (changes requested):

.travis.yml
ptagl: I've just removed the redundant checks as you suggested, thanks!

cronicc (approved):

.travis.yml
```diff
@@ -294,7 +294,7 @@ os: linux
       - bash -c "${TRAVIS_BUILD_DIR}/contrib/ci-horizen/scripts/common/travis_stay_alive.sh"
       - bash -c '${TRAVIS_BUILD_DIR}/contrib/ci-horizen/scripts/test/run_test.sh "${TEST_CMD}" "${TEST_ARGS}"'
   - &test_osx_xcode9_4
-    if: (env(SKIP_TESTS) != true) AND (env(SKIP_OSX) != true)
+    if: (NOT env(SKIP_TESTS)) AND (NOT env(SKIP_OSX)) OR tag IS present # Never skip MacOS when running the CI for a tag
```
JackPiri: As for C/C++, I believe `OR` operator has lower precedence than `AND` operator, but if that's not the case we should modify with:

`if: ((NOT env(SKIP_TESTS)) AND (NOT env(SKIP_OSX))) OR (tag IS present) # Never skip MacOS when running the CI for a tag`

otherwise just skip this comment.

ptagl: Looking at the [documentation](https://docs.travis-ci.com/user/conditions-v1#boolean-operators) it seems like Travis has standard precedence rules, so I've just completely removed the redundant square brackets.

JackPiri (approved):
The PR is fine to me, thanks.

PR #519, opened 2023-03-17
[Python Tests] Refactoring of BlockchainHelper

ptagl:
This PR includes a small refactoring of the BlockchainHelper class, in particular of the `create_sidechain()` method, just to remove the explicit `version` input parameter that was redundant as already available in the `creation_arguments` input object.

Additional changes:
- Fixed an issue in `_generate_params()` (`mc_test.py`), which was returning `None` when called more than once with the same arguments
- Added two missing Python tests to the CI


qa/rpc-tests/test_framework/blockchainhelper.py
```diff
@@ -165,15 +179,19 @@ def init_proof_keys(self, bitcoin_test_framework):
...
+        Creates a sidechain with the given name
+        (used to store it locally in a map) and
+        the creation arguments provided.
+        Note that among these arguments, only
+        "version" is mandatory.
         """
 
-        sc_input = SidechainCreationInput.from_rpc_args(self, sc_name, sidechain_version, optional_parameters)
+        # "version" is a mandatory argument
+        assert("version" in creation_arguments)
```
a-petrini: This check is redundant, as it is performed also inside `SidechainCreationInput.from_rpc_args()`

qa/rpc-tests/test_framework/blockchainhelper.py
```diff
@@ -27,6 +27,16 @@
...
 MODULUS_BITS = 255
 
+"""
+Dictionary of default parameters that can be used when calling 'create_sidechain'.
+"""
+SidechainParameters = {
+    "DEFAULT_SC_V0":                { "version": 0 },
+    "DEFAULT_SC_V1":                { "version": 1 },
+    "DEFAULT_SC_V2_CEASABLE":       { "version": 2, "withdrawalEpochLength": 10 },
+    "DEFAULT_SC_V2_NON_CEASABLE":   { "version": 2, "withdrawalEpochLength": 0 }
```
JackPiri: Could be worth to add `"wCeasedVk": ""` for non-ceasable sidechains?

JackPiri (approved):
The PR is fine to me

a-petrini (changes requested):

drgora (approved):
Looks good to me

qa/rpc-tests/test_framework/blockchainhelper.py
a-petrini: Fix looks good to me!

a-petrini (approved):

qa/rpc-tests/test_framework/blockchainhelper.py
ptagl: I like this suggestion as it also avoids having an ugly check inside `from_rpc_args()` to eventually unset `wCeasedVk`.
Fixed in the latest commit, thanks!

qa/rpc-tests/test_framework/blockchainhelper.py
ptagl: I agree, thanks for spotting, fixed in the latest commit.

PR #524, opened 2023-03-22
Adding experimental load balancer to CI Python test suite

a-petrini:
This PR adds a load balancer that redistributes the Python test workload of the parallel workers in a Travis CI task.
The algorithm is a simple Longest-processing-time-first (LPT).
rpc-tests.sh has been modified to use the results provided by the load balancer.
Execution times have been gathered from an actual test run on Travis under Ubuntu Jammy.

### DIFF
--- a/qa/pull-tester/loadbalancer.py
+++ b/qa/pull-tester/loadbalancer.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+import sys
+
+workers = int(sys.argv[1])
+selectedChunk = int(sys.argv[2])        # bash style, they go from 1 to workers
+
+# Restrict selectedChunk value in [1,workers] (no silent fail)
+selectedChunk = min(selectedChunk, workers)
+selectedChunk = max(selectedChunk, 1)
+
+## STEP-BY-STEP example
+## sys.argv:
+## ['prova1.py', '4', '1', 'paymentdisclosure.py,63', 'txn_doublespend.py,15', 'txn_doublespend.py', '--mineblock,16', 'getchaintips.py,48', 'rawtransactions.py,101']
+
+## join the parameter list and get a single string
+## newstr: paymentdisclosure.py,63 txn_doublespend.py,15 txn_doublespend.py --mineblock,16 getchaintips.py,48 rawtransactions.py,101
+newstr = ' '.join(sys.argv[3:])
+
+## divide the string, using comma (',') as separator
+## split1str: ['paymentdisclosure.py', '63 txn_doublespend.py', '15 txn_doublespend.py --mineblock', '16 getchaintips.py', '48 rawtransactions.py', '101']
+split1str = newstr.split(',')
+
+## A tuple is made by all the chars following the number of the n-th element in split1str, and the number of the n+1-th element
+## tuplelist: [('paymentdisclosure.py', 63), ('txn_doublespend.py', 15), ('txn_doublespend.py --mineblock', 16), ('getchaintips.py', 48), ('rawtransactions.py', 101)]
+tuplelist = [(split1str[0], int(split1str[1].split(' ')[0]))]
+for i in range(1, len(split1str)-1):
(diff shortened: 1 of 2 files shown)


qa/pull-tester/loadbalancer.py
```diff
@@ -0,0 +1,52 @@
...
+workers = int(sys.argv[1])
+selectedChunk = int(sys.argv[2])        # bash style, they go from 1 to workers
+
+# Restrict selectedChunk value in [1,workers] (no silent fail)
+selectedChunk = min(selectedChunk, workers)
+selectedChunk = max(selectedChunk, 1)
+
+## STEP-BY-STEP example
+## sys.argv:
+## ['prova1.py', '4', '1', 'paymentdisclosure.py,63', 'txn_doublespend.py,15', 'txn_doublespend.py', '--mineblock,16', 'getchaintips.py,48', 'rawtransactions.py,101']
```
JackPiri: Possibly `'prova1.py'` in this comment should be updated

qa/pull-tester/rpc-tests.sh
```diff
@@ -34,163 +34,163 @@ done
...
-  'txindex.py'
-  'getblockexpanded.py'
-  'sc_rpc_cmds_json_output.py'
-  'sc_version.py'
-  'sc_getscgenesisinfo.py'
-  'fundaddresses.py'
-  'sc_getcertmaturityinfo.py'
-  'sc_big_commitment_tree.py'
-  'sc_big_commitment_tree_getblockmerkleroot.py'
+  'paymentdisclosure.py',99
```
JackPiri, line 40: What about using a more "complex" separator (possibly using 2 or 3 characters) rather than a single comma?
Actually I can see current separator problematic only in case of test file name including comma (I would say probability 0.000...0001, but you never know), or maybe some other edge cases not being considered right now.
In case, a consequent update needs to be made also on Python side.

JackPiri (approved):
Very well done, approved.
Thanks

qa/pull-tester/rpc-tests.sh
a-petrini: I agree, however to avoid introducing more complexity to the scripts, at the moment I added a comment ot `rpc-tests.sh` to remind of this.

qa/pull-tester/loadbalancer.py
a-petrini: Nice! Yes, corrected with the actual filename

qa/pull-tester/rpc-tests.sh
```diff
@@ -213,20 +213,14 @@ fi
...
-    testScripts=( "${testScripts[@]:$start:$chunkSize}" )
-  fi
+else
+  chunks='1'
+  chunk='1'
 fi
+# Temporary replace the array delimiter from ' ' to '|'
+originalIFS="$IFS"; IFS='|'
+read -a testScripts <<< $(${BUILDDIR}/qa/pull-tester/loadbalancer.py $chunks $chunk ${testScripts[@]})
+IFS="$originalIFS"
```
drgora: Just throwing an idea here, feel free to ignore it. With something like the following, you should have a much easier to handle list of arguments passed to your Python script. You might be able to simplify things a bit, especially lines 15-29 of `loadbalancer.py`:
```suggestion
read -a testScripts <<< $(${BUILDDIR}/qa/pull-tester/loadbalancer.py $chunks $chunk $(printf '"%s" ' "${testScripts[@]}"))
```

drgora (approved):
I just left a minor comment, but definitely non blocking, so please feel free to ignore it. This already looks good to me, thanks.

qa/pull-tester/rpc-tests.sh
a-petrini: Thanks! I integrated this change and added a simplified scheme for passing parameter back and forth the Python script.

drgora: Awesome, thank you!

PR #526, opened 2023-03-23
Gp/zend 313

JackPiri:
This PR is the implementation of ZenIP-42204 "Deprecation of shielded pool".

Reference at: https://github.com/HorizenOfficial/ZenIPs (see both md file and update PR) and https://blog.horizen.io/what-is-zenip-42204-deprecation-of-shielded-pools/.

Notes for the reviewers:
- some numerical values regarding block mining used in already present python tests (wallet_mergetoaddress.py, wallet_shieldcoinbase.py) have been lowered in order to keep the regtest fork height closer to last fork height (480->990 being already a quite significant jump),
- actual fork heights for mainnet and testnet has to be defined (currently just a placeholder is provided).


src/sc/sidechainrpc.cpp
```diff
@@ -885,11 +885,13 @@ void ScRpcCmd::addInputs()
 
     static const bool fOnlyConfirmed = false;
     static const bool fIncludeZeroValue = false;
-    bool fProtectCoinbase = !Params().GetConsensus().fCoinbaseMustBeProtected;
-    static const bool fIncludeCoinBase = fProtectCoinbase;
-    static const bool fIncludeCommunityFund = fProtectCoinbase;
+    static bool fMustShieldCoinBase = ForkManager::getInstance().mustCoinBaseBeShielded(chainActive.Height());
```
drgora: This variable now depends on runtime state, so it should not be `static` (and hence initialized only once).
```suggestion
    const bool fMustShieldCoinBase = ForkManager::getInstance().mustCoinBaseBeShielded(chainActive.Height());
```

src/wallet/rpcwallet.cpp
```diff
@@ -40,17 +40,60 @@
...
-
     int GetJoinSplitSize(int shieldedTxVersion) {
         return JSDescription::getNewInstance(shieldedTxVersion == GROTH_TX_VERSION).GetSerializeSize(SER_NETWORK, PROTOCOL_VERSION, shieldedTxVersion);
+    }    
+}
+
+// anonymous namespace related to shielded pool deprecation hard fork
+// TODO: REMOVE ONCE SHIELDED POOL DEPRECATION HARD FORK IS "HIGHLY FINAL"
+namespace{
+    std::string shieldedPoolDeprecationErrorMessage = "This method is deprecated due to shielded pool deprecation hard fork.";
```
drgora: Just a small optimization, feel free to ignore this, but here you do not need to init a string.
```suggestion
    const char* shieldedPoolDeprecationErrorMessage = "This method is deprecated due to shielded pool deprecation hard fork.";
```

or even

```suggestion
    constexpr std::string_view shieldedPoolDeprecationErrorMessage = "This method is deprecated due to shielded pool deprecation hard fork.";
```

src/wallet/rpcwallet.cpp
```diff
@@ -40,17 +40,60 @@
...
+
+    //! Method for identifying if shielded pool deprecation hard fork is enabled on active chain
+    /*!
+      \return shielded pool deprecation hard fork enable
+    */
+    int GetShieldedPoolDeprecationForkHeight()
+    {
+        CBaseChainParams::Network network = Params().NetworkIDString() == "main" ? CBaseChainParams::Network::MAIN :
+                                            Params().NetworkIDString() == "test" ? CBaseChainParams::Network::TESTNET :
+                                                                                   CBaseChainParams::Network::REGTEST;
```
drgora, line 69: I understand this might be a bit off scope, and maybe we want to defer this to a later time, but I believe this might be a good chance to add a `CBaseChainParams::Network` field directly to `class CChainParams`.

This would allow us to avoid all this logic in this specific case, but also to replace many costly string comparisons that we have in our codebase, in future.

qa/rpc-tests/wallet_protectcoinbase.py
```diff
@@ -410,6 +412,78 @@ def run_test (self):
         notes = self.nodes[0].z_listunspent(1, 99999, False, [myzaddr])
         sum_of_notes = sum([note["amount"] for note in notes])
         assert_equal(Decimal(resp), sum_of_notes)
+        
+
+        # tests associated to shielded pool depraction hard fork
```
drgora: I'm wondering whether we could speed up these tests, and also test the dynamics of zend code better (see my comment on the static var in `src/sc/sidechainrpc.cpp`). Do you think it could make sense to launch 2 nodes, one with `-regtestprotectcoinbase` and one without, and perform independent tests on those so that we do not need to restart nodes several times? They should be disconnected from each other, but this should not be a problem for this test?

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -23,12 +23,12 @@
...
-    'MINIMAL_SC':     420,
-    'SC_VERSION':     450,
-    'NON_CEASING_SC': 480
+    'MINIMAL_SC':                420,
+    'SC_VERSION':                450,
+    'NON_CEASING_SC':            480,
+    'SHIELDED_POOL_DEPRECATION': 990
 }
 
-
```
drgora, line 31: Oh well, that blank line was actually intentional, and conforming to Python style [guidelines](https://peps.python.org/pep-0008/#blank-lines) :)

qa/rpc-tests/wallet_protectcoinbase.py
```diff
@@ -410,6 +412,78 @@ def run_test (self):
...
+            self.nodes[0].sendtoaddress(self.nodes[2].getnewaddress(), balance_0, "", "", True)
+        except JSONRPCException as e:
+            print("Unexpected exception caught during testing: " + str(sys.exc_info()[0]))
+            assert(False)
+        
+        self.nodes[1].generate(ForkHeights['SHIELDED_POOL_DEPRECATION'] - block_count + 1) # overcame hard fork mining blocks on another node
+        self.sync_all()
+        block_count = self.nodes[0].getblockcount()
+        assert_greater_than(block_count, ForkHeights['SHIELDED_POOL_DEPRECATION']) # otherwise following tests would make no sense
+        self.nodes[0].generate(101)
```
drgora: Do we actually need to generate these blocks? If so, it could be useful to have a brief comment

qa/rpc-tests/zcjoinsplit.py
```diff
@@ -4,53 +4,111 @@
...
+                    continue # other steps of this round are skipped
+
+            if (round == just_pre_fork_round):
+                # advance chain on node1 (so that shielded pool deprecation hard fork is active)
+                self.nodes[1].generate(20)
+                connect_nodes(self.nodes[0], 1)
+                sync_blocks(self.nodes)
+                # make sure now deprecated tx is node0 mempool
+                assert_equal(self.nodes[0].getmempoolinfo()["size"], 1)
+                # mine a block including now deprecated tx
```
drgora, line 83: It doesn't add much to the logic of the test, but asserting the `nodes[1]` mempool is empty might make it more immediate to understand for the reader.

drgora (changes requested):

src/sc/sidechainrpc.cpp
JackPiri: Totally agree, that was a bug. Addressed in ca71777097aece83e8f83375243baeac51c0451d.

src/wallet/rpcwallet.cpp
JackPiri: I'd go with first option which doesn't require additional modifications; addressed in ca71777097aece83e8f83375243baeac51c0451d.

src/wallet/rpcwallet.cpp
JackPiri: I am of the same opinion.
This point has been already discussed with other team members, it was agreed to track this and refactor in a separate PR (if I remember correctly, also other parts of the codebase would benefit from this refactoring).

qa/rpc-tests/wallet_protectcoinbase.py
JackPiri: At first glance I was going to agree with your proposal, but in the end I prefer keeping nodes connected; it seems to me tests are more robust and reliable when mempool/block validation is performed not only on self-generated data but also on received data. The overall execution time is surely a little bit longer but this does not represent a significant difference.

qa/rpc-tests/test_framework/test_framework.py
JackPiri: Ok, reverted in ca71777097aece83e8f83375243baeac51c0451d.

qa/rpc-tests/wallet_protectcoinbase.py
JackPiri: Yes, we need them to overcome the hard fork.
In ca71777097aece83e8f83375243baeac51c0451d I moved comment `# overcome hard fork mining blocks on another node` from line 451 on previous line in order to make it more visible.

qa/rpc-tests/zcjoinsplit.py
JackPiri: I like your proposal, added assertion also on `nodes[1]` side in ca71777097aece83e8f83375243baeac51c0451d.

src/zen/forks/fork0_originalfork.h
```diff
@@ -98,6 +98,16 @@ class OriginalFork : public Fork
...
+
+    /**
+     * @brief returns true if the coin base transactions must be shielded (i.e. sent to a z-address)
+     */
+    virtual bool mustCoinBaseBeShielded(CBaseChainParams::Network network) const;
+
+    /**
+     * @brief returns true if the shielding (t->z) transactions are forbidden
+     */
+    virtual bool isShieldingForbidden() const { return false; };
```
a-petrini: This should be `inline virtual`, according to the other functions declared and defined in this header file.

src/zen/forks/fork11_shieldedpooldeprecationfork.h
```diff
@@ -0,0 +1,25 @@
...
+
+    /**
+     * @brief returns true if the coin base transactions must be shielded (i.e. sent to a z-address)
+     */
+    inline virtual bool mustCoinBaseBeShielded(CBaseChainParams::Network network) const { return false; }
+
+    /**
+     * @brief returns true if the shielding (t->z) transactions are forbidden
+     */
+    virtual bool isShieldingForbidden() const { return true; };
```
a-petrini: This should be `inline virtual`.

qa/rpc-tests/shieldedpooldeprecation_rpc.py
```diff
@@ -0,0 +1,198 @@
...
+            try:
+                # merging shielded funds to z-address (unshielding)
+                opid = self.nodes[0].z_mergetoaddress(["ANY_ZADDR"], node0_taddr0, 0.0001, 2, 2)["opid"]
+                wait_and_assert_operationid_status(self.nodes[0], opid)
+                self.make_all_mature()
+            except JSONRPCException as e:
+                    print("Unexpected exception caught during testing: " + str(sys.exc_info()[0]))
+                    assert(False)
+
+            # z_sendmany
```
a-petrini, line 160: Totally non-blocking comment, as it may be redundant with all the extensive test cases already provided.
As `z_sendmany` takes an array of outputs, does it fail appropriately when mixing tAddresses and zAddresses after the fork?

src/wallet/rpcwallet.cpp
```diff
@@ -6001,7 +6057,9 @@ UniValue z_mergetoaddress(const UniValue& params, bool fHelp)
     if (fHelp || params.size() < 2 || params.size() > 6)
         throw runtime_error(
             "z_mergetoaddress [\"fromaddress\", ... ] \"toaddress\" ( fee ) ( transparent_limit ) ( shielded_limit ) ( memo )\n"
-            + strDisabledMsg +
+            + strDisabledMsg
+            + ShieldedPoolDeprecationWarning(false) + "\n"
+            + "Details: merging transparent funds to shielded address " + (IsShieldedPoolDeprecationForkEnabled() ? "is going to be " : "has been ") + "deprecated.\n"
             "\nMerge multiple UTXOs and notes into a single UTXO or note.  Coinbase UTXOs are ignored; use `z_shieldcoinbase`"
```
a-petrini: Maybe we should update this comment, as `z_shieldcoinbase` is going to be deprecated as well

qa/pull-tester/rpc-tests.sh
```diff
@@ -143,6 +143,7 @@ testScripts=(
   'sc_getcertmaturityinfo.py'
   'sc_big_commitment_tree.py'
   'sc_big_commitment_tree_getblockmerkleroot.py'
+  'shieldedpooldeprecation_rpc.py'
```
a-petrini: With the new CI workload rebalancer we need to specify the execution time  for each new `.py` we add to this list.
To have a rough estimate of the execution time on the CI, I suggest to measure it locally, then multiply by 1.5 for the Linux and 2.0 for MacOS. Alternatively, any placeholder value would suit well, at least for the first few runs.
Suggested change:
`'shieldedpooldeprecation_rpc.py',450,600`

a-petrini (changes requested):

src/zen/forks/fork0_originalfork.h
JackPiri: Agree, modified accordingly.

src/zen/forks/fork11_shieldedpooldeprecationfork.h
JackPiri: Agree, modified accordingly.

qa/rpc-tests/shieldedpooldeprecation_rpc.py
JackPiri: Current check on `z_sendmany` deprecation doesn't consider t-addresses recipients. Btw it seems reasonable to me to add this test case for making the outcome more robust; modified accordingly.

src/wallet/rpcwallet.cpp
JackPiri: Help documentation slighlty refactored.

qa/pull-tester/rpc-tests.sh
JackPiri: Agree, modified with slight increase of test time `'shieldedpooldeprecation_rpc.py',500,700` (due to additional test steps).

drgora (approved):
This is ok for me, thanks

a-petrini (approved):

qa/rpc-tests/shieldedpooldeprecation_rpc.py
```diff
@@ -0,0 +1,225 @@
...
+
+class ShieldedPoolDeprecationTest (BitcoinTestFramework):
+
+    def setup_chain(self):
+        print("Initializing test directory "+self.options.tmpdir)
+        initialize_chain_clean(self.options.tmpdir, 2)
+
+    def setup_network(self, split=False):
+        args = ['-debug=zrpcunsafe', '-experimentalfeatures', '-zmergetoaddress']
+        self.nodes = start_nodes(2, self.options.tmpdir, [['-debug=zrpcunsafe', '-experimentalfeatures', '-zmergetoaddress']] * 2)
```
ptagl: It seems like `args` is unused, can we get rid of it?

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -23,9 +23,10 @@
...
 
 ForkHeights = {
-    'MINIMAL_SC':     420,
-    'SC_VERSION':     450,
-    'NON_CEASING_SC': 480
+    'MINIMAL_SC':                420,
+    'SC_VERSION':                450,
+    'NON_CEASING_SC':            480,
+    'SHIELDED_POOL_DEPRECATION': 990
 }
```
ptagl, line 30: Can we reduce the fork height for `SHIELDED_POOL_DEPRECATION` to reduce execution time?

I would review all regtest hard fork heights in general in the future to see if they still make sense or if we can find a better configuration.

qa/rpc-tests/shieldedpooldeprecation_rpc.py
```diff
@@ -0,0 +1,225 @@
...
+                opid = self.nodes[0].z_sendmany(node0_zaddr0, [{"address": node0_zaddr1, "amount": 1.0}, {"address": node0_taddr0, "amount": 1.0}], 1, 0)
+                wait_and_assert_operationid_status(self.nodes[0], opid)
+                self.make_all_mature()
+            except JSONRPCException as e:
+                print("Unexpected exception caught during testing: " + str(sys.exc_info()[0]))
+                assert(False)
+
+            if (round == pre_fork_round):
+                blockcount = self.nodes[0].getblockcount()
+                if (blockcount < ForkHeight):
```
ptagl: What about explicitly asserting that we are below the fork point after the first iteration? I'm considering the possibility that we change the fork height making the test inconsistent.

qa/rpc-tests/wallet_protectcoinbase.py
ptagl, line 1: I see we are missing some calls to `sync_all()` after creating transactions. I think it's not a big deal in the context of this test, but please double-check that it can't cause sporadic failures in the future (and feel free to close the comment then).

qa/rpc-tests/wallet_shieldcoinbase.py
ptagl, line 1: I see several negative checks that expect to see an error code and/or a specific error message, but the test would pass also if those checks don't fail. I guess we should include some `asserts` there to be sure.

qa/rpc-tests/wallet_protectcoinbase.py
```diff
@@ -410,6 +412,79 @@ def run_test (self):
...
+        except JSONRPCException as e:
+            print("Unexpected exception caught during testing: " + str(sys.exc_info()[0]))
+            assert(False)
+        
+        # overcome hard fork mining blocks on another node
+        self.nodes[1].generate(ForkHeights['SHIELDED_POOL_DEPRECATION'] - block_count + 1)
+        self.sync_all()
+        block_count = self.nodes[0].getblockcount()
+        assert_greater_than(block_count, ForkHeights['SHIELDED_POOL_DEPRECATION']) # otherwise following tests would make no sense
+        self.nodes[0].generate(101)
```
ptagl: Given that we already passed the fork point, can't we avoid generating 101 blocks? I'd also suggest making this check exactly when the chain height is `ForkHeight - 1` because that is the real edge case (for instance, it would fail now because on RPC commands we are checking the status of the hard fork based on `chainActive.Height()` instead of `chainActive.Height() + 1`).

qa/rpc-tests/zcjoinsplit.py
```diff
@@ -4,53 +4,112 @@
...
+                    assert_equal(e.error["code"], RPC_VERIFY_REJECTED)
+                    continue # other steps of this round are skipped
+
+            if (round == just_pre_fork_round):
+                # advance chain on node1 (so that shielded pool deprecation hard fork is active)
+                self.nodes[1].generate(20)
+                connect_nodes(self.nodes[0], 1)
+                sync_blocks(self.nodes)
+                # make sure now deprecated tx is in node0 mempool but not in node1 mempool
+                assert_equal(self.nodes[0].getmempoolinfo()["size"], 1)
```
ptagl, line 81: I wonder if we need a mechanism to remove such invalid transaction from the mempool (`removeStaleTransactions()`), let's see also how the same thing was managed when switching from PHGR to GROTH.

src/wallet/rpcwallet.cpp
```diff
@@ -6161,6 +6220,13 @@ UniValue z_mergetoaddress(const UniValue& params, bool fHelp)
         }
     }
 
+    // we want to forbid shielding transactions
+    if (chainActive.Height() >= GetShieldedPoolDeprecationForkHeight() &&
```
ptagl: Given that the RPC command is going to add a new transaction to the mempool that will be eventually mined in the **next** block, we should compare the fork height with `chainActive.Height() + 1`.

This also applies to other occurrences included in this PR.

src/wallet/rpcwallet.cpp
```diff
@@ -6187,7 +6253,14 @@ UniValue z_mergetoaddress(const UniValue& params, bool fHelp)
...
         vector<COutput> vecOutputs;
-        pwalletMain->AvailableCoins(vecOutputs, true, NULL, false, false);
+
+        bool fMustShieldCoinBase = ForkManager::getInstance().mustCoinBaseBeShielded(chainActive.Height());
+        bool fMustShieldCommunityFund = false;
+        // CF exemption allowed only after hfCommunityFundHeight hardfork
+        if (!ForkManager::getInstance().canSendCommunityFundsToTransparentAddress(chainActive.Height()))
+            fMustShieldCommunityFund = fMustShieldCoinBase;
+
+        pwalletMain->AvailableCoins(vecOutputs, true, NULL, false, !fMustShieldCoinBase, !fMustShieldCommunityFund);
```
ptagl: A few considerations here:
- we need to check the validation rules on `chainActive.Height() + 1`
- It's correct to exclude coinbase outputs if they are required to be shielded, but I guess only when merging to a transparent address, while when `isToZaddr == true` we can include it
- Can't we treat `fMustShieldCommunityFund` as an independent variable and avoid `fMustShieldCommunityFund = fMustShieldCoinBase;`?

If the comments above are right and the code has been changed, I would also evaluate changing the name of variables `fMustShieldCoinBase` and `fMustShieldCommunityFund` to something like `fIncludeCoinbase` and `fIncludeCommunityFund` if you think it improves readability.

qa/rpc-tests/shieldedpooldeprecation_rpc.py
```diff
@@ -0,0 +1,225 @@
...
+                    self.nodes[0].generate(ForkHeight - blockcount)
```
ptagl: Shouldn't we consider `ForkHeight - blockcount - 1` as proper edge case for RPC commands?

qa/rpc-tests/zcjoinsplit.py
```diff
@@ -4,53 +4,112 @@
...
+            receive_result = self.nodes[0].zcrawreceive(zcsecretkey, joinsplit_result["encryptednote1"])
+            assert_equal(receive_result["exists"], True)
 
-        self.nodes[0].sendrawtransaction(joinsplit_result["rawtxn"])
-        self.nodes[0].generate(1)
+            if (round == pre_fork_round):
+                blockcount = self.nodes[0].getblockcount()
+                if (blockcount < ForkHeight - 10):
+                    self.nodes[0].generate(ForkHeight - 10 - blockcount)
+                    self.sync_all()
```
ptagl: A couple of considerations here:
- What happens in case the `if` condition fails? Would the test still be consistent?
- To reach the edge case we should be at the block immediately before the hard fork activates, so the generation should be `self.nodes[0].generate(ForkHeight - 10 - blockcount) - 1`

src/sc/sidechainrpc.cpp
```diff
@@ -885,11 +885,13 @@ void ScRpcCmd::addInputs()
...
-    static const bool fIncludeCoinBase = fProtectCoinbase;
-    static const bool fIncludeCommunityFund = fProtectCoinbase;
+    const bool fMustShieldCoinBase = ForkManager::getInstance().mustCoinBaseBeShielded(chainActive.Height());
+    static bool fMustShieldCommunityFund = false;
+    // CF exemption allowed only after hfCommunityFundHeight hardfork
+    if (!ForkManager::getInstance().canSendCommunityFundsToTransparentAddress(chainActive.Height()))
+        fMustShieldCommunityFund = fMustShieldCoinBase;
 
-    pwalletMain->AvailableCoins(vAvailableCoins, fOnlyConfirmed, NULL, fIncludeZeroValue, fIncludeCoinBase, fIncludeCommunityFund);
+    pwalletMain->AvailableCoins(vAvailableCoins, fOnlyConfirmed, NULL, fIncludeZeroValue, !fMustShieldCoinBase, !fMustShieldCommunityFund);
```
ptagl: I see the code was a little misleading before, but I still struggle to quickly understand the meaning. I guess that we want to consider the shielding of the community fund only if it's not been explicitly disabled from the command line in REGTEST.

If that's the case, I would consider the following changes:
- Restore the previous variable names `fIncludeCoinBase` and `fIncludeCommunityFund`, as they match the meaning of the called function
- Change `canSendCommunityFundsToTransparentAddress(chainActive.Height())` to return false if we are in REGTEST
- Directly assign `fIncludeCommunityFund = ForkManager::getInstance().canSendCommunityFundsToTransparentAddress(chainActive.Height()`

qa/rpc-tests/shieldedpooldeprecation_rpc.py
JackPiri: Definitely yes, addressed accordingly.

qa/rpc-tests/test_framework/test_framework.py
JackPiri: In my opinion the short answer is no, some tuning has been already performed trying to keep the fork height as low as possible; trying to moving further down would require reorganization of whole tests (`wallet_shieldcoinbase.py` and `wallet_mergetoaddress.py`).

qa/rpc-tests/shieldedpooldeprecation_rpc.py
JackPiri: Ok makes sense, addressed accordingly.

qa/rpc-tests/wallet_protectcoinbase.py
JackPiri: I agree with you, added `sync_all()` calls as suggested.

qa/rpc-tests/wallet_shieldcoinbase.py
JackPiri: Yes, it would make the test more robust; added some assertion and performed better error handling on `wallet_protectcoinbase.py` and `wallet_shieldcoinbase.py`.

qa/rpc-tests/wallet_protectcoinbase.py
JackPiri: I have refactored a little bit the test removing the generation of those 101 blocks. Additionally, now `chainActive.Height() + 1` is used for hard fork check.

qa/rpc-tests/zcjoinsplit.py
JackPiri: Currently, if a transaction already contained in the mempool becomes irrevocably invalid in terms of contextual check, it would remain indefinitely in the mempool, the only way of getting rid of it being a node restart. I have checked, also for the hard fork PHGR->GROTH no feature involving mempool cleaning was provided. This improvement activity is tracked and will be performed in future sprint.

src/wallet/rpcwallet.cpp
JackPiri: Good catch. Now `chainActive.Height() + 1` is used for hard fork check.

qa/rpc-tests/shieldedpooldeprecation_rpc.py
JackPiri: Yes, modified accordingly.

qa/rpc-tests/zcjoinsplit.py
JackPiri: Refactored a little bit the flow as per your suggestion.

src/wallet/rpcwallet.cpp
JackPiri: Replying to your considerations:

- now `chainActive.Height() + 1` is used for hard fork check,
- agree, made coinbase and community fund being included if sending to z-addr,
- modified `canSendCommunityFundsToTransparentAddress` in a similar way to implementation of `mustCoinBaseBeShielded`; current network is considered in addition to blockchain height; please note that command line parameter `-regtestprotectcoinbase` could affect return value of `canSendCommunityFundsToTransparentAddress` (check attached image for a table reporting current behavior). table

Applied similar modification also to `sidechainrpc.cpp` and `wallet.cpp`.

src/sc/sidechainrpc.cpp
JackPiri: Replying to your considerations:

- restored previous names,
- modified `canSendCommunityFundsToTransparentAddress` (consider above comment related to this method),
- direct assignment of variables.

qa/rpc-tests/test_framework/test_framework.py
ptagl: Ok, let's keep track of this potential investigation/refactoring for the future, thanks!

a-petrini (approved):

drgora (approved):

src/sc/sidechainrpc.cpp
```diff
@@ -885,11 +885,10 @@ void ScRpcCmd::addInputs()
...
     static const bool fOnlyConfirmed = false;
     static const bool fIncludeZeroValue = false;
-    bool fProtectCoinbase = !Params().GetConsensus().fCoinbaseMustBeProtected;
-    static const bool fIncludeCoinBase = fProtectCoinbase;
-    static const bool fIncludeCommunityFund = fProtectCoinbase;
+    const bool includeCoinBase = !ForkManager::getInstance().mustCoinBaseBeShielded(chainActive.Height() + 1);
+    const bool includeCommunityFund = ForkManager::getInstance().canSendCommunityFundsToTransparentAddress(chainActive.Height() + 1);
 
-    pwalletMain->AvailableCoins(vAvailableCoins, fOnlyConfirmed, NULL, fIncludeZeroValue, fIncludeCoinBase, fIncludeCommunityFund);
+    pwalletMain->AvailableCoins(vAvailableCoins, fOnlyConfirmed, NULL, fIncludeZeroValue, includeCoinBase, includeCommunityFund);
```
ptagl: I may be missing something here, but wasn't the goal to restore the initial variable names here (meaning `fIncludeCoinBase` and `fIncludeCommunityFund`)? It's not a big deal obviously, just a matter of style consistency.

src/wallet/wallet.cpp
```diff
@@ -3058,34 +3058,30 @@ bool CWallet::SelectCoinsMinConf(const CAmount& nTargetValue, int nConfMine, int
...
-    // If coinbase utxos can only be sent to zaddrs, exclude any coinbase utxos from coin selection.
-    bool fProtectCoinbase = Params().GetConsensus().fCoinbaseMustBeProtected;
-    bool fProtectCFCoinbase = false;
-
-    // CF exemption allowed only after hfCommunityFundHeight hardfork
-    if (!ForkManager::getInstance().canSendCommunityFundsToTransparentAddress(chainActive.Height()))
-        fProtectCFCoinbase = fProtectCoinbase;
+    // If coinbase utxos can only be sent to zaddrs, exclude any coinbase utxos from coin selection (based on current available forks).
+    const bool includeCoinBase = !ForkManager::getInstance().mustCoinBaseBeShielded(chainActive.Height() + 1);
+    const bool includeCommunityFund = ForkManager::getInstance().canSendCommunityFundsToTransparentAddress(chainActive.Height() + 1);
```
ptagl: About style consistency, I guess we should use `fIncludeCoinBase` and `fIncludeCommunityFund` also here.

ptagl (changes requested):

src/sc/sidechainrpc.cpp
JackPiri: Yes, I renamed from `fMustShieldX` to `includeX`, forgetting about the prepended `f`; adapted accordingly with `fIncludeCoinBase` and `fIncludeCommunityFund`.

src/wallet/wallet.cpp
JackPiri: Yes, I renamed from `fMustShieldX` to `includeX`, forgetting about the prepended `f`; adapted accordingly with `fIncludeCoinBase` and `fIncludeCommunityFund`.

ptagl (approved):

PR #528, opened 2023-03-24
Fix `mempool_double_spend.py` test

ptagl:
The test `mempool_double_spend.py` was sporadically failing on the CI due to a wrong synchronization of mempools.

This PR fixed the issue and also adds some small improvements to the helper script `run_until_fails.py` to:
- add the number of iteration
- set the output mode to text

### DIFF
--- a/qa/rpc-tests/mempool_double_spend.py
+++ b/qa/rpc-tests/mempool_double_spend.py
@@ -35,9 +35,6 @@ class TxnMallTest(BitcoinTestFramework):
             self.join_network()
         else:
             self.split_network()
-
-        sync_blocks(self.nodes[1 : NUMB_OF_NODES])
-        sync_mempools(self.nodes[1 : NUMB_OF_NODES])
         
         self.sync_all()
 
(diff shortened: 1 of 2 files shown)


JackPiri (approved):
The PR is fine to me.
Thanks

qa/rpc-tests/mempool_double_spend.py
```diff
@@ -134,7 +131,7 @@ def run_test(self):
 
         # Sync mempools except for the last (disconnected) node
         mark_logs("Waiting for honest nodes to sync mempool", self.nodes, DEBUG_MODE)
-        sync_mempools(self.nodes[1 : NUMB_OF_NODES - 1])
+        sync_mempools(self.nodes[0 : NUMB_OF_NODES - 1])
```
a-petrini, line 134: A minor comment definitely not blocking.
Here we are assuming that the dishonest node DOUBLE_SPEND_NODE_INDEX is the last one on the list.
A more robust approach would be to initialize the full list of nodes, removing the DOUBLE_SPEND_NODE_INDEX element, and use this list whenever we need to operate on the honest partition:
`honest_nodes_list = list(range(NUMB_OF_NODES))`
`honest_nodes_list.remove(DOUBLE_SPEND_NODE_INDEX)`

a-petrini (approved):
PR looks fine to me, left a minor comment which is definitely not blocking.

drgora (approved):

qa/rpc-tests/mempool_double_spend.py
ptagl: I like this suggestion, I think it makes the code more readable and less error-prone.

If you don't mind, I would do that in a separate refactoring PR, also to take the chance to review all the occurrences in other Python tests and apply the change to the whole repository.

a-petrini: Yes, agree!

PR #529, opened 2023-03-24
Fix rpc-tests.sh script for OSX CI

a-petrini:
- Fixed --rpc-exclude flag
- Fixed string splitting when converting the string returned by the Python script to bash array


ptagl (approved):

drgora (approved):
Looks good to me, thanks

PR #531, opened 2023-03-30
Gp/remove enable_address_indexing

JackPiri:
This PR removes the directive ENABLE_ADDRESS_INDEXING, simplifying both the release process (one set of binaries) and the codebase maintenance (one logic flow).

Regarding LevelDB configuration, it was decided to:

- remove compression setting
- remove maxopenfiles setting
- set compression disabled for both blocktreedb and coinsviewdb
- set maxopenfiles setting to 1000 for both blocktreedb and coinsviewdb

Initial discussion available on PR https://github.com/HorizenOfficial/zen/pull/499.


src/txmempool.cpp
```diff
@@ -686,13 +685,11 @@ void CTxMemPool::remove(const CTransactionBase& origTx, std::list<CTransaction>&
 
             const uint256& scid = cert.GetScId();
 
-#ifdef ENABLE_ADDRESS_INDEXING
             // are we removing a top-quality cert?
             const uint256& topQualHash = mapSidechains.at(scid).GetTopQualityCert()->second;
             CSidechain sidechain;
             assert(pcoinsTip->GetSidechain(scid, sidechain));
             bool isTopQualityCert = (topQualHash == hash) || sidechain.isNonCeasing();
-#endif // ENABLE_ADDRESS_INDEXING
```
a-petrini, line 695: Just a minor non-blocking comment: all in all, the variable `isTopQualityCert` is used only in the `if` condition at line 730. Moving these 5 lines inside the if statement would save accessing the two maps whenever address indexing is not enabled.

a-petrini (approved):
Left a minor non-blocking comment. PR looks very good!

src/leveldbwrapper.h
```diff
@@ -16,6 +16,27 @@
...
+// do not interfere with select() loops. On 64-bit Unix hosts this value is
+// also OK, because up to that amount LevelDB will use an mmap
+// implementation that does not use extra file descriptors (the fds are
+// closed after being mmaped).
+// Increasing the value beyond the default is dangerous because LevelDB will
+// fall back to a non-mmap implementation when the file count is too large.
+// On 32-bit Unix host we should decrease the value because the handles use
+// up real fds, and we want to avoid fd exhaustion issues.
+constexpr unsigned int DB_MAX_OPEN_FILES = 1000;
+
```
AndreaLanfranchi: Imho all of this is not needed any more.
- DB_COMPRESSION is only used in GetOptions() function where we can set `options.compression = leveldb::kNoCompression`
- DB_MAX_OPENFILES is only used in GetOptions() function where we can even remove the line `options.max_open_files = DB_MAX_OPEN_FILES` as the value is already hardcoded in the explicit constructor of  `leveldb::Options` struct

Eventually move the comments in `GetOptions` function to explain the rationale behind the setting of these constants

src/main.h
```diff
@@ -49,6 +47,28 @@ class CTxUndo;
...
+    #endif
+#else
+    #error Cannot detect compiler or compiler is not supported
+#endif
+#if defined(BITNESS_32)
+#error "Zend is not supported on 32 bit architecture"
+#endif
+#undef BITNESS_32
+#undef BITNESS_64
+
```
AndreaLanfranchi, line 65: Given we know [there are problems on big endian archtectures](https://zcash.readthedocs.io/en/latest/rtd_pages/security_warnings.html#x86-64-linux-only) I believe is worth to add the proper checks here

```suggestion
#if defined(_WIN32)
    // On Windows we assume little-endian
#elif defined(__BYTE_ORDER__)
  #if (__BYTE_ODER__ != __ORDER_LITTLE_ENDIAN__)
    #error "Zend is not supported on big-endian architectures"
  #endif
#else
  #error "Undetectable endianness"
#endif
```

src/main.h
```diff
@@ -47,6 +47,28 @@ class CTxUndo;
...
+    #if defined(__x86_64)
+        #define BITNESS_64
+    #else
+        #define BITNESS_32
+    #endif
+#else
+    #error Cannot detect compiler or compiler is not supported
+#endif
+#if defined(BITNESS_32)
+#error "Zend is not supported on 32 bit architecture"
```
drgora: Sorry to resume this, but after my last pass I noticed that here we are saying that we do support MSVC and Intel compilers, which is not quite true as of today. Same is true for architectures different from `x86_64`.

I'd suggest to remove lines 50-56, and also remove `__INTEL_COMPILER` from line 57. Finally, I'd change the message at line 64 to include include an unsupported architecture as a possible problem.

src/Makefile.am
```diff
@@ -229,11 +224,10 @@ BITCOIN_CORE_H = \
...
   zmq/zmqnotificationinterface.h \
-  zmq/zmqpublishnotifier.h
-
-if ENABLE_ADDRESS_INDEXING
-  BITCOIN_CORE_H += $(ADDR_INDEX_H)
-endif
+  zmq/zmqpublishnotifier.h \
+  addressindex.h \
+  spentindex.h \
+  timestampindex.h
```
drgora: Just a minor thing here, but I'd suggest to keep the alphabetical order for this list

src/main.h
AndreaLanfranchi: I do believe the whole point here is to detect BITNESS while making no assertion we do support any of these compilers.
It is true we do not support MSVC but nothing prevents anybody from trying to build with it (fork the project and try): and at least BITNESS constraint will be honored.
About this

> Same is true for architectures different from x86_64.

I'm not sure I understand the point unless you mean to change line 66 to `#if !defined(BITNESS_64)`

drgora: > I do believe the whole point here is to detect BITNESS while making no assertion we do support any of these compilers. It is true we do not support MSVC but nothing prevents anybody from trying to build with it (fork the project and try): and at least BITNESS constraint will be honored.

I see your point, that was also my very first thought, but then 2 things made me prefer my suggestion: (1) we are adding lines of code to explicitly support this check for MSVC, and this is not consistent with the rest of our codebase and workflow, where never consider or check compatibility with MSVC explicitly. And (2), I read the message `or compiler is not supported` at line 67, combined with the explicit check for MSVC, as "we support the compilers above, including MSVC". I'd rather prevent this misunderstanding.

I also agree on the fact that someone might try to compile zend with MSVC (and we should also do that at some point), but given that we do not test it regularly and there might be many undocumented side effects (e.g. `NDEBUG`), I believe it makes sense to have this check fail out of the box. The brave user wanting to test MSVC can fix it very quickly, and we might even add a comment to facilitate that, in case.

> 
> > Same is true for architectures different from x86_64.
> 
> I'm not sure I understand the point unless you mean to change line 66 to `#if !defined(BITNESS_64)`

Yeah sorry that was not very clear. With the current code, if a brave user tries to compile for `ARM64` on `clang` or `g++`, we print the error message `Zend is not supported on 32 bit architecture`, which is misleading. I am suggesting to adjust the error message to account for this. For example, we might update the error message to something like `Zend only supports x86_64 architecture`.

src/Makefile.am
JackPiri: Ok, modified accordingly.

src/main.h
JackPiri: Ok, modified as per your suggestion.

src/leveldbwrapper.h
JackPiri: I agree with your opinion; I've moved the comments inside `GetOptions`, I prefer making `options.max_open_files = 1000` explicit rather than relying on the default of `leveldb::Options` structure.

src/txmempool.cpp
JackPiri: Ok, seems reasonable, modified accordingly.

src/main.h
AndreaLanfranchi: Modify or remove according to your needs

src/txmempool.cpp
a-petrini: Looks fine to me!

JackPiri: Reverted this because of failure on `sc_cert_addressindex.py` and `sc_cert_addrmempool.py`.
Analysing the code and inspecting some logs, I'd say the problem with proposed change is that the `GetTopQualityCert` call is moved after `EraseCert` call; I see occurrences where the original top quality cert is the one being erased, hence the new top quality cert is different from the original, leading to some unexpected behavior.

drgora (approved):
Looks good to me, thanks

ptagl (approved):

PR #532, opened 2023-03-31
Gp/automated release preparation

JackPiri:
This PR introduces a new script for automating the release preparation procedure.

Notes for the reviewers:

- if trying to run the script from `gp/automated_release_preparation`, lines 216-217-218 have to be commented out or removed, because they perform a check that currently checkedout branch is named `main`
- an example of configuration file for generating dummy release "9.9.9-50" is provided in attachment [prepare_release_config-9.9.9.txt](https://github.com/HorizenOfficial/zen/files/11123804/prepare_release_config-9.9.9.txt) (`.yaml` extension modified to `.txt` for GitHub not supporting the first one) (check template file `./contrib/devtools/prepare_release/prepare_release_config.yaml` for more general reference)
- few TODOs are left at the end of the script for temporary reference, they will be removed if agree on not including them in this PR


contrib/devtools/prepare_release/prepare_release.py
```diff
@@ -0,0 +1,526 @@
...
+   result = subprocess.run(["git", "checkout", "-b", branch_name], capture_output=True, text=True, cwd=config[k_repository_root])  # Commit changes with the given message
+   if (result.returncode != 0):
+        print("Branch creation failed")
+        sys.exit()
+
+# utility function
+def get_version_string_details(version_string: str):
+    assert(version_string.count(".") == 2)
+    digits = [int(digit) for digit in version_string if digit.isdigit()]
+    assert(len(digits) == 3)
```
a-petrini: This will break for
1) versions not having single-digit components, i.e. `3.10.3`
2) versions not following a strict `major.minor.patch` semantic versioning convention, i.e. `4.0.0-beta2`.

One way could be something like this:
`digits = [int(digit) for digit in version_string.split(sep='.') if digit.isdigit()]`
that does not fail for 1) and explicitly fails for 2), thus enforcing semantic versioning

contrib/devtools/prepare_release/prepare_release.py
```diff
@@ -0,0 +1,526 @@
...
+        print("[25, 49] when releasing a release candidate (e.g. 3.3.1-rc1, 3.3.1-rc2, etc.)")
+        print("    [50] when making a standard release (e.g. 3.3.1)")
+        config[k_build] = input("")
+
+    branch_name = f"release/{config[k_version]}"
+    build_number = int(config[k_build])
+    if (1 <= build_number and build_number <= 24):
+        branch_name += f"-beta{config[k_build]}"
+    elif (25 <= build_number and build_number <= 49):
+        branch_name += f"-rc{config[k_build]}"
```
a-petrini: As in this case 25 <= `build_number` < 50, are we appending this number to the branch name, after `-rc`?

contrib/devtools/prepare_release/prepare_release.py
```diff
@@ -0,0 +1,526 @@
...
+    git_commit(f"Set clientversion {config[k_version]} (build {config[k_build]})")
+
+def update_checkpoints():
+    # Get a mainnet checkpoint candidate
+    mainnet_checkpoint = get_checkpoint_candidate(k_mainnet, "https://explorer.horizen.io/")
+    testnet_checkpoint = get_checkpoint_candidate(k_testnet, "https://explorer-testnet.horizen.io/")
+
+    # Find the line numbers containing the last checkpoint
+    # for both mainnet and testnet
+    line_numbers = find_lines_containing(os.path.join(config[k_repository_root], "src/chainparams.cpp"), r'")),')
```
a-petrini: While searching for the substring `")),` works (there are just three occurrences in the whole file, conveniently placed) I may say that it is not very robust.
Relying on a regex pattern may be better in this case:
`pattrn = re.compile('.*\(.*uint256S.*0x.*\).*\).*\,')`
`line_numbers = [i for i, line in enumerate(lines) if pattrn.match(line) is not None]`

contrib/devtools/prepare_release/prepare_release.py
```diff
@@ -0,0 +1,526 @@
...
+    assert(version_string.count(".") == 2)
+    digits = [int(digit) for digit in version_string if digit.isdigit()]
+    assert(len(digits) == 3)
+    return digits
+
+def replace_string_in_file(filepath: str, old_string_regex: str, new_string: str):
+    with open(filepath, "r") as file:
+        file_data = file.read()
+
+    new_data = re.sub(old_string_regex, new_string, file_data)
```
a-petrini, line 87: It would be safer to check whether `re.sub` managed to do the substitution by comparing `new_data` with `file_data`, and return an error message / quit the script otherwise.

contrib/devtools/prepare_release/prepare_release.py
```diff
@@ -0,0 +1,526 @@
...
+
+def update_checkpoints():
+    # Get a mainnet checkpoint candidate
+    mainnet_checkpoint = get_checkpoint_candidate(k_mainnet, "https://explorer.horizen.io/")
+    testnet_checkpoint = get_checkpoint_candidate(k_testnet, "https://explorer-testnet.horizen.io/")
+
+    # Find the line numbers containing the last checkpoint
+    # for both mainnet and testnet
+    line_numbers = find_lines_containing(os.path.join(config[k_repository_root], "src/chainparams.cpp"), r'")),')
+    print(line_numbers)
```
a-petrini: Also, asserting that we get exactly 3 results would improve safety

contrib/devtools/prepare_release/prepare_release.py
```diff
@@ -0,0 +1,526 @@
...
+    line = re.sub(r"(\d+)", config[k_mainnet][k_total_transactions], line).rstrip()
+    replace_file_line(os.path.join(config[k_repository_root], "src/chainparams.cpp"), line_numbers[0] + 1, line)
+    
+    if (interactive):
+        config[k_testnet][k_total_transactions] = input(f"Please, enter the total number of transactions for TESTNET block {testnet_checkpoint['height']}: ")
+    line = read_file_line(os.path.join(config[k_repository_root], "src/chainparams.cpp"), line_numbers[1] + 1)
+    line = re.sub(r"(\d+)", config[k_testnet][k_total_transactions], line).rstrip()
+    replace_file_line(os.path.join(config[k_repository_root], "src/chainparams.cpp"), line_numbers[1] + 1, line)
+
+    average_mainnet_transactions = round(int(config[k_testnet][k_total_transactions]) / (mainnet_checkpoint['height'] / (24 * 24)))
```
a-petrini: `round(int(config[k_mainnet][k_total_transactions])`
`                 ~~~~~~~~~`

contrib/devtools/prepare_release/prepare_release.py
```diff
@@ -0,0 +1,526 @@
...
+
+    git_commit("Update Debian package info")
+
+def update_deprecation_height():
+    if (interactive):
+        config[k_approx_release_height] = input("Enter the approximate release height (mainnet): ")
+        config[k_weeks_until_deprecation] = input("Enter the weeks until deprecation: ")
+
+    approx_release_height_line = read_file_line(os.path.join(config[k_repository_root], "src/deprecation.h"), 7)
+    weeks_until_deprecation_line = read_file_line(os.path.join(config[k_repository_root], "src/deprecation.h"), 8)
```
a-petrini: Here, as in other sections of the script, we are relying on the fact that the information we are looking for are always on a pre-determined line number, and it's ok if never manually edit their position.
I was wondering if it may be the case of adding a comment on these line, such as:
`// Automatically filled by prepare_release.py - do not alter the position of this line`
or modify the script so that it automatically detects the line number. In this case, something like:
```
with open(os.path.join(config[k_repository_root], "src/deprecation.h"), 'r') as inF:
    for i, line in enumerate(inF):
        if re.match("static const int APPROX_RELEASE_HEIGHT.*", line):
            approx_release_height_line = line
            approx_release_height_line_num = i
        if re.match("static const int WEEKS_UNTIL_DEPRECATION.*", line):
            weeks_until_deprecation_line = line
            weeks_until_deprecation_line_num = i

```

a-petrini (changes requested):

contrib/devtools/prepare_release/prepare_release.py
JackPiri: Correct, that was a bug; modified accordingly.
Regarding the second point, the version is actually just the `major.minor.patch` string, the part related to build version is used just for branch name; the comment in config file already mentioned this, I have also updated the interactive instruction message in order to make it more clear.

contrib/devtools/prepare_release/prepare_release.py
JackPiri: Yes, this choice was made based on current release branches contained in repository (`release/4.0.0-rc1`).

contrib/devtools/prepare_release/prepare_release.py
JackPiri: This seems more robust, updated accordingly.

contrib/devtools/prepare_release/prepare_release.py
JackPiri: Ok, added the check; print and exit if failing.

contrib/devtools/prepare_release/prepare_release.py
JackPiri: Checking `assert(len(line_numbers >= 2))` since we are actually using only the first two elements.

contrib/devtools/prepare_release/prepare_release.py
JackPiri: Correct, that was a bug; modified accordingly.

contrib/devtools/prepare_release/prepare_release.py
JackPiri: I've opted for the second option which seems to me more reliable.

contrib/devtools/prepare_release/prepare_release.py
```diff
@@ -0,0 +1,542 @@
...
+def git_check_currently_on_main():
+    result = subprocess.run(["git", "rev-parse", "--abbrev-ref", "HEAD"], capture_output=True, text=True, cwd=config[k_repository_root])
+    return result.returncode == 0 and result.stdout.rstrip() == "main"
+
+def git_check_pending_changes():
+    result = subprocess.run(["git", "diff"], capture_output=True, text=True, cwd=config[k_repository_root])
+    return result.returncode == 0 and result.stdout != ""
+
+def git_commit(commit_title: str):
+   subprocess.run(["git", "add", "."], cwd=config[k_repository_root])  # Add all changes to the index
```
drgora: The scope of this git add is quite broad, and we might end up adding unwanted files (e.g. build outputs, logs...) to the git repository.
We could either use `git commit -a` directly, or if we need to add new files, run `git add` only on the directories where new files are supposed to be added? I guess we could pass the directory as a parameter.

contrib/devtools/README.md
```diff
@@ -100,3 +100,29 @@ maintained:
...
+==================
+
+Inside this folder the following are provided:
+
+- a script (prepare_release.py),
+- a template config file (prepare_release_config.yaml)
+- various config files associated to past releases preparations (prepare_release_config-X.Y.Z.yaml)
+
+Run the script for preparing a new release.
+It is suggested to run the script from a terminal located a repository root.
```
ptagl:
```suggestion
It is suggested to run the script from a terminal located at repository root. 
```

contrib/devtools/prepare_release/prepare_release.py
```diff
@@ -0,0 +1,542 @@
...
+    future_deprecation_date = last_block_date + datetime.timedelta(seconds=(future_deprecation_height - last_block_height) * block_time)
+
+    # Estimate the next deprecation date
+    date_string = future_deprecation_date.strftime('%Y-%m-%d')
+
+    git_commit(f"Set deprecation height {future_deprecation_height}/{date_string}")
+
+def build_zend():
+    result_clean = subprocess.run(["make", "distclean"], cwd=config[k_repository_root])
+    result_build = subprocess.run([f"./zcutil/build.sh", "-j8"], cwd=config[k_repository_root])
```
drgora: I guess this is a leftover, since you are not using any variable in this string:
```suggestion
    result_build = subprocess.run(["./zcutil/build.sh", "-j8"], cwd=config[k_repository_root])
```

contrib/devtools/prepare_release/prepare_release.py
a-petrini: Nice! Thanks for also updating the string message in interactive mode.

contrib/devtools/prepare_release/prepare_release.py
a-petrini: ok!

contrib/devtools/prepare_release/prepare_release.py
a-petrini: ok!

contrib/devtools/prepare_release/prepare_release.py
a-petrini: nice, thanks!

contrib/devtools/prepare_release/prepare_release.py
a-petrini: Thanks for the correction, I see that now the right number gets attached to the build version string.

a-petrini (approved):

contrib/devtools/prepare_release/prepare_release_config.yaml
```diff
@@ -0,0 +1,79 @@
+# THIS IS A TEMPLATE FILE, FOR EXAMPLES ON ACTUAL RELEASES CHECK CONTENT OF prepare_release-X.Y.Z.yaml FILES
+# path to zend project root
+repository_root: "./"
+# new release version (format "major.minor.patch")
+version: ""
+# build number
```
ptagl: What about adding some instructions on how to set the proper number (e.g. 1-24 for beta versions, etc.)?

drgora (approved):
I left only two non blocking comments, but this is ok for merge by me

contrib/devtools/prepare_release/prepare_release.py
JackPiri: Yes you are right, modified accordingly.

contrib/devtools/prepare_release/prepare_release.py
JackPiri: Added optional parameter for specifying a list of directories involved in the `git add` command; current flow uses this otpion twice.

contrib/devtools/prepare_release/prepare_release_config.yaml
```diff
@@ -0,0 +1,79 @@
+# THIS IS A TEMPLATE FILE, FOR EXAMPLES ON ACTUAL RELEASES CHECK CONTENT OF prepare_release-X.Y.Z.yaml FILES
+# path to zend project root
+repository_root: "./"
+# new release version (format "major.minor.patch")
+version: ""
```
ptagl: Can we provide an example and also specify what to do when we want to release a version like `4.0.0-rc2`? Should I include `rc2` in the version field?

contrib/devtools/prepare_release/prepare_release_config.yaml
```diff
@@ -0,0 +1,79 @@
...
+  # total transaction from genesis block up to checkpoint (can be found in "debug.log" file)
+  total_transactions: ""
+# testnet config
+testnet:
+  # height for checkpoint (if empty, current last block is used)
+  checkpoint_height: ""
+  # total transaction from genesis block up to checkpoint (can be found in "debug.log" file)
+  total_transactions: ""
+# release date (format "%a, %d %b %Y")
+release_date: ""
```
ptagl: Can we add an example date to make it more immediate to understand the format string?

contrib/devtools/prepare_release/prepare_release.py
drgora: Thank you, I like it

contrib/devtools/README.md
JackPiri: Correct.

contrib/devtools/prepare_release/prepare_release_config.yaml
JackPiri: Added instructions in the comment.

contrib/devtools/prepare_release/prepare_release_config.yaml
JackPiri: Added details in the comment `suffix must not be included if beta or candidate`.

contrib/devtools/prepare_release/prepare_release_config.yaml
JackPiri: In some sense I agree with you, but I prefer to keep this file as a template (no examples included); we agreed on inserting the config file used for each release generation in the `zen/contrib/devtools/prepare_release` folder (similar to what we already do for release notes files), so if a user need to check for an example there are actual files available for inspection.

JackPiri: Updated just the comment.

ptagl (approved):

PR #536, opened 2023-04-13
Update OSX CI and other dependencies

drgora:
The main goal of this PR is to fix compilation on/for recent macOS versions, which becomes necessary after Boost update to version 1.81.
macOS binaries compiled after this PR only will be compatible with macOS version 10.14 (`Mojave`) or newer.

In order to achieve such a goal, we:
- update the CI to use the `xcode12.3` image, which includes macOS 11.1  and Xcode 12.3
- update or backport patches for dependencies (gtest, leveldb, zeromq)
- fix many compilation warnings generated by recent clang versions

This PR also includes a number of other improvements. Specifically, we:
- ditch the usage of the `zencash-apple` repository altogether, since now the prefer the build tools provided by travis images
- move from `boost::optional` to `std::optional` (fixing errors with recent gtest)
- fix several inconsistencies in the build flags used for different libraries by the build system
- fix warnings generated by `Make`
- default to `-O3` optimization level for all release builds (linux, mingw, and osx)


ptagl (approved):
It looks good to me, great job, thanks!

src/util.h
```diff
@@ -56,12 +57,14 @@ extern std::atomic<bool> fReopenDebugLog;
...
+ * Translation function: Call Translate signal on UI interface, which returns a std::optional result.
  * If no translation slot is registered, nothing is returned, and simply return the input.
  */
 inline std::string _(const char* psz)
 {
-    boost::optional<std::string> rv = translationInterface.Translate(psz);
+    std::optional<std::string> rv;
+    boost::optional<std::string> tmp = translationInterface.Translate(psz);
+    if (tmp) rv = tmp.get();
     return rv ? (*rv) : psz;
```
a-petrini: Totally non-blocking comment: we could use `std::optional` member function `value_or()` to avoid the ternary operator on the return statement.

contrib/ci-horizen/scripts/prepare/build_zencash-apple.sh
```diff
@@ -1,9 +0,0 @@
-#!/bin/bash
-
-set -euo pipefail
-
-git clone "${CLONE_REPO}" "${CLONE_TARGET}"
```
a-petrini, line 5: Another non-blocking comment: as we no longer rely on the `zencash-apple` repository, we could also remove the definition of the env variable `CLONE_REPO` from `setup_environment.sh`

a-petrini (approved):
Left a couple of non-blocking comments, and everything looks great to me. High quality PR which introduces many improvements!

src/util.h
drgora: Good idea, I implemented the suggested change, thanks

contrib/ci-horizen/scripts/prepare/build_zencash-apple.sh
drgora: Definitely, I removed that and few other variables from `setup_environment.sh` which are no longer needed. Good catch, thanks!

PR #537, opened 2023-04-14
Consider spent tx in p2p

drgora:
This patch makes sure that zend does not ask for tx data to its peers for tx which have already been received and spent.

Notice: this is a rebased clone of #494, which had previously been reverted.

### DIFF
--- a/qa/pull-tester/rpc-tests.sh
+++ b/qa/pull-tester/rpc-tests.sh
@@ -161,7 +161,7 @@ testScripts=(
   'sc_getcertmaturityinfo.py',68,219
   'sc_big_commitment_tree.py',63,110
   'sc_big_commitment_tree_getblockmerkleroot.py',11,25
-  'p2p_ignore_spent_tx.py',1,1
+  'p2p_ignore_spent_tx.py',215,443
 );
 
 testScriptsExt=(
--- a/qa/pull-tester/rpc-tests.sh
+++ b/qa/pull-tester/rpc-tests.sh
@@ -161,6 +161,7 @@ testScripts=(
   'sc_getcertmaturityinfo.py',68,219
   'sc_big_commitment_tree.py',63,110
   'sc_big_commitment_tree_getblockmerkleroot.py',11,25
+  'p2p_ignore_spent_tx.py',1,1
 );
 
 testScriptsExt=(
--- a/qa/rpc-tests/p2p_ignore_spent_tx.py
+++ b/qa/rpc-tests/p2p_ignore_spent_tx.py
@@ -21,9 +21,6 @@ class old_tx_ignored(BitcoinTestFramework):
     def setup_chain(self, split=False):
         print("Initializing test directory " + self.options.tmpdir)
         initialize_chain_clean(self.options.tmpdir, NUMB_OF_NODES)
-        self.alert_filename = os.path.join(self.options.tmpdir, "alert.txt")
-        with open(self.alert_filename, 'w'):
-            pass  # Just open then close to create zero-length file
 
     def setup_network(self, split=False):
         self.nodes = []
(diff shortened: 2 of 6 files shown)


qa/pull-tester/rpc-tests.sh
```diff
@@ -161,6 +161,7 @@ testScripts=(
   'sc_getcertmaturityinfo.py',68,219
   'sc_big_commitment_tree.py',63,110
   'sc_big_commitment_tree_getblockmerkleroot.py',11,25
+  'p2p_ignore_spent_tx.py',1,1
```
JackPiri: It could be worth to have a couple of runs on Travis in order to have an estimation of the execution times and update them accordingly.

src/net.h
```diff
@@ -176,6 +178,7 @@ extern std::map<CInv, CDataStream> mapRelay;
 extern std::deque<std::pair<int64_t, CInv> > vRelayExpiration;
 extern CCriticalSection cs_mapRelay;
 extern limitedmap<CInv, int64_t> mapAlreadyAskedFor;
+extern limitedmap<CInv, int64_t> mapAlreadyReceived;
```
JackPiri, line 181: Just a light opinion, why not renaming `mapAlreadyReceived` with something like `mapAlreadyReceivedTxs`? I'd like to make explicit this map refers to txs only, compared to other data-structure like `mapAlreadyAskedFor`, `setAskFor`, `mapAskFor` which work generically on invs.

src/net.cpp
```diff
@@ -2456,6 +2457,13 @@ void CNode::AskFor(const CInv& inv)
     if (!setAskFor.insert(inv.hash).second)
         return;
 
+    // If we need to ask for this inv again (after it has already been received)
+    // then pretend we never received it before so that the request is actually performed.
+    // Otherwise, this request would be blocked in main::SendMessages.
+    if (mapAlreadyReceived.erase(inv)) {
+        LogPrint("net", "%s():%d - askfor %s even though it was received already in the past\n", __func__, __LINE__, inv.ToString());
```
JackPiri, line 2464: Could be worth making this condition like `if (inv.type == MSG_TX && mapAlreadyReceived.erase(inv))` (not particularly interested in performance, rather for making explicit we are targeting txs)?

qa/rpc-tests/p2p_ignore_spent_tx.py
```diff
@@ -0,0 +1,88 @@
...
+NUMB_OF_NODES = 3
+FEE = Decimal("0.0001")
+
+class old_tx_ignored(BitcoinTestFramework):
+    alert_filename = None
+
+    def setup_chain(self, split=False):
+        print("Initializing test directory " + self.options.tmpdir)
+        initialize_chain_clean(self.options.tmpdir, NUMB_OF_NODES)
+        self.alert_filename = os.path.join(self.options.tmpdir, "alert.txt")
```
JackPiri:
```
        self.alert_filename = os.path.join(self.options.tmpdir, "alert.txt")
        with open(self.alert_filename, 'w'):
            pass  # Just open then close to create zero-length file
```

I think this is a leftover due to copy-paste from other test (btw, I see this in many Python tests, almost always unused; maybe it has some meaning I am not aware of...).

JackPiri (approved):
I left a few non-blocking and opinion-based comments; feel free to address or ignore them.
The PR is already fine to me.
Thanks

qa/pull-tester/rpc-tests.sh
drgora: Definitely, this was just a placeholder for the first run of the CI. Updated with actual measurements.

qa/rpc-tests/p2p_ignore_spent_tx.py
drgora: Good point, you're right! I removed these lines, which seems to be unneeded.

src/net.h
drgora: While the description of this PR focuses on the tx type, I'd rather leave it generic. `mapAlreadyReceived` can be used for generic invs, and in fact it could also be used to filter blocks that did not make it into `mapBlockIndex`, for example in case of failed header validation.

src/net.cpp
drgora: Please see my comment above. I'd rather keep it generic, for any `inv`.

ptagl (approved):

PR #539, opened 2023-04-19
Gp/fix prioritisetransaction

JackPiri:
This PR fixes some issues related to `prioritisetransaction` test.
The main issue was related to the fact that monitored transactions were checked for being included/mined (or not included/not mined) in blocktemplate/block without taking into account the two different criteria of inclusion/mining: by priority and by fee rate. Hence, some small fluctuations (hashes, txs sizes, fees rates) could make the test passing or failing in a non reliable way. This PR makes explicit distinction between the two criteria and allows for proper detection of inclusion/mining (or not inclusion/not mining).
As side refactoring, the` run_until_fails.py` utility script has been updated for using `python3` in place of `python` and a bug fix related to `mempoolToJSON` method has been included.

### DIFF
--- a/qa/rpc-tests/prioritisetransaction.py
+++ b/qa/rpc-tests/prioritisetransaction.py
@@ -97,8 +97,6 @@ class PrioritiseTransactionTest (BitcoinTestFramework):
         self.nodes[0].generate(501)
         self.sync_all()
 
-        base_fee = self.nodes[0].getnetworkinfo()['relayfee']
-
         # 11 kb blocks will only hold about 50 txs, so this will fill mempool with older txs
         taddr = self.nodes[1].getnewaddress()
         for _ in range(900):
@@ -147,6 +145,10 @@ class PrioritiseTransactionTest (BitcoinTestFramework):
             self.nodes[0].sendtoaddress(self.nodes[1].getnewaddress(), 0.1)
             self.sync_all()
 
+            # Keep track of which will be the last tx to mine by priority
+            last_tx_to_include_by_priority_0 = self.get_last_tx_selectable_by_priority(self.nodes[0].getrawmempool(True))
+            assert_equal(last_tx_to_include_by_priority_0 != None, True)
+
             # Check that priority_tx_0 was not included by priority through getblocktemplate()
             # (too soon)
             block_template = self.nodes[0].getblocktemplate()
@@ -182,8 +184,7 @@ class PrioritiseTransactionTest (BitcoinTestFramework):
             self.nodes[1].prioritisetransaction(priority_tx_1, max_priority, 0)
 
             # Keep track of which will be the last tx to mine by priority
-            rawmempool = self.nodes[0].getrawmempool(True)
-            last_tx_to_mine_by_priority_0 = self.get_last_tx_selectable_by_priority(rawmempool)
+            last_tx_to_mine_by_priority_0 = self.get_last_tx_selectable_by_priority(self.nodes[0].getrawmempool(True))
             assert_equal(last_tx_to_mine_by_priority_0 != None, True)
 
             # Mine block on node 0 and sync
--- a/qa/rpc-tests/prioritisetransaction.py
+++ b/qa/rpc-tests/prioritisetransaction.py
@@ -53,12 +53,11 @@ class PrioritiseTransactionTest (BitcoinTestFramework):
         last_tx_selectable_by_priority = None
         for index in range(len(sorted_rawmempool)):
             tx_size = sorted_rawmempool[index][1]["size"]
+            last_tx_selectable_by_priority = sorted_rawmempool[index][0]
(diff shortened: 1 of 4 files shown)


qa/rpc-tests/prioritisetransaction.py
```diff
@@ -9,23 +9,88 @@
...
+        return sorted(newdict.items(), key=lambda x: x[1]["total_priority" if priority_otherwise_fee else "total_fee"], reverse=True)
+    
+    def get_max_priority_and_max_fee(self, rawmempool):
+        sorted_rawmempool = self.get_sorted_by(rawmempool, True)
+        max_priority = sorted_rawmempool[0][1]["total_priority"]
+        sorted_rawmempool = self.get_sorted_by(rawmempool, False)
+        max_fee = sorted_rawmempool[0][1]["total_fee"]
+        return max_priority, max_fee
+
+    def get_last_tx_selectable_by_priority(self, rawmempool):
```
drgora, line 50: If the mempool does not fill completely the high priority partition of the block, then the name of this function becomes misleading, in my opinion. I'd suggest renaming, but I cannot suggest any meaningful yet compact name, or maybe just clarify in a comment that this function expects that the mempool has enough transactions to return not `None`.

qa/rpc-tests/prioritisetransaction.py
```diff
@@ -38,95 +103,127 @@ def run_test (self):
...
+            # Mempool is probed every 10 seconds. We'll give getblocktemplate() a maximum of {timeout_seconds} seconds to refresh
+            start = time.time()
+            in_block_template = False
+            while in_block_template == False:
+                block_template = self.nodes[0].getblocktemplate()
+                tx_0_included_by_priority, tx_0_included = self.check_if_included_by_priority(priority_tx_0, last_tx_to_include_by_priority_0, block_template)
+                if (round == 0 and tx_0_included_by_priority):
+                    break
+                elif (round == 1 and tx_0_included):
+                    break
```
drgora, line 185: Just a style thing, but I find it a bit more readable and debug friendly (e.g. at round 0, if tx is included but not by priority). How do you like this suggestion, which also replaces lines 172-176?
```suggestion
                if (tx_0_included):
                    assert_equal(round == 0, tx_0_included_by_priority)
                    assert_equal(round == 1, !tx_0_included_by_priority)
                    break
```

src/rpc/blockchain.cpp
```diff
@@ -436,8 +436,11 @@ UniValue mempoolToJSON(bool fVerbose = false)
             const auto& p = entry.second.first;
             const auto& f = entry.second.second;
             UniValue info(UniValue::VOBJ);
-            info.pushKV("fee", ValueFromAmount(f));
-            info.pushKV("priority", p);
+            if (o.exists(hash.ToString()))
```
drgora, line 439: Nice catch!

qa/rpc-tests/prioritisetransaction.py
```diff
@@ -9,124 +9,222 @@
...
+            if (block["tx"][index] == last_tx_by_priority):
+                last_tx_by_priority_mined = True
+            if (block["tx"][index] == tx):
+                tx_mined = True
+                if (not last_tx_by_priority_mined):
+                    tx_mined_by_priority = True
+                break
+        return tx_mined_by_priority, tx_mined
+
     def run_test (self):
```
drgora, line 91: This might be a good chance to add a brief high level description as we do for other tests, to give some quick context on the purpose of this test

drgora (approved):
I just left few minor comments, but feel free to ignore them. This is already good for merge to me.

qa/rpc-tests/prioritisetransaction.py
JackPiri: I agree with your comment, I have moved the assignment `last_tx_selectable_by_priority = sorted_rawmempool[index][0]` upward outside the `if` condition; in this way even in the case of a not-filling mempool the desired tx should be selected (quickly tested and it seems fine).

qa/rpc-tests/prioritisetransaction.py
JackPiri: I am not sure about improved readability and even more critical is the fact that in current implementation if the condition is met the loop breaks, but if the condition is not met another loop round takes place (conversely, asserting would result in test failure).
I'd keep this part as it is, unless also another team member suggests to modify as per your comment (being in minority in that case :grin:).

qa/rpc-tests/prioritisetransaction.py
JackPiri: I agree we can take chance of this refactoring/fixing to increase the test documentation; added a bunch of lines at the start of `run_test` method.

a-petrini (approved):
Nothing to  add, PR looks good to me!

ptagl (approved):

PR #540, opened 2023-04-20
Ap/zend 276

a-petrini:
This pull request follows https://github.com/bitcoin/bitcoin/pull/7692 and removes the p2p alert system.


qa/rpc-tests/blockdelay_2.py
```diff
@@ -19,14 +19,9 @@
 import time
 class blockdelay_2(BitcoinTestFramework):
 
-    alert_filename = None
```
JackPiri, line 22: This alert file code can be removed also from following Python modules: `blockdelay.py` (not used at all, being in comment section) and `forknotify.py` (passed to nodes as startup parameter, but actually never used).

Also, `alert_filename = None` can be removed from `p2p_ignore_spent_tx.py` Python module.

src/main.cpp
```diff
@@ -2160,6 +2158,24 @@ bool fLargeWorkForkFound = false;
 bool fLargeWorkInvalidChainFound = false;
 CBlockIndex *pindexBestForkTip = NULL, *pindexBestForkBase = NULL;
 
+void alertNotify(const std::string& strMessage, bool fThread) {
```
JackPiri, line 2161: I strongly suggest to name this function as `internalAlertNotify` or even `internalAlarmNotify` (or something similar) in order to make explicit this is not related to old alert system (messages received from P2P network) but rather to internal notification logic.

src/init.cpp
```diff
@@ -342,7 +340,6 @@ std::string HelpMessage(HelpMessageMode mode)
 
     string strUsage = HelpMessageGroup(_("Options:"));
     strUsage += HelpMessageOpt("-?", _("This help message"));
-    strUsage += HelpMessageOpt("-alerts", strprintf(_("Receive and display P2P network alerts (default: %u)"), DEFAULT_ALERTS));
     strUsage += HelpMessageOpt("-alertnotify=<cmd>", _("Execute command when a relevant alert is received or we see a really long fork (%s in cmd is replaced by message)"));
```
JackPiri: Similarly to comment related to function `alertNotify`, I'd rephrase the description of `-alertnotify` parameter to something like `"Execute command when a relevant internal alert is raised (%s in cmd is replaced by message)"` (or something similar).

src/test/alert_tests.cpp
```diff
@@ -3,385 +3,19 @@
...
-    BOOST_CHECK(!alerts[1].AppliesTo(999002, "/MagicBean:0.1.0/"));
-    BOOST_CHECK(!alerts[1].AppliesTo(1, "/MagicBean:0.2.0/"));
-
-    BOOST_CHECK(!alerts[2].AppliesTo(1, "/MagicBean:0.3.0/"));
-
-    SetMockTime(0);
-}
-
-
-BOOST_AUTO_TEST_CASE(AlertNotify)
```
JackPiri, line 324: Since we are keeping a portion of this file, should the test `AlertNotify` be kept (adapting it to use `alertNotify` method)?

src/Makefile.test.include
```diff
@@ -36,7 +36,7 @@ JSON_TEST_FILES = \
   test/data/g1_compressed.json \
   test/data/g2_compressed.json
 
-#RAW_TEST_FILES = test/data/alertTests.raw
```
JackPiri, line 39: It seems to me variables `RAW_TEST_FILES` and `GENERATED_TEST_FILES` can be dropped, correct?

JackPiri (approved):
The PR is fine to me.
I left a few comments for cleaning-up some unused/leftover code, minor importance.
Additionally, please have a look at this comment https://github.com/HorizenOfficial/zen/pull/540/files#r1182726188: I think some renaming should be performed (wherever the word _alert_ is mentioned) in order to make totally clear old alert system has definitely gone for good.
Thanks

src/main.cpp
```diff
@@ -2160,6 +2158,24 @@ bool fLargeWorkForkFound = false;
...
+    // Alert text should be plain ascii coming from a trusted source, but to
+    // be safe we first strip anything not in safeChars, then add single quotes around
+    // the whole string before passing it to the shell:
+    std::string singleQuote("'");
+    std::string safeStatus = SanitizeString(strMessage);
+    safeStatus = singleQuote+safeStatus+singleQuote;
+    boost::replace_all(strCmd, "%s", safeStatus);
+
+    if (fThread)
+        boost::thread t(runCommand, strCmd); // thread runs free
```
drgora: I'd suggest replacing these `boost` (also at line 2171) functions with `std` alternatives, considering it should be low effort.

drgora (approved):
I left one non-blocking comment, but the PR is good by me.

qa/rpc-tests/blockdelay_2.py
a-petrini: Thanks, removed from `block_delay.py` and `p2p_ignore_spent_tx.py`. Not sure about `forknotify.py`, but I agree that there are no explicit tests on the contents of `alert.txt` file, once it has been opened and - possibly - written to.

src/Makefile.test.include
a-petrini: Yes, but only `RAW_TEST_FILES`. Wanted to follow a comment on Bitcoin PR, but I do not see any purpose for keeping this one here.

src/init.cpp
a-petrini: Agree, the help string has been rephrased.

src/main.cpp
a-petrini: While I agree that this function is triggered by an internal condition, rather than from a broadcasted message, I would keep the name as it is, to match Bitcoin code.

src/main.cpp
a-petrini: Yes, I modified the function to accommodate to `std::regex_replace` instead of `boost::replace_all`, and `std::thread` instead of `boost::thread`.

src/test/alert_tests.cpp
a-petrini: Yes, I added a Boost test that checks the function proper behavior.

src/main.cpp
drgora: Great, thanks!

ptagl (approved):

PR #541, opened 2023-04-21
Gp/support indexing for p2pk

JackPiri:
This PR adds support for Pay-To-Public-Key (P2PK) scripts; before this PR only Pay-To-Public-Key-Hash (P2PKH) and Pay-To-Script-Hash (P2SH) scripts were supported.
PR associated to issue [#352](https://github.com/HorizenOfficial/zen/issues/352).
Notice: this is a rebased clone of PR [#480](https://github.com/HorizenOfficial/zen/pull/480).


drgora (approved):
Looks good to me

src/main.h
```diff
@@ -46,6 +46,9 @@ class CValidationState;
 class CTxUndo;
 struct CNodeStateStats;
 class CTxInUndo;
+#ifdef ENABLE_ADDRESS_INDEXING
```
a-petrini: This define has been removed from the codebase with PR https://github.com/HorizenOfficial/zen/pull/531 so it should not be introduced back.

src/rpc/blockchain.cpp
```diff
@@ -3,7 +3,9 @@
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
+#ifdef ENABLE_ADDRESS_INDEXING
```
a-petrini: As in `main.h`, this define has been removed from the codebase with PR https://github.com/HorizenOfficial/zen/pull/531 so it should not be introduced back.

src/rpc/misc.cpp
```diff
@@ -3,7 +3,9 @@
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
+#ifdef ENABLE_ADDRESS_INDEXING
```
a-petrini: As in `main.h`, this define has been removed from the codebase with PR https://github.com/HorizenOfficial/zen/pull/531 so it should not be introduced back.

a-petrini (changes requested):

src/rpc/blockchain.cpp
JackPiri: Yes, this is a wrong rebasing; modified accordingly.

src/main.h
JackPiri: Yes, this is a wrong rebasing; modified accordingly.

src/rpc/misc.cpp
JackPiri: Yes, this is a wrong rebasing; modified accordingly.

src/main.h
a-petrini: Nice, thanks!

a-petrini (approved):

ptagl (approved):

PR #542, opened 2023-04-24
Fix for LevelDB file usage

JackPiri:
- Set maxopenfiles default to 400
- A command line parameter is provided for setting maxopenfiles explicitly for each database (blocktree and coinsview)

### DIFF
--- a/src/leveldbwrapper.h
+++ b/src/leveldbwrapper.h
@@ -16,6 +16,24 @@
 #include <leveldb/db.h>
 #include <leveldb/write_batch.h>
 
+// https://github.com/bitcoin/bitcoin/pull/12495
+// On most platforms the default setting of max_open_files (which is 1000)
+// is optimal. On Windows using a large file count is OK because the handles
+// do not interfere with select() loops. On 64-bit Unix hosts this value is
+// also OK, because up to that amount LevelDB will use an mmap
+// implementation that does not use extra file descriptors (the fds are
+// closed after being mmaped).
+// Increasing the value beyond the default is dangerous because LevelDB will
+// fall back to a non-mmap implementation when the file count is too large.
+//
+// The following default value has been chosen, considering that:
+// 1. we want zend to use at most 1024 FD
+// 2. with default config, zend approximately uses 125 FD for connections,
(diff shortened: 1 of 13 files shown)


drgora (approved):
Looks good to me

a-petrini (approved):
After last change, everything looks good to me

src/leveldbwrapper.h
```diff
@@ -16,6 +16,24 @@
...
 
+// https://github.com/bitcoin/bitcoin/pull/12495
+// On most platforms the default setting of max_open_files (which is 1000)
+// is optimal. On Windows using a large file count is OK because the handles
+// do not interfere with select() loops. On 64-bit Unix hosts this value is
+// also OK, because up to that amount LevelDB will use an mmap
+// implementation that does not use extra file descriptors (the fds are
+// closed after being mmaped).
+// Increasing the value beyond the default is dangerous because LevelDB will
+// fall back to a non-mmap implementation when the file count is too large.
```
cronicc, line 27: @JackPiri @ptagl 
Do we know at which point LevelDB falls back non-mmap implementation. Is there a recommendation for the maximum values to configure e.g. for high load environments like the block explorer? Will performance get worse if too high limits are configured?

drgora: Leveldb is set to use at most `1000` mmaped files. This value is global, so these 1000 files are shared by all leveldb databases in use by zend. There is no clear recommendation on the maximum values to use, and it is definitely not easy to predict performance with different configurations and on different systems.
In general, higher limits gives leveldb more flexibility in resource handling, but we have not investigated leveldb's logic in depth to draw accurate conclusions. As you can imagine, more resource usage could have unpredictable consequences on different systems. We have chosen the defaults in this PR after direct experimentation with systems that should be comparable to typical systems that we expect to run zend.

Notice that with our current codebase I highly discourage increasing these limits beyond what we have set in this PR, at least on Linux. This is not really for performance concerns, but for other limitations that we currently have in our codebase: having more than 1024 open file descriptors in our zend process can have negative impact on the net code in terms of functionality (e.g. failures establishing new connections).

PR #544, opened 2023-05-08
Python tests fixes

JackPiri:
This PR fixes two Python tests:

- shieldedpooldeprecation_rpc.py: due to tx-hash-based selection of utxos/notes, z_mergetoaddress output can change from run to run, eventually incurring in insufficiently mature funds; this has been fixed by making funds mature earlier (no performance impact),

- wallet_mergetoaddress_2.py: after updating the chain setup archive, an assertion with previously ad hoc defined value was failing; this has been fixed updating that value.


a-petrini (approved):
PR looks good

drgora (approved):
This is fine by me, thanks

ptagl (approved):

PR #545, opened 2023-05-09
Update main branch version

drgora:
This PR updates the software version on the `main` branch to clarify that `main` contains software being actively developed, which does not represent a stable release.


JackPiri (approved):
The PR is fine to me.
Thanks

a-petrini (approved):
PR looks good, nothing to report

README.md
```diff
@@ -1,4 +1,4 @@
-Zen 3.3.1
+Zen 4.0.99
```
ptagl, line 1: I wonder if this version shouldn't be `4.1.99` considering that we are currently working on features planned for `v4.1.0`.

What do you think?

ptagl (changes requested):

README.md
drgora: That's a very interesting point. I think of what we have in main now as what will eventually be included in version `4.1`, so it is something that happens before `4.1`. This is my reasoning for preferring `4.0.99`, like other projects do.

On the other hand I believe both forms are generally used, and I see that bitcoin is using the `4.1.99` scheme, so that might be a point in favor of it.

All in all, I don't see any strong point in favor of either option. If we all agree, I can change it to `4.1.99`.

ptagl: I didn't consider this perspective at the beginning but it definitely makes sense: it feels more "natural" switching from `4.0.99` to `4.1.0` when creating the next release compared to switching from `4.1.99` back to `4.1.0`.

It's still possible, however, that we will end up having to release `4.0.1` or `4.0.2`, but typically the release preparation should start from `v4.0.x` branch with some eventual cherry-picks.

All in all, I like the proposed approach, let's keep it as it is, thanks!

ptagl (approved):

PR #546, opened 2023-05-09
Adapt repository for incremental release notes

JackPiri:
This PR adapt repository for incremental release notes:

- release note template file "./doc/release-notes.md" has been removed
- current release note file has been created and populated "./doc/release-notes/release-notes-current.md"

In this particular case, the current release note file has been populated starting from main branch commit f3e17aa67d0c62287604d31c15f1c55df350e671 and excluding:

- e33877ec5579f10b848f1735cd49da7a7ba1868d [#532] (because this PR targets only an internal tool used for release preparation)
- 54f5020ae7c2b211a58792f12e635f7e2399eca2 [#542] (because this PR is already included in 4.0.0-rc3)

### DIFF
--- a/doc/release-notes/release-notes-current.md
+++ b/doc/release-notes/release-notes-current.md
@@ -1,0 +1,26 @@
+zend v4.0.99
+=========
+
+## Important notes
+- PR [#526](https://github.com/HorizenOfficial/zen/pull/526) implements [ZenIP-42204](https://github.com/HorizenOfficial/ZenIPs/blob/master/zenip-42204.md), which introduces a hard fork disabling the possibility to move transparent funds to shield addresses. After the hard fork, only shielded-to-shielded and shielded-to-transparent transactions will be allowed. For this reason, the following RPC commands have been (partially or fully) deprecated: `z_sendmany`, `z_shieldcoinbase` and `z_mergetoaddress` (check inline documentation for additional details).
+- PR [#539](https://github.com/HorizenOfficial/zen/pull/541) modifies the data structure returned by RPC method `getrawmempool` [#539](https://github.com/HorizenOfficial/zen/pull/539)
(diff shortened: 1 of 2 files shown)


a-petrini (approved):
Looks good!
Thanks

doc/release-notes/release-notes-current.md
```diff
@@ -0,0 +1,25 @@
+zend v4.0.99
+=========
+
```
ptagl, line 3: Given that in this release we introduced a couple of changes potentially having quite an impact on users, I think it could be useful to add some notes to make it more explicit what they could expect from the next release, for instance something like this:

```suggestion

Important notes:
- PR [#541](https://github.com/HorizenOfficial/zen/pull/541) fixes an issue that made Explorers, relying on `zend` RPC commands, not to show old transactions including P2PK scripts. The fix will have an effect only on blocks processed after the update, so if you are running an explorer you need to run a (fast) reindex to fix how the old blocks affected are shown. However, reindexing is not mandatory.
- PR [#526](https://github.com/HorizenOfficial/zen/pull/526) implements [ZenIP-42204](https://github.com/HorizenOfficial/ZenIPs/blob/master/zenip-42204.md), which introduces a hard fork disabling the possibility to move transparent funds to shield addresses. After the hard fork, only shielded-to-shielded and shielded-to-transparent transactions will be allowed. For this reason, some RPC commands related to shielded addresses have been deprecated and may partially or completely stop working after the hard fork.

```

ptagl (changes requested):

doc/release-notes/release-notes-current.md
JackPiri: Ok aligned based on your suggestion, with minor revisions and changes,

drgora (approved):

ptagl (approved):

PR #547, opened 2023-05-10
Ap/sync mempools improvement

a-petrini:
This PR modifies a few Python tests by adding a flexible way of dividing the node network.

Many test scenarios involve network splits and joins, and usually the last (ordinal) node among those launched by the test plays the role of isolated / dishonest actor.
This PR modifies these tests so that any node can play the role of isolated / dishonest node.

### DIFF
--- a/qa/rpc-tests/checkblockatheight.py
+++ b/qa/rpc-tests/checkblockatheight.py
@@ -34,11 +34,11 @@ MAIN_NODE = HONEST_NODES[0]                 # Hardcoded alias, do not change
 
 class checkblockatheight(BitcoinTestFramework):
 
-    def setup_chain(self, split=False):
+    def setup_chain(self):
         print("Initializing test directory "+self.options.tmpdir)
         initialize_chain_clean(self.options.tmpdir, NUMB_OF_NODES)
 
-    def setup_network(self, split=False, minAge=FINALITY_MIN_AGE):
+    def setup_network(self, minAge=FINALITY_MIN_AGE):
         assert_equal(MAIN_NODE, HONEST_NODES[0])
 
         self.nodes = []
@@ -52,7 +52,7 @@ class checkblockatheight(BitcoinTestFramework):
 
         for idx in range(NUMB_OF_NODES - 1):
             connect_nodes_bi(self.nodes, idx, idx + 1)
-        self.is_network_split = split
+        self.is_network_split = False
         self.sync_all()
 
     def split_network(self):
@@ -269,7 +269,7 @@ class checkblockatheight(BitcoinTestFramework):
 
         stop_nodes(self.nodes)
         wait_bitcoinds()
-        self.setup_network(False, 0)
+        self.setup_network(minAge=0)
 
         chunks = 305
         self.mark_logs(f"  Node {MAIN_NODE} generates {chunks} blocks")
@@ -446,7 +446,7 @@ class checkblockatheight(BitcoinTestFramework):
         # restart Nodes and check their balance, at this point the 1000 coins should be in the wallet of node2
         stop_nodes(self.nodes)
         wait_bitcoinds()
-        self.setup_network(False, 0)
+        self.setup_network(minAge=0)
 
         final_bal1 = Decimal(self.nodes[HONEST_NODES[1]].getbalance() )
         final_bal2 = Decimal(self.nodes[HONEST_NODES[2]].getbalance() )
--- a/qa/rpc-tests/checkblockatheight.py
+++ b/qa/rpc-tests/checkblockatheight.py
@@ -5,24 +5,17 @@
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 from test_framework.test_framework import BitcoinTestFramework
 from test_framework.authproxy import JSONRPCException
-from test_framework.script import OP_DUP, OP_EQUALVERIFY, OP_HASH160, OP_EQUAL, hash160, OP_CHECKSIG, OP_CHECKBLOCKATHEIGHT
+from test_framework.script import OP_DUP, OP_EQUALVERIFY, OP_HASH160, OP_CHECKSIG, OP_CHECKBLOCKATHEIGHT
 from test_framework.util import assert_equal, assert_greater_than, bytes_to_hex_str, initialize_chain_clean, \
-    start_nodes, start_node, connect_nodes, stop_node, stop_nodes, \
-    sync_blocks, sync_mempools, connect_nodes_bi, wait_bitcoinds, p2p_port, check_json_precision, disconnect_nodes
+    start_nodes, stop_nodes, sync_blocks, sync_mempools, connect_nodes_bi, wait_bitcoinds, disconnect_nodes
 from test_framework.script import CScript
 from test_framework.mininode import CTransaction, ToHex
 from test_framework.util import hex_str_to_bytes, swap_bytes
-import traceback
 from binascii import unhexlify
 from io import BytesIO
-import os,sys
-import shutil
+import os
 from decimal import Decimal
-import binascii
-import codecs
-import pprint
 
-import time
 
 NUMB_OF_NODES = 4
 
@@ -34,6 +27,11 @@ FINALITY_MIN_AGE = 75
 
 # mainchain uses this value for targeting a blockhash when building cbh script starting from chain tip backwards
 CBH_DELTA_HEIGHT = 300
+
+DISHONEST_NODE_INDEX = NUMB_OF_NODES - 1    # Dishonest node
+HONEST_NODES = list(range(NUMB_OF_NODES))
+HONEST_NODES.remove(DISHONEST_NODE_INDEX)
+MAIN_NODE = HONEST_NODES[0]
 
 class checkblockatheight(BitcoinTestFramework):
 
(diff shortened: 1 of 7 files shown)


qa/rpc-tests/checkblockatheight.py
```diff
@@ -267,18 +272,18 @@ def run_test(self):
...
-        blocks.extend(self.nodes[0].generate(1)) 
+        self.mark_logs(f"  Node {MAIN_NODE} generating 1 honest block")
+        blocks.extend(self.nodes[MAIN_NODE].generate(1)) 
         sync_blocks(self.nodes, 1, False, 5)
 
-        h_current = self.nodes[1].getblockcount()
+        h_current = self.nodes[HONEST_NODES[1]].getblockcount()
 
         # we will perform on attack aimed at reverting from this block (latest generated) upward
         h_attacked = h_current
```
JackPiri, line 289: Let's take this chance to get rid of the `;` at line 290.

qa/rpc-tests/checkblockatheight.py
```diff
@@ -52,33 +50,41 @@ def setup_network(self, split=False, minAge=FINALITY_MIN_AGE):
...
+        if idx == NUMB_OF_NODES - 1:
+            connect_nodes_bi(self.nodes, idx, idx - 1)
+        elif idx == 0:
+            connect_nodes_bi(self.nodes, idx, idx + 1)
+        else:
+            disconnect_nodes(self.nodes[idx - 1], idx + 1)
+            disconnect_nodes(self.nodes[idx + 1], idx - 1)
+            connect_nodes_bi(self.nodes, idx, idx - 1)
+            connect_nodes_bi(self.nodes, idx, idx + 1)
         sync_blocks(self.nodes, 1, False, 5)
```
JackPiri: I'm thinking about the usage of `sync_blocks(self.nodes, 1, False, 5)` limiting wait until only 5 loops; I know it's something not directly introduced by this PR but maybe it is worth double checking if it's not better to leave unlimited wait (and rely on CI timeout).
Same comment applies also to line 284.

qa/rpc-tests/checkblockatheight.py
```diff
@@ -35,6 +28,11 @@
 # mainchain uses this value for targeting a blockhash when building cbh script starting from chain tip backwards
 CBH_DELTA_HEIGHT = 300
 
+DISHONEST_NODE_INDEX = NUMB_OF_NODES - 1    # Dishonest node
+HONEST_NODES = list(range(NUMB_OF_NODES))
+HONEST_NODES.remove(DISHONEST_NODE_INDEX)
+MAIN_NODE = HONEST_NODES[0]
```
JackPiri: I've tested changing `MAIN_NODE` value to `HONEST_NODES[1]` or `HONEST_NODES[2]` and found the test is failing.
I see some options here:

- put an assert at test startup, checking `MAIN_NODE` equals `HONEST_NODES[0]` (probably the best option, zero effort),
- changing the logic/usage of `HONEST_NODES` to something like `HONEST_NODES_BUT_MAIN`,
- providing a simple inline function that extract the n-th `HONEST_NODE` different from `MAIN_NODE` (and use it in the appropriate places).

qa/rpc-tests/sc_ft_and_mbtr_fees_update.py
```diff
@@ -23,13 +22,6 @@
...
 NUM_TXS       = 10
 
-#### temporary logs functions
-def cc(c, inp):
-    return inp
-def mark_logs_temp(msg, nodes, debug = 0, color = 'n'):
-    return mark_logs(msg, nodes, debug)
-####
-
 class provaFee(BitcoinTestFramework):
```
JackPiri: Just seen this accidentally, probably we want to take this chance to rename `provaFee` class (being the file under refactoring).

qa/rpc-tests/sc_cert_ceasing_split.py
```diff
@@ -204,111 +212,104 @@ def run_test(self):
...
             any_error = True
 
 
-        mark_logs("And that no info are available too...".format(tx_fwd), self.nodes, DEBUG_MODE)
+        mark_logs("And that no info are available too...", self.nodes, DEBUG_MODE)
         try:
-            dec = self.nodes[0].getrawtransaction(tx_fwd, 1)
-            print("FIX FIX FIX!!! tx_fwd has info in Node0" )
+            dec = self.nodes[MAIN_NODE].getrawtransaction(tx_fwd, 1)
+            print(f"FIX FIX FIX!!! tx_fwd has info in Node {self.nodes[MAIN_NODE]}" )
```
JackPiri: This was already bugged (and actually not hit due to exception), should be `print(f"FIX FIX FIX!!! tx_fwd has info in Node {MAIN_NODE}" )`.
Same comment applies at lines 279 and 294.

qa/rpc-tests/sc_cert_memcleanup_split.py
```diff
@@ -38,31 +44,42 @@ def setup_network(self, split=False):
...
+            connect_nodes_bi(self.nodes, idx, idx - 1)
+        elif idx == 0:
+            connect_nodes_bi(self.nodes, idx, idx + 1)
+        else:
+            disconnect_nodes(self.nodes[idx - 1], idx + 1)
+            disconnect_nodes(self.nodes[idx + 1], idx - 1)
+            connect_nodes_bi(self.nodes, idx, idx - 1)
+            connect_nodes_bi(self.nodes, idx, idx + 1)
         self.is_network_split = False
+        time.sleep(2)
```
JackPiri: Is this `time.sleep(2)` really needed? Can it be replaced with some sync utility method?
Same comment applies to `sc_csw_memcleanup_split.py`.

qa/rpc-tests/sc_csw_nullifier.py
```diff
@@ -38,29 +42,40 @@ def setup_network(self, split=False):
...
+            connect_nodes_bi(self.nodes, idx, idx+1)
+        self.is_network_split = False
         self.sync_all()
 
     def split_network(self):
-        # Split the network of three nodes into nodes 0-1 and 2.
-        assert not self.is_network_split
-        disconnect_nodes(self.nodes[1], 2)
-        disconnect_nodes(self.nodes[2], 1)
+        # Disconnect the "double spend" node from the network
```
JackPiri: This comment should be `# Disconnect the "to be isolated" node from the network` (probably a copy-paste from `mempool_double_spend.py`).

JackPiri (approved):
The PR is fine to me. I left some minor comments, feel free to address or ignore them (some are very pedantic, but since we are refactoring the code I opted to report them anyway).
One last generic comment: since `MAIN_NODE` and `HONEST_NODES`/`NET_NODES` have been introduced to make the test more flexible and generic, a renaming of node-hardcoded variables can be considered (e.g. `amount_node1` in `checkblockatheight.py` or `node0_address` in `mempool_double_spend.py`, ecc).
Thanks

qa/rpc-tests/checkblockatheight.py
a-petrini: Well spotted! Also took a chance to remove another sneaky one from `sc_cert_ceasing_split.py`

qa/rpc-tests/checkblockatheight.py
a-petrini: Yes, agree. I revised the `sync_blocks` mechanism on the whole test, and replaced the timed version with non-timed one throughout the test

qa/rpc-tests/sc_ft_and_mbtr_fees_update.py
a-petrini: Yes, this had been left in from original development. Replaced with a less temporary name.

qa/rpc-tests/sc_cert_ceasing_split.py
a-petrini: Correct, all the instances have been replaced

qa/rpc-tests/sc_cert_memcleanup_split.py
a-petrini: Removed the timed delay from both tests and replaced with `sync_blocks()` to reduce the chance of sporadic failures

qa/rpc-tests/sc_csw_nullifier.py
a-petrini: Yes, corrected

qa/rpc-tests/checkblockatheight.py
a-petrini: I agree: this refactor aimed at fixing the indexing of the dishonest / isolated node, but the `MAIN_NODE` alias introduces a constraint on this role. As now, I followed the first suggestion and added a note + an assert to prevent tests to fail in an unpredictable manner.

qa/rpc-tests/checkblockatheight.py
```diff
@@ -35,15 +27,21 @@
...
+HONEST_NODES.remove(DISHONEST_NODE_INDEX)
+MAIN_NODE = HONEST_NODES[0]                 # Hardcoded alias, do not change
+
 class checkblockatheight(BitcoinTestFramework):
 
     def setup_chain(self, split=False):
         print("Initializing test directory "+self.options.tmpdir)
         initialize_chain_clean(self.options.tmpdir, NUMB_OF_NODES)
 
     def setup_network(self, split=False, minAge=FINALITY_MIN_AGE):
```
drgora: Given that the `split` parameter is not used anymore, I'd suggest to remove it altogether from the signature (if possible), to prevent wrong expectations on the result of the function on the caller side.
Another option would be to `assert_equal(split, False)`.

The same applies also to other tests in this PR.

drgora (changes requested):
I left one comment. Once that is solved, this PR is ok for merge by me.

qa/rpc-tests/checkblockatheight.py
a-petrini: Agree with this suggestion, thanks.
I updated all the tests and removed all the unnecessary parameters and variables

drgora: Great, thank you!

drgora (approved):

a-petrini:
After PRs https://github.com/HorizenOfficial/zen/pull/563 and https://github.com/HorizenOfficial/zen/pull/586 , this one needs a major refactor. As a side effect, it increases the level of complexity of the affected tests, adding only a minor improvement in flexibility. For this reasons, I suggest to close this PR without merging into main.

PR #554, opened 2023-05-18
fReindexFast and fReindex change to atomic<bool>

titusen:
Due to thread safety these two variables are changed atomic.


titusen:
Incorrect git gpg certifiacte was used to commit

PR #559, opened 2023-05-24
[ZEND-517] Unnecessary log lines removal

titusen:
As above

### DIFF
--- a/src/clientversion.cpp
+++ b/src/clientversion.cpp
@@ -102,14 +102,25 @@ const std::string CLIENT_DATE(BUILD_DATE);
 
 std::string FormatVersion(int nVersion)
 {
-    if (nVersion % 100 < 25)
-        return strprintf("%d.%d.%d-beta%d", nVersion / 1000000, (nVersion / 10000) % 100, (nVersion / 100) % 100, (nVersion % 100)+1);
-    if (nVersion % 100 < 50)
+    if (nVersion % 100 == 0) 
+    {
+        return strprintf("%d.%d.%d-alpha", nVersion / 1000000, (nVersion / 10000) % 100, (nVersion / 100) % 100);
+    } 
+    else if (nVersion % 100 < 25)
+    {
+        return strprintf("%d.%d.%d-beta%d", nVersion / 1000000, (nVersion / 10000) % 100, (nVersion / 100) % 100, (nVersion % 100));
+    } 
+    else if (nVersion % 100 < 50)
+    {
         return strprintf("%d.%d.%d-rc%d", nVersion / 1000000, (nVersion / 10000) % 100, (nVersion / 100) % 100, (nVersion % 100)-24);
-    else if (nVersion % 100 == 50)
+    } 
+    else if (nVersion % 100 == 50) 
+    {
         return strprintf("%d.%d.%d", nVersion / 1000000, (nVersion / 10000) % 100, (nVersion / 100) % 100);
-    else
+    } 
+    else {
         return strprintf("%d.%d.%d-%d", nVersion / 1000000, (nVersion / 10000) % 100, (nVersion / 100) % 100, (nVersion % 100)-50);
+    }
 }
 
 std::string FormatFullVersion()
--- a/src/metrics.cpp
+++ b/src/metrics.cpp
@@ -12,6 +12,7 @@
 #include "utiltime.h"
 #include "utilmoneystr.h"
 #include "utilstrencodings.h"
+#include "clientversion.h"
 
 #include <boost/thread.hpp>
 #include <boost/thread/synchronized_value.hpp>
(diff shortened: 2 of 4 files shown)


JackPiri (changes requested):
Here it is better using already available function `FormatVersion`, since it makes this in a more concise way and it formats the build number as a suffix.
Please note `FormatVersion` should be fixed because it doesn't handle `CLIENT_VERSION_BUILD=0` case and because it is bugged on beta builds (adding a `+1`).

titusen:
> 

Giacomo, I don't think this log line is not needed at all cause the full version with commit is printed here https://github.com/HorizenOfficial/zen/blob/f53067edd84352917e32ee92eb0b74ccdf9e96fd/src/init.cpp#L960

src/clientversion.cpp
```diff
@@ -103,7 +103,7 @@ const std::string CLIENT_DATE(BUILD_DATE);
 std::string FormatVersion(int nVersion)
 {
     if (nVersion % 100 < 25)
-        return strprintf("%d.%d.%d-beta%d", nVersion / 1000000, (nVersion / 10000) % 100, (nVersion / 100) % 100, (nVersion % 100)+1);
```
JackPiri, line 106: Additionally also the case of build number being 0 should be handled (since this is the build number for `main` branch).
Something like:
```
...
if (nVersion % 100 == 0)
  return strprintf("%d.%d.%d-alpha", nVersion / 1000000, (nVersion / 10000) % 100, (nVersion / 100) % 100);
if (nVersion % 100 > 0 && nVersion % 100 < 25)
...
```

src/metrics.cpp
```diff
@@ -517,9 +518,10 @@ void ThreadShowMetricsScreen()
 
         // Thank you text
         std::cout << _("Zen is economic freedom. Thanks for running a node.") << std::endl;
-        std::cout << _("仕方が無い") << std::endl;
-        std::cout << _("Shikata ga nai.") << std::endl;
-        std::cout << _("它不能得到帮助") << std::endl << std::endl;
+        if (fPrintToConsole) {
```
JackPiri: Probably we could drop the `fPrintToConsole` condition, I'd like to have the version being printed to screen regardless of that option.

JackPiri (changes requested):
Left a couple of comments, nothing big but better to address in this PR.

JackPiri:
> > 
> 
> Giacomo, I don't think this log line is not needed at all cause the full version with commit is printed here
> 
> https://github.com/HorizenOfficial/zen/blob/f53067edd84352917e32ee92eb0b74ccdf9e96fd/src/init.cpp#L960

Yes, I agree with you. Just remove one among:

- `LogPrintf("Horizen version %s (%s)\n", FormatFullVersion(), CLIENT_DATE);`
- `LogPrintf("Zen version %s (%s)\n", FormatFullVersion(), CLIENT_DATE);`

src/clientversion.cpp
titusen: Done

src/metrics.cpp
titusen: Done

JackPiri (approved):
The PR is fine to me.
Thanks

JackPiri (commented):
Sorry I didn't notice the CI failure, `BOOST_AUTO_TEST_CASE(test_FormatSubVersion)` should be aligned accordingly.

a-petrini (approved):
PR looks good, thanks!

drgora (approved):
This is ok for me, thanks

src/test/util_tests.cpp
```diff
@@ -414,16 +414,16 @@ BOOST_AUTO_TEST_CASE(test_FormatSubVersion)
     std::vector<std::string> comments2;
     comments2.push_back(std::string("comment1"));
     comments2.push_back(std::string("comment2"));
-    BOOST_CHECK_EQUAL(FormatSubVersion("Test", 99900, std::vector<std::string>()), std::string("/Test:0.9.99-beta1/"));
-    BOOST_CHECK_EQUAL(FormatSubVersion("Test", 99924, std::vector<std::string>()), std::string("/Test:0.9.99-beta25/"));
+    BOOST_CHECK_EQUAL(FormatSubVersion("Test", 99900, std::vector<std::string>()), std::string("/Test:0.9.99-alpha/"));
+    BOOST_CHECK_EQUAL(FormatSubVersion("Test", 99924, std::vector<std::string>()), std::string("/Test:0.9.99-beta24/"));
```
ptagl, line 419: Not very important here, but what about also restoring the check for `beta1` to keep the original pattern of the test (checking both limits for any label like `rc1` and `rc2`)?

titusen: Done

src/metrics.cpp
```diff
@@ -517,9 +518,9 @@ void ThreadShowMetricsScreen()
 
         // Thank you text
         std::cout << _("Zen is economic freedom. Thanks for running a node.") << std::endl;
-        std::cout << _("仕方が無い") << std::endl;
-        std::cout << _("Shikata ga nai.") << std::endl;
-        std::cout << _("它不能得到帮助") << std::endl << std::endl;
+
+        std::cout << "Software Version: " << FormatVersion(CLIENT_VERSION);
+        std::cout << std::endl;
```
ptagl: Can we now merge all together on one line?

```suggestion
        std::cout << "Software Version: " << FormatVersion(CLIENT_VERSION) << std::endl;
```

ptagl (changes requested):

src/metrics.cpp
titusen: done

ptagl (approved):

PR #560, opened 2023-05-24
[ZEND-528] "cmake" addition to required deps

titusen:
CMake is now required to build the project

### DIFF
--- a/doc/dependencies.md
+++ b/doc/dependencies.md
@@ -16,3 +16,4 @@ These are the dependencies needed to build Zen on Linux.
 | [Libtool](https://www.gnu.org/software/libtool/) | [2.4.6] |
 | [pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/) | [0.29.2] |
 | [hexdump](https://www.kernel.org/pub/linux/utils/util-linux/) | [2.37.2] |
+| [cmake](https://cmake.org/download/) | [3.16.3] |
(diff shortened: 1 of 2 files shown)


drgora (changes requested):
Thanks for the PR. The suggested changes are ok for me, but I believe we should also update the table in `doc/dependencies.md`, and add `cmake` there.

doc/dependencies.md
```diff
@@ -16,3 +16,4 @@ These are the dependencies needed to build Zen on Linux.
 | [Libtool](https://www.gnu.org/software/libtool/) | [2.4.6] |
 | [pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/) | [0.29.2] |
 | [hexdump](https://www.kernel.org/pub/linux/utils/util-linux/) | [2.37.2] |
+| [cmake](https://cmake.org/download/) | [3.22.1] |
```
a-petrini: If this does not cause any issue or breaks any dependency build, I would lower the version requirement for CMake, as `apt` package manager on Ubuntu Focal by default installs version 3.16.3 (see https://packages.ubuntu.com/focal/devel/cmake). Thanks!

a-petrini (changes requested):

doc/dependencies.md
titusen: Done

drgora (approved):
Looks good to me, thanks

JackPiri (approved):
The PR is fine to me.
Thanks

a-petrini (approved):

ptagl (approved):

PR #561, opened 2023-05-26
[ZEND-415] Hardcoded dns look up limit set to 256 as in Bitcoin

titusen:
Porting from Bitcoin

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -1476,29 +1476,35 @@ void ThreadDNSAddressSeed()
     }
 
     const vector<CDNSSeedData> &vSeeds = Params().DNSSeeds();
+    const unsigned int nMaxIPs = 256;
+    const int nThreeDays = 3 * 24 * 3600;
+    const int nFourDays = 4 * 24 * 3600;
     int found = 0;
 
     LogPrintf("Loading addresses from DNS seeds (could take a while)\n");
 
-    BOOST_FOREACH(const CDNSSeedData &seed, vSeeds) {
+    for(const CDNSSeedData &seed : vSeeds)
+    {
         if (HaveNameProxy()) {
             AddOneShot(seed.host);
-        } else {
-            vector<CNetAddr> vIPs;
-            vector<CAddress> vAdd;
-            if (LookupHost(seed.host.c_str(), vIPs))
-            {
-                BOOST_FOREACH(const CNetAddr& ip, vIPs)
-                {
-                    int nOneDay = 24*3600;
-                    CAddress addr = CAddress(CService(ip, Params().GetDefaultPort()));
-                    addr.nTime = GetTime() - 3*nOneDay - GetRand(4*nOneDay); // use a random age between 3 and 7 days old
-                    vAdd.push_back(addr);
-                    found++;
-                }
-            }
-            addrman.Add(vAdd, CNetAddr(seed.name, true));
+            continue;
+        } 
+    
+        vector<CNetAddr> vIPs;
+        if (!LookupHost(seed.host.c_str(), vIPs, nMaxIPs))
+        {
+            continue;
         }
+
+        vector<CAddress> vAdd;
+        for(const CNetAddr& ip : vIPs)
+        {
+            vAdd.emplace_back(CService(ip, Params().GetDefaultPort()));
+            vAdd.back().nTime = GetTime() - nThreeDays - GetRand(nFourDays); // use a random age between 3 and 7 days old
+        }
+        
+        addrman.Add(vAdd, CNetAddr(seed.name, true));
+        found += vAdd.size();
     }
 
     LogPrintf("%d addresses found from DNS seeds\n", found);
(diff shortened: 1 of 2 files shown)


src/net.cpp
```diff
@@ -1480,21 +1480,23 @@ void ThreadDNSAddressSeed()
...
-                BOOST_FOREACH(const CNetAddr& ip, vIPs)
+                const unsigned int nMaxIPs = 256; 
+                for(const CNetAddr& ip : vIPs)
                 {
                     int nOneDay = 24*3600;
                     CAddress addr = CAddress(CService(ip, Params().GetDefaultPort()));
                     addr.nTime = GetTime() - 3*nOneDay - GetRand(4*nOneDay); // use a random age between 3 and 7 days old
                     vAdd.push_back(addr);
-                    found++;
+                    if (++found == nMaxIPs) break;
```
AndreaLanfranchi: This is actually not correct: if the threshold of 256 found items is reached the break statement will exit the inner loop `for(const CNetAddr& ip : vIPs)` but will continue the outer loop `for(const CDNSSeedData &seed : vSeeds)` which will eventually cause a pre-increment of `found` on next iteration causing the test for equality to never trigger again.
Need to explain which is the intention here: limit the **overall** number of found IPs or max number of IPs per seed ?

AndreaLanfranchi: Also ... let's help the compiler:

- `nOneDay` is used only as base to get the seconds in two and three days. Hence is actually not needed
- transform `3*nOneDay` in a const var (below nMaxIps) defined as `const uint32_t nThreeDays = 86400*3;`
- transform `4*nOneDay` in a const var (below nMaxIps) defined as `const uint32_t nFourDays = 86400*4;`

AndreaLanfranchi (changes requested):

src/net.cpp
AndreaLanfranchi: Reading better the intent of the implementation from [bitcoin code here](https://github.com/bitcoin/bitcoin/blob/master/src/net.cpp#L1498) it appears the intent of `nMaxIPs` is to limit the number of items inserted into `vIPs` by the function `LookupHost` hence has nothing to do with the overall number recorded in `found`

titusen: Which constraint is wanted? On one seed, global or both?

AndreaLanfranchi: As the code (linked above) says the constraint is on number of IPs per seed.

```
unsigned int nMaxIPs = 256; // Limits number of IPs learned from a DNS seed
```

src/net.cpp
```diff
@@ -1476,25 +1476,28 @@ void ThreadDNSAddressSeed()
...
-                    int nOneDay = 24*3600;
+                    ++found;
+                    const int nThreeDays = 3*24*3600;
+                    const int nFourDays = 4*24*3600;
                     CAddress addr = CAddress(CService(ip, Params().GetDefaultPort()));
-                    addr.nTime = GetTime() - 3*nOneDay - GetRand(4*nOneDay); // use a random age between 3 and 7 days old
+                    addr.nTime = GetTime() - nThreeDays - GetRand(nFourDays); // use a random age between 3 and 7 days old
                     vAdd.push_back(addr);
-                    found++;
                 }
```
AndreaLanfranchi: Sorry for being nitpicking but ... is it worth force an increment by one for each iteration when we already know that `found` will be incremented exactly by `vIPs.size()` ? 


```suggestion
                for(const CNetAddr& ip : vIPs)
                {
                    const int nThreeDays = 3*24*3600;
                    const int nFourDays = 4*24*3600;
                    CAddress addr = CAddress(CService(ip, Params().GetDefaultPort()));
                    addr.nTime = GetTime() - nThreeDays - GetRand(nFourDays); // use a random age between 3 and 7 days old
                    vAdd.push_back(addr);
                }
                found += vIPs.size();
```

src/net.cpp
```diff
@@ -1476,25 +1476,28 @@ void ThreadDNSAddressSeed()
...
-            if (LookupHost(seed.host.c_str(), vIPs))
+            if (LookupHost(seed.host.c_str(), vIPs, nMaxIPs))
             {
-                BOOST_FOREACH(const CNetAddr& ip, vIPs)
+                for(const CNetAddr& ip : vIPs)
                 {
-                    int nOneDay = 24*3600;
+                    ++found;
+                    const int nThreeDays = 3*24*3600;
+                    const int nFourDays = 4*24*3600;
```
AndreaLanfranchi: I do believe it doesn't make sense to initialize those constants inside the loop.
They get reinitialized at every loop (even if sometimes the compiler may be smart enough to optimize this out)
Consts that do not depend from loop values should be always initialized outside of it.

AndreaLanfranchi: Don't get mad at me but ... outside all the loops (pretty much as you did with nMaxIPs)

titusen: :D

drgora (approved):
Looks good to me, thanks

AndreaLanfranchi (approved):

a-petrini (approved):
PR looks good! Thanks!

ptagl (approved):

PR #563, opened 2023-06-06
[ZEND-557] Split_network join_network setup_network method refactoring

titusen:
Common Split_network method implentation move to "BitcoinTestFramework" class

### DIFF
--- a/qa/rpc-tests/blockdelay.py
+++ b/qa/rpc-tests/blockdelay.py
@@ -23,35 +23,6 @@ class blockdelay(BitcoinTestFramework):
         print("Initializing test directory "+self.options.tmpdir)
         initialize_chain_clean(self.options.tmpdir, 5)
 
-    def setup_network(self, split=False):
-        self.nodes = []
-
-        # -exportdir option means we must provide a valid path to the destination folder for wallet backups
-        ed0 = "-exportdir=" + self.options.tmpdir + "/node0"
-        ed1 = "-exportdir=" + self.options.tmpdir + "/node1"
-        ed2 = "-exportdir=" + self.options.tmpdir + "/node2"
-        ed3 = "-exportdir=" + self.options.tmpdir + "/node3"
-        self.nodes = start_nodes(4, self.options.tmpdir)
-
-
-        if not split:
-            connect_nodes_bi(self.nodes, 1, 2)
-            sync_blocks(self.nodes[1:3])
-            sync_mempools(self.nodes[1:3])
-
-        connect_nodes_bi(self.nodes, 0, 1)
-        connect_nodes_bi(self.nodes, 2, 3)
-        self.is_network_split = split
-        self.sync_all()
-
-    def split_network(self):
-        # Split the network of four nodes into nodes 0/1 and 2/3.
-        assert not self.is_network_split
-        disconnect_nodes(self.nodes[1], 2)
-        disconnect_nodes(self.nodes[2], 1)
-        self.is_network_split = True
-
-
     def join_network(self):
         #Join the (previously split) network halves together.
         assert self.is_network_split
@@ -155,7 +126,7 @@ class blockdelay(BitcoinTestFramework):
 # Node(3): [0]->..->[104]
 
         print("\n\nSplit network")
-        self.split_network()
+        self.split_network(1)
         print("The network is split")
 
         print("\nNode0 sends 3.0 coins to Node1\n")
--- a/qa/rpc-tests/blockdelay_2.py
+++ b/qa/rpc-tests/blockdelay_2.py
@@ -23,45 +23,19 @@ class blockdelay_2(BitcoinTestFramework):
         print("Initializing test directory "+self.options.tmpdir)
         initialize_chain_clean(self.options.tmpdir, 3)
 
-    def setup_network(self, split=False):
-        self.nodes = []
-
+    def setup_nodes(self, split=False):
         self.nodes = start_nodes(3, self.options.tmpdir)
-
-        connect_nodes_bi(self.nodes, 0, 1)
-        connect_nodes_bi(self.nodes, 1, 2)
-        #connect_nodes_bi(self.nodes, 0, 2)
-        self.is_network_split = split
-        self.sync_all()
-
-    def split_network(self):
-        # Split the network of three nodes into nodes 0 and 1/2.
-#        assert not self.is_network_split
-        disconnect_nodes(self.nodes[1], 0)
-        disconnect_nodes(self.nodes[0], 1)
-        #disconnect_nodes(self.nodes[0], 2)
-        #disconnect_nodes(self.nodes[2], 0)
-        self.is_network_split = True
         
-    def split_network_2(self):
-        # Split the network of three nodes into nodes 0, 1 and 2.
-        disconnect_nodes(self.nodes[1], 2)
-        disconnect_nodes(self.nodes[2], 1)
-
-
-    def join_network(self):
+    def join_network(self, id = 0):
         #Join the (previously split) network halves together.
-        assert self.is_network_split
-        connect_nodes_bi(self.nodes, 0, 1)
-        connect_nodes_bi(self.nodes, 1, 0)
-        
-        sync_blocks(self.nodes[1:2])
-        sync_mempools(self.nodes[1:2])
-        
-    def join_network_2(self):
-        connect_nodes_bi(self.nodes, 1, 2)
-        sync_blocks(self.nodes[2:3])
-        sync_mempools(self.nodes[2:3])
+        connect_nodes_bi(self.nodes, id, id + 1)
+        sync_blocks(self.nodes[id + 1: id + 2])
+        sync_mempools(self.nodes[id + 1:id + 2])
+
+    def split_network(self, id = 1):
+        # Split the network of between adjanced nodes in linear topology ep. nodes 0-1 and 2-3.
+        disconnect_nodes(self.nodes[id], id + 1)
+        disconnect_nodes(self.nodes[id + 1], id)
 
     def dump_ordered_tips(self, tip_list):
         sorted_x = sorted(tip_list, key=lambda k: k['status'])
@@ -174,7 +148,7 @@ class blockdelay_2(BitcoinTestFramework):
         
         print("\nFirst split. Best block hash is %s: "+first_main_blockhash)
         print("\n\nSplit network")
-        self.split_network()
+        self.split_network(0)
         print("The network is splitted")
         print("\n0 ... 1 <--> 2")
         self.get_network_info()
@@ -280,7 +254,7 @@ class blockdelay_2(BitcoinTestFramework):
         print("\nSecond split. Best block hash is %s: "+second_main_blockhash)
         
         print("\n\nSplit network again")
-        self.split_network_2()
+        self.split_network(1)
         print("The network is split again")
         print("\n0 ... 1 ... 2")
         self.get_network_info()
@@ -351,7 +325,7 @@ class blockdelay_2(BitcoinTestFramework):
 #                                                 +-->[32]...->[37]
 
         print("\nJoining network...")
-        self.join_network()
+        self.join_network(0)
         time.sleep(5)
         print("\nNetwork joined")
         self.get_network_info()
@@ -454,7 +428,7 @@ class blockdelay_2(BitcoinTestFramework):
             print("Best block hash for node "+str(i)+": "+self.nodes[i].getbestblockhash())
             
         print("\nJoining network again...")
-        self.join_network_2()
+        self.join_network(1)
         time.sleep(5)
         print("\nNetwork joined")
         self.get_network_info()
@@ -537,7 +511,7 @@ class blockdelay_2(BitcoinTestFramework):
         print("\nThird split. Best block hash is %s: "+self.nodes[1].getbestblockhash())
         
         print("\n\nSplit network again")
-        self.split_network_2()
+        self.split_network(1)
         print("The network is split again")
         print("\n0 <--> 1 ... 2")
         self.get_network_info()
@@ -594,7 +568,7 @@ class blockdelay_2(BitcoinTestFramework):
         assert self.nodes[0].getbestblockhash() != self.nodes[2].getbestblockhash()
         
         print("\nJoining network...")
-        self.join_network_2()
+        self.join_network(1)
         time.sleep(5)
         print("\nNetwork joined")
         self.get_network_info()
--- a/qa/rpc-tests/getchaintips.py
+++ b/qa/rpc-tests/getchaintips.py
@@ -6,11 +6,48 @@
 # Exercise the getchaintips API.  We introduce a network split, work
 # on chains of different lengths, and join the network together again.
 # This gives us two tips, verify that it works.
-
+import time
 from test_framework.test_framework import BitcoinTestFramework
-from test_framework.util import assert_equal
+from test_framework.util import assert_equal, stop_nodes, wait_bitcoinds, \
+    sync_blocks, sync_mempools, wait_bitcoinds, \
+    connect_nodes_bi
 
 class GetChainTipsTest (BitcoinTestFramework):
+
+    def setup_network(self, split = False):
+        self.setup_nodes()
+
+        # Connect the nodes as a "chain".  This allows us
+        # to split the network between nodes 1 and 2 to get
+        # two halves that can work on competing chains.
+        # If we joined network halves, connect the nodes from the joint
+        # on outward.  This ensures that chains are properly reorganised.
+        if not split:
+            connect_nodes_bi(self.nodes, 1, 2)
+            sync_blocks(self.nodes[1:3])
+            sync_mempools(self.nodes[1:3])
+        connect_nodes_bi(self.nodes, 0, 1)
+        connect_nodes_bi(self.nodes, 2, 3)
+        self.is_network_split = split
+        self.sync_all()
+       
+    def split_network(self):
+        """
+        Split the network of four nodes into nodes 0/1 and 2/3.
+        """
+        assert not self.is_network_split
+        stop_nodes(self.nodes)
+        wait_bitcoinds()
+        self.setup_network(True)
+
+    def join_network(self):
+        """
+        Join the (previously split) network halves together.
+        """
+        assert self.is_network_split
+        stop_nodes(self.nodes)
+        wait_bitcoinds()
+        self.setup_network(False)
 
     def run_test (self):
         BitcoinTestFramework.run_test (self)
@@ -25,6 +62,7 @@ class GetChainTipsTest (BitcoinTestFramework):
         self.split_network ()
         self.nodes[0].generate(10)
         self.nodes[2].generate(20)
+
         self.sync_all ()
 
         tips = self.nodes[1].getchaintips ()
@@ -43,7 +81,8 @@ class GetChainTipsTest (BitcoinTestFramework):
 
         # Join the network halves and check that we now have two tips
         # (at least at the nodes that previously had the short chain).
-        self.join_network ()
+        self.join_network()
+        time.sleep(2)
 
         tips = self.nodes[0].getchaintips ()
         assert_equal (len (tips), 2)
--- a/qa/rpc-tests/headers_01.py
+++ b/qa/rpc-tests/headers_01.py
@@ -23,38 +23,12 @@ class headers(BitcoinTestFramework):
         print("Initializing test directory "+self.options.tmpdir)
         initialize_chain_clean(self.options.tmpdir, 2)
 
-    def setup_network(self, split=False):
-        self.nodes = []
-
+    def setup_nodes(self):
         self.nodes = start_nodes(2, self.options.tmpdir,
             extra_args = [
                 ["-debug=1", "-logtimemicros=1"],
                 ["-debug=1", "-logtimemicros=1"]
             ])
-
-        if not split:
-            connect_nodes_bi(self.nodes, 0, 1)
-            sync_blocks(self.nodes[0:2])
-            sync_mempools(self.nodes[0:2])
-
-        self.is_network_split = split
-        self.sync_all()
-
-    def split_network(self):
-        # Split the network of two nodes into nodes 0 and 1.
-        assert not self.is_network_split
-        disconnect_nodes(self.nodes[0], 1)
-        disconnect_nodes(self.nodes[1], 0)
-        self.is_network_split = True
-
-
-    def join_network(self):
-        #Join the (previously split) network halves together.
-        assert self.is_network_split
-        connect_nodes_bi(self.nodes, 0, 1)
-        connect_nodes_bi(self.nodes, 1, 0)
-        self.sync_all()
-        self.is_network_split = False
 
     def dump_ordered_tips(self, tip_list):
         sorted_x = sorted(tip_list, key=lambda k: k['status'])
@@ -65,7 +39,6 @@ class headers(BitcoinTestFramework):
             else:
                 print(" ",y)
             c = 1
-
 
     def run_test(self):
         blocks = []
--- a/qa/rpc-tests/headers_04.py
+++ b/qa/rpc-tests/headers_04.py
@@ -23,35 +23,8 @@ class headers(BitcoinTestFramework):
         print("Initializing test directory "+self.options.tmpdir)
         initialize_chain_clean(self.options.tmpdir, 3)
 
-    def setup_network(self, split=False):
-        self.nodes = []
-
+    def setup_nodes(self):
         self.nodes = start_nodes(3, self.options.tmpdir)
-
-        if not split:
-            # 1 and 2 are joint only if split==false
-            connect_nodes_bi(self.nodes, 1, 2)
-            sync_blocks(self.nodes[1:3])
-            sync_mempools(self.nodes[1:3])
-
-        connect_nodes_bi(self.nodes, 0, 1)
-        self.is_network_split = split
-
-    def split_network(self):
-        # Split the network of three nodes into nodes 0-1 and 2.
-        assert not self.is_network_split
-        disconnect_nodes(self.nodes[1], 2)
-        disconnect_nodes(self.nodes[2], 1)
-        self.is_network_split = True
-
-    def join_network(self):
-        #Join the (previously split) network pieces together: 0-1-2
-        assert self.is_network_split
-        connect_nodes_bi(self.nodes, 1, 2)
-        connect_nodes_bi(self.nodes, 2, 1)
-        self.sync_all()
-        self.is_network_split = False
-
 
     def mark_logs(self, msg):
         self.nodes[0].dbg_log(msg)
@@ -90,7 +63,7 @@ class headers(BitcoinTestFramework):
 # Node(2): [0]->[1]
 
         print("\n\nSplit network")
-        self.split_network()
+        self.split_network(1)
         print("The network is split")
         self.mark_logs("The network is split")
 
@@ -131,7 +104,7 @@ class headers(BitcoinTestFramework):
         print("\n\nJoin network")
 #        raw_input("press enter to join the netorks..")
         self.mark_logs("Joining network")
-        self.join_network()
+        self.join_network_sync()
 
         time.sleep(2)
         print("\nNetwork joined\n") 
--- a/qa/rpc-tests/sbh_rpc_cmds.py
+++ b/qa/rpc-tests/sbh_rpc_cmds.py
@@ -48,22 +48,6 @@ class sbh_rpc_cmds(BitcoinTestFramework):
         self.is_network_split = split
         self.sync_all()
 
-    def split_network(self):
-        # Split the network of three nodes into nodes 0-1 and 2.
-        assert not self.is_network_split
-        disconnect_nodes(self.nodes[1], 2)
-        disconnect_nodes(self.nodes[2], 1)
-        self.is_network_split = True
-
-    def join_network(self):
-        # Join the (previously split) network pieces together: 0-1-2
-        assert self.is_network_split
-        connect_nodes_bi(self.nodes, 1, 2)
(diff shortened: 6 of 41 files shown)


qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -50,35 +52,29 @@ def setup_chain(self):
...
-        if not split:
-            connect_nodes_bi(self.nodes, 1, 2)
-            sync_blocks(self.nodes[1:3])
-            sync_mempools(self.nodes[1:3])
-
-        connect_nodes_bi(self.nodes, 0, 1)
-        connect_nodes_bi(self.nodes, 2, 3)
+        sync_blocks(self.nodes[1:self.numb_of_nodes])
+        sync_mempools(self.nodes[1:self.numb_of_nodes])
         self.is_network_split = split
```
ptagl: I like the idea of this PR and the approach of reducing the complexity of Python tests by making the connection functions of the framework more generic.

If I understand correctly, `setup_network()` still takes `split` as input parameter but the function always acts as it is always set to `False`. This may be a problem for some tests (like `qa/rpc-tests/sc_cert_addressindex.py` maybe) in which the network is split, then nodes are stopped and restarted and we expect to keep the previous state (split).

So, the question here is: can we find a way to keep using `self.is_network_split` to dynamically set up the network? In case, we would probably need to extend the current function with:
- a new input parameter for `setup_network()` to specify the node at the edge of the network (that is not connected to the next one)
- a class property to keep track of this node so that we can set up the network in the same way again in case of restart

Another possibility could be to completely get rid of `split` as input parameter, but these are just few ideas, let's wait for feedback also from the other reviewers and discuss what could be done next.

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -90,14 +86,21 @@ def sync_all(self):
...
+    def join_network(self, id = 1):
+        # Join the (previously split) network pieces together: 0-1
+        # assert self.is_network_split
+        connect_nodes_bi(self.nodes, id, id + 1)
+        connect_nodes_bi(self.nodes, id + 1, id)
+        # sync_blocks(self.nodes, 1, False, 5)
+        # sync_mempools(self.nodes[id:id + 1])
+        self.is_network_split = False
+
+    def join_network_sync(self, id = 1):
```
drgora, line 91: Just a minor thing here, but could we reuse `join_network` here, instead of reimplementing it?

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -90,14 +86,21 @@ def sync_all(self):
...
-        """
-        Join the (previously split) network halves together.
-        """
-        assert self.is_network_split
-        stop_nodes(self.nodes)
-        wait_bitcoinds()
-        self.setup_network(False)
+    def join_network(self, id = 1):
+        # Join the (previously split) network pieces together: 0-1
+        # assert self.is_network_split
```
drgora: Do we really need to comment this line out?
It looks safer to me with the `assert`, which might warn the user he is doing something unneeded

qa/rpc-tests/sc_cert_addressindex.py
```diff
@@ -33,35 +33,27 @@ def setup_chain(self):
...
         self.is_network_split = False
+    
+    def join_network(self):
+        """
+        Join the (previously split) network halves together.
+        """
+        assert self.is_network_split
+        stop_nodes(self.nodes)
+        wait_bitcoinds()
+        self.setup_network()
```
a-petrini: this function appears twice in this class, which one should we keep?

qa/rpc-tests/sc_quality_nodes.py
```diff
@@ -37,21 +37,6 @@ def setup_network(self, split=False):
         sync_blocks(self.nodes[1:NUMB_OF_NODES])
         sync_mempools(self.nodes[1:NUMB_OF_NODES])
         self.is_network_split = split
```
a-petrini: I haven't re-run this test, but I think that removing the `sync_all()` at the end of the `setup_network()` is not safe

qa/rpc-tests/sc_cert_memcleanup_split.py
```diff
@@ -250,7 +235,8 @@ def run_test(self):
 
         #============================================================================================
         mark_logs("\nJoining network", self.nodes, DEBUG_MODE)
-        self.join_network()
+        self.join_network(2)
+        time.sleep(2)
```
a-petrini, line 239: That's more a memo for me, actually, but I would like to investigate what is the purpose of this sleep and, eventually, if there could be a more reliable time-independent method for doing whatever it does, as it's present in other tests

titusen: There was a sleep in the join_network method of that class and without it the test tend to fail.

qa/rpc-tests/sc_quality_nodes.py
titusen: I removed this function, cause it can be inherited.

qa/rpc-tests/sc_cert_addressindex.py
titusen: Removed

qa/rpc-tests/sc_quality_nodes.py
a-petrini: Sorry, maybe my comment was a bit confusing, as I was referring to the last instruction of the `setup_network()` defined here: there was a `self.sync_all()` that has been removed, I don't know if it is intentional

titusen: The setup_network is derived from the base class. Now, it's only setup_node overidden.

a-petrini: Oh, I see! I was confused by the renamed function, thanks!

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -50,35 +51,27 @@ def setup_chain(self):
...
-        self.sync_all()
+        self.is_network_split = False
+        self.sync_all()       
 
-    def split_network(self):
-        """
-        Split the network of four nodes into nodes 0/1 and 2/3.
-        """
+    def split_network(self, id = 1):
+        # Split the network of between adjanced nodes in linear topology ep. nodes 0-1 and 2-3.
```
JackPiri, line 69: Possibly an assertion on `id` value can be added (e.g. for handling cases of out-of-range).
Similar comment applies to `join_network` and `join_network_sync`.

qa/rpc-tests/blockdelay.py
```diff
@@ -23,35 +23,9 @@ def setup_chain(self, split=False):
...
 
-    def setup_network(self, split=False):
-        self.nodes = []
-
-        # -exportdir option means we must provide a valid path to the destination folder for wallet backups
-        ed0 = "-exportdir=" + self.options.tmpdir + "/node0"
-        ed1 = "-exportdir=" + self.options.tmpdir + "/node1"
-        ed2 = "-exportdir=" + self.options.tmpdir + "/node2"
-        ed3 = "-exportdir=" + self.options.tmpdir + "/node3"
+    def setup_nodes(self):
```
JackPiri: It seems to me there is no need to override `setup_nodes` with this specific function (having the exact same implementation).

qa/rpc-tests/blockdelay_2.py
```diff
@@ -23,40 +23,15 @@ def setup_chain(self, split=False):
...
-        self.is_network_split = True
         
-    def split_network_2(self):
-        # Split the network of three nodes into nodes 0, 1 and 2.
-        disconnect_nodes(self.nodes[1], 2)
-        disconnect_nodes(self.nodes[2], 1)
-
-
-    def join_network(self):
+    def join_network(self, id = 0):
```
JackPiri, line 29: I propose to remove both `join_network` and `join_network_2` from this test; the base class implementation can be used setting properly the `id` parameter (please note `sync_blocks` and `sync_mempools` are currently useless, being called on single-element-set, like `self.nodes[X:X+1]`).
Please replace accordingly calls to `join_network_2` with `join_network` with parameter `id=1` (lines 432, 572).

qa/rpc-tests/headers_01.py
```diff
@@ -23,39 +23,20 @@ def setup_chain(self, split=False):
...
 
-        if not split:
-            connect_nodes_bi(self.nodes, 0, 1)
-            sync_blocks(self.nodes[0:2])
-            sync_mempools(self.nodes[0:2])
-
-        self.is_network_split = split
-        self.sync_all()
-
     def split_network(self):
```
JackPiri: `split_network` can be removed (since it is not used).

qa/rpc-tests/sc_cert_nonceasing.py
```diff
@@ -43,13 +43,6 @@ def setup_network(self, split=False):
...
         self.sync_all()
 
-    def split_network(self):
-        # Split the network of three nodes into nodes 0 and 1
-        assert not self.is_network_split
-        disconnect_nodes(self.nodes[0], 1)
-        disconnect_nodes(self.nodes[1], 0)
-        self.is_network_split = True
-
     def join_network(self):
```
JackPiri: `join_network` can be removed and the call at line 316 modified inserting parameter `id=0` (if the `time.sleep` call is really mandatory, it can be inserted just after, at line 317).

qa/rpc-tests/sc_csw_actcertdata.py
```diff
@@ -53,14 +53,6 @@ def split_network(self):
         disconnect_nodes(self.nodes[2], 1)
```
JackPiri: `split_network` can be removed (since it is not used).

qa/rpc-tests/sc_csw_fundrawtransaction.py
```diff
@@ -52,14 +52,6 @@ def split_network(self):
         disconnect_nodes(self.nodes[2], 1)
```
JackPiri: `split_network` can be removed (since it is not used).

qa/rpc-tests/sc_ft_and_mbtr_fees_update.py
```diff
@@ -49,12 +49,6 @@ def setup_network(self, split=False):
         self.sync_all()
         self.firstRound = True
 
-    def split_network(self):
-        # Split the network of three nodes into nodes 0/1 and 2/3.
-        disconnect_nodes(self.nodes[1], 2)
-        disconnect_nodes(self.nodes[2], 1)
-        self.is_network_split = True
-
     def join_network(self):
```
JackPiri: `join_network` can be removed (base impl can be used, with parameter `id=1`).

qa/rpc-tests/sc_fwd_maturity.py
```diff
@@ -50,16 +50,6 @@ def split_network(self):
         disconnect_nodes(self.nodes[2], 1)
```
JackPiri: `split_network` can be removed (since it is not used).

JackPiri (changes requested):
Good PR, it is a big effort toward making the test suite more clean and reusable.
I left some comments, it seems to me a good chance for pushing further with the refactoring; I think the PR needs to address them before being merged.
Additionally to the PR-involved-files comments, please consider:

- `sc_csw_balance_exceeding.py`: remove `split_network`/`join_network `(not even used),
- possibly remove unused imports (due to reference removal) on top of Pyhton tests.

Thanks,

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -90,14 +83,18 @@ def sync_all(self):
...
-        self.setup_network(False)
+        connect_nodes_bi(self.nodes, id, id + 1)
+        self.is_network_split = False
+
+    def join_network_sync(self, id = 1):
+        # Join the (previously split) adjanced nodes in linear network together: 0-1 >-< 2-3 and syncs both parts
+        assert self.is_network_split
+        connect_nodes_bi(self.nodes, id, id + 1)
+        self.sync_all()
+        self.is_network_split = False
```
a-petrini: I believe that `is_network_split` has to be set to `False` before calling `sync_all()`, since the behaviour of the latter function depends on this flag. In particular, with the current version of the code, mempool and chain synchronization happens amongst each group. Instead, we would like that after the execution of this function all the nodes have consistent mempool and blockchain.

Fixing the order will break / freeze `sc_invalidate.py` and `sc_csw_nullifier.py`, while logic in `sc_split.py` is not preserved.

qa/rpc-tests/sc_split.py
```diff
@@ -179,7 +164,7 @@ def run_test(self):
         self.sync_all()
 
         mark_logs("\nJoining network", self.nodes, DEBUG_MODE)
-        self.join_network()
+        self.join_network_sync()
```
a-petrini: This is somehow linked to my other comment, but in this test the old `join_network()` was not enforcing a sync on the nodes. As a matter of fact, in following `Checking mempools...` loop we are differentiating between the mempool groups. Still, the old `time.sleep(2)` plays a major role in not crashing the test... (see comment on `test_framework.py`)

qa/rpc-tests/test_framework/test_framework.py
titusen: If any method is called with out of range index, the "Unexpected exception caught during testing: list index out of range" is thrown and handled in test_framework.py.

```Split nodes (1)----x   x---(2)
Unexpected exception caught during testing: list index out of range
  File "/home/titusen/workspace/zen/qa/rpc-tests/test_framework/test_framework.py", line 132, in main
    self.run_test()
  File "/home/titusen/workspace/zen/qa/rpc-tests/headers_06.py", line 130, in run_test
    self.split_network(10)
  File "/home/titusen/workspace/zen/qa/rpc-tests/test_framework/test_framework.py", line 71, in split_network
    disconnect_nodes(self.nodes[id], id + 1)
Stopping nodes
Cleaning up
Failed
!!! FAIL: headers_06.py !!!```

File: qa/rpc-tests/blockdelay.py (hunk 8 above)
Comment (titusen) [reply qa/rpc-tests/blockdelay.py]: The default is 4 nodes, Here are 3.

File: qa/rpc-tests/headers_01.py (hunk 10 above)
Comment (titusen) [reply qa/rpc-tests/headers_01.py]: Done

File: qa/rpc-tests/sc_csw_actcertdata.py (hunk 12 above)
Comment (titusen) [reply qa/rpc-tests/sc_csw_actcertdata.py]: Done

File: qa/rpc-tests/sc_csw_fundrawtransaction.py (hunk 13 above)
Comment (titusen) [reply qa/rpc-tests/sc_csw_fundrawtransaction.py]: Done

File: qa/rpc-tests/sc_ft_and_mbtr_fees_update.py (hunk 14 above)
Comment (titusen) [reply qa/rpc-tests/sc_ft_and_mbtr_fees_update.py]: Done

File: qa/rpc-tests/blockdelay.py (hunk 8 above)
Comment (JackPiri) [reply qa/rpc-tests/blockdelay.py]: Mhm, I see 4 nodes in the overriding:
```
def setup_nodes(self):
        self.nodes = start_nodes(4, self.options.tmpdir)
```
so it seems to me it's the same implementation.

**Review by JackPiri — APPROVED** [review]
The PR is fine to me.
Thanks

File: qa/rpc-tests/sc_cert_ceasing_split.py [hunk 18]
```diff
@@ -231,7 +205,8 @@ def run_test(self):
 
         #============================================================================================
         mark_logs("\nJoining network", self.nodes, DEBUG_MODE)
-        self.join_network()
+        self.join_network_sync(2)
+
```
a-petrini: Assert at line 217 sporadically fails after this sync (verified with `run_until_fails.py` script). Using the `join_network_sync` made the test fail, so I suggest to revert back to the old one, for the meanwhile

qa/rpc-tests/sc_cert_addressindex.py
a-petrini: I believe you removed the wrong one: now `join_network()` is the function that stops and restarts the node (as defined here), while `split_network()` is the one in the base class, which just reconnect the nodes. It would be nice to preserve the symmetry of the functions originally defined.

As a side note, in the future, **please let the reviewer mark the conversation as resolved**, as issues like these may be overlooked.

qa/rpc-tests/sc_cert_addressindex.py
```diff
@@ -390,7 +374,7 @@ def run_test_internal(self, sidechain_version, is_non_ceasing=False):
 
         # Split the network: (0) / (1)
         mark_logs("\nSplit network", self.nodes, DEBUG_MODE)
-        self.split_network()
+        self.split_network(0)
```
a-petrini, line 368: This will come handy after fixing the previous comment: teh subsequent call to `join_network()` at around line 440 has no parameter, so it will try to connect nodes 1 and 2, while the split created here are between nodes 0 and 1.

qa/rpc-tests/sbh_rpc_cmds.py
```diff
@@ -55,15 +55,6 @@ def split_network(self):
         disconnect_nodes(self.nodes[2], 1)
         self.is_network_split = True
```
a-petrini: `split_network()` is not called anywhere in the test, so it can be removed

qa/rpc-tests/sc_invalidate.py
```diff
@@ -214,7 +198,7 @@ def run_test(self):
         self.sync_all()
 
         mark_logs("\nJoining network", self.nodes, DEBUG_MODE)
-        self.join_network()
+        self.join_network_sync()
```
a-petrini: (see comment on `test_framework.py`)

qa/rpc-tests/sc_csw_nullifier.py
```diff
@@ -460,7 +445,7 @@ def run_test(self):
 
         #============================================================================================
         mark_logs("\nJoining network", self.nodes, DEBUG_MODE)
-        self.join_network()
+        self.join_network_sync()
```
a-petrini: (see comment on `test_framework.py`)

qa/rpc-tests/sc_quality_nodes.py
```diff
@@ -232,7 +210,7 @@ def run_test(self):
         cert2_mined = self.nodes[1].generate(1)[0]
         assert_true(cert_2_epoch_0 in self.nodes[1].getblock(cert2_mined, True)['cert'])
 
-        self.join_network()
+        self.join_network(0)
```
a-petrini, line 213: The new `join_network()` does not have the `time.sleep(2)`. Do you mind add it here, as done for `sc_cert_ceasing_split.py` and  `sc_cert_memcleanup_split.py`?

qa/rpc-tests/sc_quality_nodes.py
```diff
@@ -331,7 +309,7 @@ def run_test(self):
         cert2_mined = self.nodes[1].generate(1)[0]
         assert_true(cert_2_epoch_1 in self.nodes[1].getblock(cert2_mined, True)['cert'])
 
-        self.join_network()
+        self.join_network(0)
```
a-petrini, line 312: (see previous comment)

qa/rpc-tests/sc_csw_memcleanup_split.py
```diff
@@ -248,7 +233,7 @@ def run_test(self):
 
         #============================================================================================
         mark_logs("\nJoining network", self.nodes, DEBUG_MODE)
-        self.join_network()
+        self.join_network(2)
```
a-petrini, line 236: The new join_network() does not have the time.sleep(2). Do you mind add it here, as done for sc_cert_ceasing_split.py and sc_cert_memcleanup_split.py?

a-petrini (changes requested):
Sorry for the delay, I've found these issues that should at least be discussed.
There are a few things that I have left out from the `headers_xx.py` test, that we could address in a separate ticket, since they need a major refactoring.

qa/rpc-tests/sc_split.py
titusen: done

qa/rpc-tests/test_framework/test_framework.py
titusen: I thought that syncing entire network after join makes sense. I agree that it was this way.

qa/rpc-tests/sc_csw_memcleanup_split.py
titusen: Done

qa/rpc-tests/sc_cert_ceasing_split.py
titusen: done

qa/rpc-tests/sc_cert_addressindex.py
titusen: Done

qa/rpc-tests/sc_cert_addressindex.py
titusen: done

qa/rpc-tests/sbh_rpc_cmds.py
titusen: Done

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -90,14 +82,18 @@ def sync_all(self):
...
```
drgora: I agree with @a-petrini. This code, in its current form, is not actually synchronizing the entire network, but only nodes `0-1` and `2-3`. With the following modification, you should synchronize the entire network instead:
```suggestion
        self.is_network_split = False
        self.sync_all()
```

qa/rpc-tests/sc_invalidate.py
```diff
@@ -215,6 +199,7 @@ def run_test(self):
 
         mark_logs("\nJoining network", self.nodes, DEBUG_MODE)
         self.join_network()
+        time.sleep(2)
```
drgora, line 202: Would it be possible to use the new `join_network_sync()` instead of `time.sleep()` here?
```suggestion
        self.join_network_sync()
```

qa/rpc-tests/sc_cert_nonceasing.py
```diff
@@ -320,7 +305,8 @@ def run_test(self):
         #------------------------------------------------
         mark_logs("## Join network ##", self.nodes, DEBUG_MODE)
         #------------------------------------------------
-        self.join_network()
+        self.join_network(0)
+        time.sleep(2)
         self.sync_all()
```
drgora, line 310: Here we synchronize already, couldn't we drop the `sleep()`?

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -50,35 +51,26 @@ def setup_chain(self):
...
         assert not self.is_network_split
-        stop_nodes(self.nodes)
-        wait_bitcoinds()
-        self.setup_network(True)
+        disconnect_nodes(self.nodes[id], id + 1)
+        disconnect_nodes(self.nodes[id + 1], id)
+        self.is_network_split = True
 
     def sync_all(self):
         if self.is_network_split:
```
drgora, line 76: This branch behaves as expected only if the network is split in the expected pair of nodes. Now that `split_network()` is more flexible, this might break.
Ideally, we should assess that the expected condition is met, and looking at `disconnect_nodes` I believe we already have all the needed components to implement a `nodes_disconnected()` helper function.
I understand this requires more effort, so feel free to ignore this comment  and keep it as a reminder for a future PR.

qa/rpc-tests/getchaintips.py
```diff
@@ -21,8 +58,8 @@ def run_test (self):
         assert_equal (tips[0]['height'], 200)
         assert_equal (tips[0]['status'], 'active')
 
-        # Split the network and build two chains of different lengths.
-        self.split_network ()
+        # Split the network and build two chains of different lengths. ??????
```
drgora: I guess the `??????` is a leftover?

qa/rpc-tests/getchaintips.py
```diff
@@ -43,7 +80,8 @@ def run_test (self):
 
         # Join the network halves and check that we now have two tips
         # (at least at the nodes that previously had the short chain).
-        self.join_network ()
+        self.join_network()
+        time.sleep(2)
```
drgora, line 85: This `join_network()` already executes a `sync_all()`. Is it possible to drop this `time.sleep(2)`?

qa/rpc-tests/headers_04.py
```diff
@@ -131,7 +104,7 @@ def run_test(self):
         print("\n\nJoin network")
 #        raw_input("press enter to join the netorks..")
         self.mark_logs("Joining network")
-        self.join_network()
+        self.join_network_sync()
 
         time.sleep(2)
```
drgora, line 109: In this and all other cases where we have a `join_network_sync()` followed by `time.sleep(2)`, could we remove `time.sleep(2)`?

qa/rpc-tests/sc_quality_nodes.py
drgora: Is that actually needed in this case? I believe the `sync_all()` at the following line should be a more reliable way of achieving the same state?

drgora (changes requested):
Thank you for the PR! I really like how the new implementation avoids stopping and restarting nodes in `split_network` and `join_network`, this looks much more efficient.

I left few comments. In particular, I'd suggest to make sure if/why we need the many `time.sleep()` calls that we have. I see that most of the times we had them already, but this might be a great opportunity to get rid of them as much as possible. In general, and whenever possible, I would prefer having deterministic states, as example after a `sync_*` call, instead of relying on non-deterministic after sleep states.
This would make test more robust, and make the code simpler and more readable.

qa/rpc-tests/test_framework/test_framework.py
titusen: Will do

qa/rpc-tests/getchaintips.py
titusen: Moved to ZEND-616

qa/rpc-tests/headers_04.py
titusen: Moved to ZEND-616](https://horizenlabs.atlassian.net/browse/ZEND-616)

qa/rpc-tests/sc_invalidate.py
titusen: Moved to ZEND-616](https://horizenlabs.atlassian.net/browse/ZEND-616)

qa/rpc-tests/sc_quality_nodes.py
titusen: Moved to ZEND-616](https://horizenlabs.atlassian.net/browse/ZEND-616)

qa/rpc-tests/sc_quality_nodes.py
titusen: Moved to ZEND-616](https://horizenlabs.atlassian.net/browse/ZEND-616)

qa/rpc-tests/test_framework/test_framework.py
titusen: Done

qa/rpc-tests/test_framework/test_framework.py
titusen: Done

qa/rpc-tests/test_framework/test_framework.py
titusen: Moved to ZEND-616](https://horizenlabs.atlassian.net/browse/ZEND-616)

qa/rpc-tests/sc_cert_addressindex.py
a-petrini: Nice, thanks!

qa/rpc-tests/sc_cert_addressindex.py
a-petrini: thanks

qa/rpc-tests/test_framework/test_framework.py
a-petrini: Nice, thanks!

qa/rpc-tests/sc_cert_ceasing_split.py
a-petrini: ok

qa/rpc-tests/sc_quality_nodes.py
a-petrini: @drgora yes, I agree that relying on a timed even is not safe, but I like the idea to investigate in another ticket why some tests do not need an explicit sync after join; therefore, my suggestion is to momentarily keep things as they were before, an move on with that refactoring in the other ticket

qa/rpc-tests/sc_csw_memcleanup_split.py
a-petrini: ok

qa/rpc-tests/sc_quality_nodes.py
a-petrini: thanks

a-petrini (approved):

qa/rpc-tests/sbh_rpc_cmds.py
a-petrini: thanks

qa/rpc-tests/sc_cert_nonceasing.py
titusen: In this test the sleep was in join. Moved to https://horizenlabs.atlassian.net/browse/ZEND-616

qa/rpc-tests/test_framework/test_framework.py
drgora: Great, thanks!

qa/rpc-tests/sc_quality_nodes.py
drgora: Fine for me, thanks

drgora (approved):
Looks good to me, thanks

ptagl (approved):

PR #565, opened 2023-06-09
Remove libzendoo hardcoded dependency

la10736:
The legacy implementation inherit from the librustzcash needs to define a dependency for each crate used from libzendoo. This implementation create a local cargo registry and download all crates in this registry : this implementation force to update all these dependency definition every time libzendoo change its dependency... it's just a useless duplication of the Cargo.lock file posted on libzendoo repo.

Later, in another PR, we'll move the Cargo.lock from libzendoo to zen repo in order to create this file just when we freeze the branch to prepare the release, provide it to who would rebuild zen from scratch and don't use it at all in developing stage.


JackPiri (approved):
The PR is fine to me.
I saw there are some old unused `*.mk` files (not associated with libzendoo) in the `depends/packages` directory, we could get rid of them in a later cleaning PR:
- `crate_ff.mk`
- `crate_ff_derive.mk`
- `crate_fuchsia-cprng.mk`
- `crate_group.mk`
- `crate_itertools.mk`
- `crate_wasi.mk`

Thanks,

drgora (approved):
Looks very good to me, thanks

ptagl (approved):
Just to be sure, we performed some synchronization tests with `zend` on Linux, Windows, and MacOS, and as expected, we didn't find any issues.

Thanks!

PR #567, opened 2023-06-23
Codacy warning clearance

titusen:
Methods arguments change to reference or l-value.

### DIFF
--- a/src/wallet-utility.cpp
+++ b/src/wallet-utility.cpp
@@ -47,9 +47,9 @@ class WalletUtilityDB : public CDB
 
         std::string getAddress(CDataStream ssKey);
         std::string getKey(CDataStream ssKey, CDataStream ssValue);
-        std::string getCryptedKey(CDataStream ssKey, CDataStream ssValue, std::string masterPass);
+        std::string getCryptedKey(CDataStream ssKey, CDataStream ssValue, const std::string& masterPass);
         bool updateMasterKeys(CDataStream ssKey, CDataStream ssValue);
-        bool parseKeys(bool dumppriv, std::string masterPass);
+        bool parseKeys(bool dumppriv, const std::string& masterPass);
 
         bool DecryptSecret(const std::vector<unsigned char>& vchCiphertext, const uint256& nIV, CKeyingMaterial& vchPlaintext);
         bool Unlock();
@@ -141,7 +141,7 @@ bool WalletUtilityDB::DecryptKey(const std::vector<unsigned char>& vchCryptedSec
 /*
  * Encrypted private key in WIF format
  */
-std::string WalletUtilityDB::getCryptedKey(CDataStream ssKey, CDataStream ssValue, std::string masterPass)
+std::string WalletUtilityDB::getCryptedKey(CDataStream ssKey, CDataStream ssValue, const std::string& masterPass)
 {
     mPass = masterPass.c_str();
     CPubKey vchPubKey;
@@ -188,7 +188,7 @@ bool WalletUtilityDB::updateMasterKeys(CDataStream ssKey, CDataStream ssValue)
 /*
  * Look at all the records and parse keys for addresses and private keys
  */
-bool WalletUtilityDB::parseKeys(bool dumppriv, std::string masterPass)
+bool WalletUtilityDB::parseKeys(bool dumppriv, const std::string& masterPass)
 {
     DBErrors result = DB_LOAD_OK;
     std::string strType;
--- a/src/wallet/asyncrpcoperation_sendmany.cpp
+++ b/src/wallet/asyncrpcoperation_sendmany.cpp
@@ -50,14 +50,14 @@ int find_output(UniValue obj, int n) {
 
 AsyncRPCOperation_sendmany::AsyncRPCOperation_sendmany(
         CMutableTransaction contextualTx,
-        std::string fromAddress,
-        std::vector<SendManyRecipient> tOutputs,
-        std::vector<SendManyRecipient> zOutputs,
+        const std::string& fromAddress,
+        std::vector<SendManyRecipient>&& tOutputs,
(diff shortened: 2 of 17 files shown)


src/wallet-utility.cpp
```diff
@@ -141,7 +141,7 @@ bool WalletUtilityDB::DecryptKey(const std::vector<unsigned char>& vchCryptedSec
 /*
  * Encrypted private key in WIF format
  */
-std::string WalletUtilityDB::getCryptedKey(CDataStream ssKey, CDataStream ssValue, std::string masterPass)
+std::string WalletUtilityDB::getCryptedKey(CDataStream ssKey, CDataStream ssValue, const std::string& masterPass)
 {
     mPass = masterPass.c_str();
```
drgora, line 146: Not actually introduced by your PR, but the previous version made me worry for the validity of the `char*` for a second. I believe this looks better:
```suggestion
    mPass = masterPass;
```

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -50,9 +50,9 @@ int find_output(UniValue obj, int n) {
 
 AsyncRPCOperation_sendmany::AsyncRPCOperation_sendmany(
         CMutableTransaction contextualTx,
-        std::string fromAddress,
-        std::vector<SendManyRecipient> tOutputs,
-        std::vector<SendManyRecipient> zOutputs,
+        const std::string& fromAddress,
+        std::vector<SendManyRecipient>&& tOutputs,
+        std::vector<SendManyRecipient>&& zOutputs,
```
drgora, line 55: If I understand correctly your PR, I believe you want to `std::move(tOutputs)` and `std::move(zOutputs)` below, at line 60.
That would actually take advantage of the move semantics, otherwise you would still end up copying the content of the vectors (also notice line 72).

src/zcash/IncrementalMerkleTree.hpp
```diff
@@ -51,7 +51,7 @@ class MerklePath {
 
     MerklePath() { }
 
-    MerklePath(std::vector<std::vector<bool>> authentication_path, std::vector<bool> index)
+    MerklePath(std::vector<std::vector<bool>>&& authentication_path, std::vector<bool>&& index)
     : authentication_path(authentication_path), index(index) { }
```
drgora: Similarly to above, I believe here you might want to `std::move` this vectors?

drgora (approved):
Thank you for the PR, looks good to me!
I left only a couple of non critical comments.

src/wallet/asyncrpcoperation_sendmany.cpp
titusen: DOne

src/wallet-utility.cpp
titusen: Removed cause variable is not used.

src/zcash/IncrementalMerkleTree.hpp
titusen: done

src/wallet/asyncrpcoperation_sendmany.h
```diff
@@ -51,7 +51,7 @@ struct WitnessAnchorData {
 
 class AsyncRPCOperation_sendmany : public AsyncRPCOperation {
 public:
-    AsyncRPCOperation_sendmany(CMutableTransaction contextualTx, std::string fromAddress, std::vector<SendManyRecipient> tOutputs, std::vector<SendManyRecipient> zOutputs, int minDepth, CAmount fee = ASYNC_RPC_OPERATION_DEFAULT_MINERS_FEE, UniValue contextInfo = NullUniValue, bool sendChangeToSource = false);
+    AsyncRPCOperation_sendmany(CMutableTransaction contextualTx, const std::string& fromAddress, std::vector<SendManyRecipient>&& tOutputs, std::vector<SendManyRecipient>&& zOutputs, int minDepth, CAmount fee = ASYNC_RPC_OPERATION_DEFAULT_MINERS_FEE, UniValue contextInfo = NullUniValue, bool sendChangeToSource = false);
     virtual ~AsyncRPCOperation_sendmany();
     
     // We don't want to be copied or moved around
```
JackPiri, line 57: Checking the list by Codacy, it seems to me also the method:
`std::array<unsigned char, ZC_MEMO_SIZE> get_memo_from_hex_string(const std::string& s);`
at line 103 should be modified (together with its implementation in `asyncrpcoperation_sendmany.cpp`).

JackPiri (changes requested):
The PR is fine to me, apart from `src/wallet/asyncrpcoperation_sendmany.h/.cpp` which require additional refactoring on `get_memo_from_hex_string`.
Please address this comment and I'll mark my review as approved.
Thanks

src/wallet/asyncrpcoperation_sendmany.h
titusen: Done

src/wallet/asyncrpcoperation_sendmany.cpp
drgora: Thank you, that looks good. But now you need to adjust line 72 accordingly, where you should use the members of the class instead of the parameters that you moved from.

src/wallet-utility.cpp
drgora: Mh, from a quick looks it seems to me that it is passed as a parameter `SetKeyFromPassphrase()`. Am I missing something?

JackPiri (approved):

src/wallet/asyncrpcoperation_sendmany.cpp
titusen: Done

ptagl (approved):

a-petrini (approved):
Nice PR, thanks!

PR #569, opened 2023-06-30
Use rust 1.70.0 to compile rust dependencies

la10736:
Updated the rust version to 1.70.0

Also updated zendoo-mc-cryptolib to 0.4.0. The pk/vk generate with this version are different to the ones generate by the previous one but

- the old saved pk/vk will continue to work also with this version
- If a use would to che the old vk/pk should use the old version library

The `librustzcash` compilation emit 2 (trivial) warning that I shouted down with `RUSTFLAGS` env variable

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -304,6 +304,7 @@ case $host in
      fi
 
      CPPFLAGS="$CPPFLAGS -D_MT -DWIN32 -D_WINDOWS -DBOOST_THREAD_USE_LIB"
+     LIBS="$LIBS -lntdll"
      LEVELDB_TARGET_FLAGS="TARGET_OS=OS_WINDOWS_CROSSCOMPILE"
      if test "x$CXXFLAGS_overridden" = "xno"; then
        CXXFLAGS="$CXXFLAGS -w"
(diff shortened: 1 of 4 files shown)


titusen (approved):
LGTM, thanks.

JackPiri (approved):
The PR is fine to me.
Thanks

configure.ac
```diff
@@ -304,6 +304,7 @@ case $host in
      fi
 
      CPPFLAGS="$CPPFLAGS -D_MT -DWIN32 -D_WINDOWS -DBOOST_THREAD_USE_LIB"
+     LIBS="$LIBS -lntdll"
```
drgora: I believe it would be cleaner to be consistent with lines 270-290. Could we use `AC_CHECK_LIB` here too, for `ntdll`?

drgora (changes requested):
Thank you for the PR. I just left one small comment, not really a blocker, but I think it's worth implementing.
Everything else looks good to me.

configure.ac
JackPiri: Done, please double check if we are aligned on this.

drgora: Yes, that is perfect, thank you!

drgora (approved):

a-petrini (approved):
Looks good, thanks for this PR!

ptagl (approved):

PR #573, opened 2023-07-07
Networking subsystem refactoring: CConnman class - ZEND-577

a-petrini:
This PR adds a new class named CConnman which encapsulates some functions and variables that manage several aspect of the node network subsystem.
This follows the modularization approach currently present in Bitcoin, which has been taken as a reference while developing this refactoring. Also, this is meant to be only the first step for the modularization of this subsystem, and code has been prepared to accommodate changes for a better management of threads, or the creation of a class for managing node ban


src/zen/tlsmanager.cpp
```diff
@@ -12,6 +12,8 @@
 #include "tlsmanager.h"
 #include "utiltls.h"
 
+extern std::unique_ptr<CConnman> connman;
```
titusen, line 15: Generally, having a global variables is a bad practice. Is this like a standard inherited from bitcoin?

src/init.cpp
```diff
@@ -1353,6 +1353,9 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
 
     RegisterNodeSignals(GetNodeSignals());
 
+    connman.reset();
```
titusen: if conman is reseted anyway why not directly with the new value.

src/zen/tlsmanager.cpp
a-petrini: Yes, I totally agree with you!
Bitcoin groups all the pointers to modules in the global structure `NodeContext` (see src/node/context.h in Bitcoin repo). The long term plan for us is to replicate this model and go on with the modularization of most of zend subsystems, but at the moment only the connection manager and the mempool are ready for this.
A possible plan for the short term is to create the NodeContext and add these two components in a future PR, as this could be done in a separate and independent task; at the moment, however, we should endure with globals...

src/init.cpp
a-petrini: Yes, and this is actually useless, as at this point `connman` should contain a `nullptr`, and the following instruction calls the CConnman constructor, effectively creating a new object. I would replace this one with an assertion, checking that connman is `nullptr`.

src/net.cpp
```diff
@@ -615,8 +613,7 @@ void CNode::GetBanned(std::map<CSubNet, int64_t> &banMap)
     banMap = setBanned; //create a thread safe copy
 }
 
-/// To be moved to CConnman
-bool CNode::IsWhitelistedRange(const CNetAddr &addr) {
+bool CConnman::IsWhitelistedRange(const CNetAddr &addr) {
```
titusen, line 613: I guess for the follow-up to review data structures used here. Linear lookup?!

src/zen/tlsmanager.h
```diff
@@ -20,16 +20,16 @@ using namespace std;
...
+//     std::string ipAddr;
+//     int64_t time; // time in msec, of an attempt to connect via TLS
 
-    _NODE_ADDR(std::string _ipAddr, int64_t _time = 0) : ipAddr(_ipAddr), time(_time) {}
-bool operator==(const _NODE_ADDR b) const
-{
-    return (ipAddr == b.ipAddr);
-}
-} NODE_ADDR, *PNODE_ADDR;
+//     _NODE_ADDR(std::string _ipAddr, int64_t _time = 0) : ipAddr(_ipAddr), time(_time) {}
```
titusen: Unnecessary comments. Please, remove.

src/net.h
```diff
@@ -724,4 +682,158 @@ class CAddrDB
...
+
+    void SocketSendData(CNode *pnode);
+
+    CConnman();
+    ~CConnman();
+
+    bool fAddressesInitialized {false};
+    std::unique_ptr<CSemaphore> semOutbound = nullptr;
+    std::unique_ptr<CNode> pnodeLocalHost = nullptr;
+
```
titusen: I don't believe all these fields have to be public. Please consider encapsulation. Possible follow-up.

src/gtest/test_mempool.cpp
```diff
@@ -443,8 +443,12 @@ class CNodeExt : public CNode
     }
 };
 
+extern std::unique_ptr<CConnman> connman;
+
 TEST(ProcessMempoolMsgTest, TxesInMempoolAreRelayed)
 {
+    connman.reset();
```
JackPiri: Maybe this can be made a one-liner:
```suggestion
    connman.reset(new CConnman());
```
if you agree, please consider also three occurrences in `DoS_tests.cpp`.

src/net.h
```diff
@@ -687,16 +655,6 @@ class CNode
...
-
-    // Network stats
-    static void RecordBytesRecv(uint64_t bytes);
-    static void RecordBytesSent(uint64_t bytes);
-
-    static uint64_t GetTotalBytesRecv();
-    static uint64_t GetTotalBytesSent();
-
     // resource deallocation on cleanup, called at node shutdown
     static void NetCleanup();
```
JackPiri: It seems to me that `static void NetCleanup();` can be removed, being only used the `CConnman` implementation.

src/net.h
```diff
@@ -724,4 +682,158 @@ class CAddrDB
...
+    std::vector<ListenSocket> vhListenSocket;
+
+    std::atomic<NodeId> nLastNodeId{0};
+    NodeId GetNewNodeId();
+
+    std::vector<zen::NODE_ADDR> vNonTLSNodesInbound;
+    CCriticalSection cs_vNonTLSNodesInbound;
+    std::vector<zen::NODE_ADDR> vNonTLSNodesOutbound;
+    CCriticalSection cs_vNonTLSNodesOutbound;
+    std::set<CNetAddr> setservAddNodeAddresses;     // This could be removed!
```
JackPiri: I agree with the comment, let's remove `setservAddNodeAddresses` and `cs_setservAddNodeAddresses` and the associated code; it seems to me it's write-only and never used.

src/zen/tlsmanager.h
```diff
@@ -20,16 +20,16 @@ using namespace std;
 
 namespace zen
 {
-typedef struct _NODE_ADDR {
-    std::string ipAddr;
-    int64_t time; // time in msec, of an attempt to connect via TLS
+// typedef struct _NODE_ADDR {
```
JackPiri: The `struct _NODE_ADDR` seems to me has been completely moved to `net.h`, let's remove the commented code definitely.

src/net.h
```diff
@@ -724,4 +682,158 @@ class CAddrDB
...
+
+    bool fAddressesInitialized {false};
+    std::unique_ptr<CSemaphore> semOutbound = nullptr;
+    std::unique_ptr<CNode> pnodeLocalHost = nullptr;
+
+    std::vector<CNode*> vNodes;
+    CCriticalSection cs_vNodes;
+    std::vector<std::string> vAddedNodes;
+    CCriticalSection cs_vAddedNodes;
+    std::list<CNode*> vNodesDisconnected;
```
JackPiri, line 764: I see `vNodesDisconnected` is used in just two places, `CConnman::NetCleanup` and `ThreadSocketHandler`; actually it seems to me there is no real synchronization of those two, since `ThreadSocketHandler` runs indefinitely and `CConnman::NetCleanup` is called in `Shutdown` and in `CConnman` destructor. Is it worth to add a locking mechanism through `CCriticalSection`?

src/init.cpp
```diff
@@ -1353,6 +1342,10 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
 
     RegisterNodeSignals(GetNodeSignals());
 
+    assert(!connman);
+    connman = std::make_unique<CConnman>();
+    CConnman::Options connOptions;
```
JackPiri, line 1355: Maybe it's just a feeling, but it seems a little confusing to me here:

1. we instantiate `connman` with void `connOptions`
2. we prepare `connOptions` (using `connman`)
3. later (at `StartNode`) we reinject `connOptions` in `connman`

I guess it's working fine the current way, nonetheless I'd prefer to have this flow a little bit more fluent and simple (if possible).

JackPiri (approved):
This PR is a good way to refactor our code, really appreciated.
I left some minor comments, please feel free to address or skip.
Thanks

src/net.cpp
a-petrini: Yes, agree. For this PR I did not change the underlying logic, and there is space for improvements

src/zen/tlsmanager.h
a-petrini: yes, I forgot to remove that, thanks.

src/zen/tlsmanager.h
a-petrini: I forgot to remove that, thanks!

src/net.h
a-petrini: Done

src/gtest/test_mempool.cpp
a-petrini: Yes, I also like this more compact notation. Changed as requested.

src/net.h
a-petrini: Nicely spotted! Totally forgot to remove that from `CNode`...

src/net.h
a-petrini: It is possible that `ThreadSocketHandler` may access `vNodesDisconnected` when `NetCleanup()` requests its destruction so, although rare, this could happen.
However, with the follow up boost::thread refactoring, we are going to stop all the network threads before calling `NetCleanup()`, so this will be fixed without introducing critical sections (the destructor will call `CConnman::StopNode()`, then `CConnman::StopThreads()` and finally `CConnman::NetCleanup()`)

JackPiri: Ok makes sens, thanks for the explanation.

src/gtest/test_mempool.cpp
JackPiri: If there is no rethinking let's remove the old commented code for those four occurrences, thanks.

src/net.h
```diff
@@ -724,4 +679,156 @@ class CAddrDB
...
+
+    _NODE_ADDR(std::string _ipAddr, int64_t _time = 0) : ipAddr(_ipAddr), time(_time) {}
+    bool operator==(const _NODE_ADDR b) const
+    {
+        return (ipAddr == b.ipAddr);
+    }
+} NODE_ADDR, *PNODE_ADDR;
+}
+
+class CConnman {
```
titusen, line 709: As a start of dissection of this file please move to separate one.

src/net.h
a-petrini: Yes, planned for the follow-up PR of boost::thread removal, as a few member can be made private (`vhListenSocket` and `vNodesDisconnected`, for instance) without writing any get or set function.
As now, I moved `fAddressesInitialized` and `pnodeLocalHost` to the list of private members.

src/gtest/test_mempool.cpp
a-petrini: You are right, forgot to remove the commented lines. Fixed.

src/init.cpp
a-petrini: The description of the sequence of tasks follows the current implementation, and it is derived from `CConnman`  initialization in Bitcoin. I suppose it has been written in this way because AppInit() relies on a few structures and functions of `CConnman` that should be called before starting the actual networking subsystem (with `StartNode()`). We could think of something simpler, but it would diverge a bit from what Bitcoin does. Maybe we can think of a follow-up discussion / PR for this purpose.

src/net.h
a-petrini: I like this suggestion, but at the moment I prefer to keep all the networking related classes into `net.h/.cpp`

src/init.cpp
JackPiri: Ok makes sense, thanks.

src/init.cpp
titusen: ok

ptagl (approved):
Very nice, thanks!

PR #574, opened 2023-07-07
Limit mempool size

drgora:
This PR adds a new CLI parameter, `-maxmempool`, that allows setting an upper limit on the size in bytes of the content of the mempool.
In order to respect such limit, transactions or certificates can be rejected/evicted according to context specific rules.

### DIFF
--- a/qa/rpc-tests/mempool_size_limit.py
+++ b/qa/rpc-tests/mempool_size_limit.py
@@ -1,0 +1,592 @@
+#!/usr/bin/env python3
+# Copyright (c) 2014 The Bitcoin Core developers
+# Copyright (c) 2018 The Zencash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.test_framework import ForkHeights
+from test_framework.authproxy import JSONRPCException
+from test_framework.util import assert_equal, initialize_chain_clean, \
+    start_nodes, sync_blocks, sync_mempools, connect_nodes_bi, mark_logs,\
+    get_epoch_data, swap_bytes
+from test_framework.mc_test.mc_test import *
+import os
+import zipfile
+import time
+from decimal import Decimal, ROUND_DOWN
+
+USE_SNAPSHOT = True # set to False to regenerate test data.
+
+DEBUG_MODE = 1
+NUMB_OF_NODES = 2
+EPOCH_LENGTH = 10
+
+FT_SC_FEE      = Decimal('0')
+MBTR_SC_FEE    = Decimal('0')
+CERT_FEE       = Decimal('0.00015')
+PARAMS_NAME = "sc"
+
+NODE0_LIMIT_B = 5000000
+NODE1_LIMIT_B = 4000000
+
+NODE0_CERT_LIMIT_B = NODE0_LIMIT_B / 2
+NODE1_CERT_LIMIT_B = NODE1_LIMIT_B / 2
+
+EPSILON = 100000
+MAX_FEE = Decimal("999999")
+
+NUM_CEASING_SIDECHAINS    = 2
+NUM_NONCEASING_SIDECHAINS = 2
+
+class mempool_size_limit(BitcoinTestFramework):
+    def import_data_to_data_dir(self):
+        # importing datadir resource
+        # Tests checkpoint creation (during startup rescan) and usage, checks everything ok with old wallet
+        resource_file = os.sep.join([os.path.dirname(__file__), 'resources', 'mempool_size_limit', 'test_setup_.zip'])
+        with zipfile.ZipFile(resource_file, 'r') as zip_ref:
+            zip_ref.extractall(self.options.tmpdir)
+
+    def setup_chain(self, split=False):
+        if (USE_SNAPSHOT):
+            self.import_data_to_data_dir()
+            os.remove(self.options.tmpdir+'/node0/regtest/debug.log') # make sure that we have only logs from this test
+            os.remove(self.options.tmpdir+'/node1/regtest/debug.log')
+            os.remove(self.options.tmpdir+'/node0/regtest/wallet.dat')
+            os.remove(self.options.tmpdir+'/node1/regtest/wallet.dat')
+
+        print("Initializing test directory " + self.options.tmpdir)
+        initialize_chain_clean(self.options.tmpdir, NUMB_OF_NODES)
+
+    def setup_network(self, split=False):
+        self.nodes = []
+
+        self.nodes = start_nodes(NUMB_OF_NODES, self.options.tmpdir, extra_args =
+            [
+                ['-debug=cert', '-debug=sc', '-debug=mempool', '-maxorphantx=10000', f'-maxmempool={NODE0_LIMIT_B / 1000000}', '-minconf=0', '-allownonstandardtx'],
+                ['-debug=cert', '-debug=sc', '-debug=mempool', '-maxorphantx=10000', f'-maxmempool={NODE1_LIMIT_B / 1000000}', '-minconf=0', '-allownonstandardtx']
+            ])
+
+        connect_nodes_bi(self.nodes, 0, 1)
+        sync_blocks(self.nodes[1:NUMB_OF_NODES])
+        sync_mempools(self.nodes[1:NUMB_OF_NODES])
+        self.sync_all()
+
+    def satoshi_round(self, amount):
+        return Decimal(amount).quantize(Decimal('0.00000001'), rounding=ROUND_DOWN)
+
+    def create_sc_certificates(self, scidx, epoch_number, quality, fee, size):
+        certs = []
+        tot_size = 0
+        raw_bwt_outs = []
+        os.makedirs(self.options.tmpdir + "/certs", exist_ok=True)
+        os.makedirs(self.options.tmpdir + f"/certs/{scidx}", exist_ok=True)
+        for i in range(128):
+            raw_bwt_outs.append({"address":self.nodes[1].getnewaddress(), "amount":Decimal("0.0001")})
+        i = 0
+        while (tot_size < size and len(self.utxos) > 0):
+            t = self.utxos.pop()
+            raw_inputs = [{'txid' : t['txid'], 'vout' : t['vout']}]
+            amt = t["amount"] - fee
+            raw_outs = {self.nodes[0].getnewaddress(): amt}
+
+            sc = self.sc[scidx]
+            scid = sc["id"]
+            epoch_len = sc["epoch_len"]
+            par_name = sc["params"]
+            constant = sc["constant"]
+            ref_height = self.nodes[0].getblockcount()
+            ceasing = True if epoch_len > 0 else False
+
+            scid_swapped = str(swap_bytes(scid))
+            _, epoch_cum_tree_hash, prev_cert_hash = get_epoch_data(scid, self.nodes[0], epoch_len, is_non_ceasing = not ceasing, reference_height = ref_height)
+
+            proof = self.mcTest.create_test_proof(par_name,
+                                             scid_swapped,
+                                             epoch_number,
+                                             quality,
+                                             MBTR_SC_FEE,
+                                             FT_SC_FEE,
+                                             epoch_cum_tree_hash,
+                                             prev_cert_hash,
+                                             constant = constant,
+                                             pks      = [raw_bwt_outs[i]["address"] for i in range(128)],
+                                             amounts  = [raw_bwt_outs[i]["amount"] for i in range(128)])
+
+            raw_params = {
+                "scid": scid,
+                "quality": quality,
+                "endEpochCumScTxCommTreeRoot": epoch_cum_tree_hash,
+                "scProof": proof,
+                "withdrawalEpochNumber": epoch_number
+            }
+
+            raw_cert    = self.nodes[0].createrawcertificate(raw_inputs, raw_outs, raw_bwt_outs, raw_params)
+            signed_cert = self.nodes[0].signrawtransaction(raw_cert)
+            certFile = open(self.options.tmpdir + f"/certs/{scidx}/c_{quality}", "wb")
+            certFile.write(bytes(str(fee) + '\n', encoding='utf8'))
+            certFile.write(bytes(signed_cert["hex"], encoding='utf8'))
+            certFile.close()
+            certs.append({'quality': i, 'fee': fee, 'hex': signed_cert["hex"]})
+            tot_size += len(signed_cert['hex'])//2
+            quality += 1
+            print("Tot cert: " + str(tot_size))
+            i += 1
+            if (epoch_len == 0):
+                return certs
+
+        return certs
+
+    def create_tx_script(self):
+        # Some pre-processing to create a bunch of OP_RETURN txouts to insert into transactions we create
+        # So we have big transactions (and therefore can't fit very many into each block)
+        # create one script_pubkey
+        self.script_pubkey = "6a4d0200" #OP_RETURN OP_PUSH2 512 bytes
+        for i in range (512):
+            self.script_pubkey = self.script_pubkey + "01"
+
+        #script_pubkey = script_pubkey + signature_imposter
+
+        ## concatenate 128 txouts of above script_pubkey which we'll insert before the txout for change
+        self.txouts = "80"
+        for k in range(128):
+            # add txout value
+            self.txouts = self.txouts + "b816000000000000"
+            # add length of script_pubkey
+            self.txouts = self.txouts + "fd0402"
+            # add script_pubkey
+            self.txouts = self.txouts + self.script_pubkey
+
+    def load_from_file(self, rFile):
+        fee = Decimal(rFile.readline().decode("utf-8"))
+        hex = rFile.readline().decode("utf-8")
+        return {"fee": fee, "hex": hex}
+
+    def load_from_storage(self, path):
+        res = []
+        certs = "certs" in path
+        for rf in os.listdir(path):
+            rFile = open(path + rf, "rb")
+            entry = self.load_from_file(rFile)
+            if (certs):
+                assert(rf.find("c_") == 0)
+                entry["quality"] = int(rf[2:])
+            res.append(entry)
+            rFile.close()
+        return res
+
+
+    def create_chained_transactions(self):
+        os.makedirs(self.options.tmpdir + "/ctxs")
+        addr = self.nodes[0].getnewaddress()
+        txs = []
+
+        low_fee = Decimal("0.0001")
+        high_fee = Decimal("10")
+        amt = Decimal("0")
+        while (amt < high_fee + low_fee):
+            t = self.utxos.pop()
+            amt = t['amount']
+
+        inputs = []
+        inputs.append({ "txid" : t["txid"], "vout" : t["vout"]})
+        outputs = {}
+        send_value = t['amount'] - low_fee
+        outputs[addr] = self.satoshi_round(send_value)
+        rawtx = self.nodes[0].createrawtransaction(inputs, outputs)
+        signresult = self.nodes[0].signrawtransaction(rawtx, None, None, "NONE")
+        csize = len(signresult['hex']) // 2
+
+        txFile = open(self.options.tmpdir + "/ctxs/ctx0", "wb")
+        txFile.write(bytes(str((low_fee) / csize) + '\n', encoding='utf8'))
+        txFile.write(bytes(signresult["hex"], encoding='utf8'))
+        txFile.close()
+        txs.append({'fee': high_fee / csize, 'hex': signresult["hex"]})
+
+        ptx = self.nodes[0].decoderawtransaction(signresult["hex"])
+
+        assert(high_fee < t['amount'])
+        inputs = []
+        inputs.append({ "txid" : ptx['txid'], "vout" : 0})
+        outputs = {}
+        send_value = t['amount'] - low_fee - high_fee
+        outputs[self.nodes[0].getnewaddress()] = self.satoshi_round(send_value)
+        rawtx = self.nodes[0].createrawtransaction(inputs, outputs)
+        signresult = self.nodes[0].signrawtransaction(rawtx, [{"txid": ptx['txid'], "vout": 0, "scriptPubKey": ptx['vout'][0]['scriptPubKey']['hex']}], None, "NONE")
+        csize = len(signresult['hex']) // 2
+
+        txFile = open(self.options.tmpdir + "/ctxs/ctx1", "wb")
+        txFile.write(bytes(str(high_fee / csize) + '\n', encoding='utf8'))
+        txFile.write(bytes(signresult["hex"], encoding='utf8'))
+        txFile.close()
+        txs.append({'fee': high_fee / csize, 'hex': signresult["hex"]})
+
+        return txs
+
+    def create_big_transactions(self, size):
+        self.create_tx_script()
+        addr = self.nodes[0].getnewaddress()
+        txs = []
+        tot_size = 0
+        i = 0
+        os.makedirs(self.options.tmpdir + "/txs")
+        while (tot_size < size):
+            t = self.utxos.pop()
+            inputs = []
+            inputs.append({ "txid" : t["txid"], "vout" : t["vout"]})
+            outputs = {}
+            send_value = t['amount']
+            outputs[addr] = self.satoshi_round(send_value)
+            rawtx = self.nodes[0].createrawtransaction(inputs, outputs)
+            out_num_pos = rawtx.find('ffffffff')
+            script_position = rawtx.find('76a914')
+            script_length = int(rawtx[script_position - 2 : script_position], 16) * 2
+
+            newtx = rawtx[0:out_num_pos + 8]
+            newtx = newtx + self.txouts
+            newtx = newtx + rawtx[script_position + script_length:]
+
+            signresult = self.nodes[0].signrawtransaction(newtx, None, None, "NONE")
+            tsize = len(signresult['hex']) // 2
+            tot_size += tsize
+            txFile = open(self.options.tmpdir + "/txs/tx" + str(i), "wb")
+            txFile.write(bytes(str(t['amount'] / tsize) + '\n', encoding='utf8'))
+            txFile.write(bytes(signresult["hex"], encoding='utf8'))
+            txFile.close()
+            txs.append({'fee': t['amount'] / tsize, 'hex': signresult["hex"]})
+            i += 1
+        print("Created big transaction for a total size: " + str(tot_size))
+        return txs
+
+    def create_sidechains(self, num, ceasing):
+        # generate wCertVk and constant
+        for i in range(num):
+            newsc = {}
+            constant = generate_random_field_element_hex()
+            ep_len = EPOCH_LENGTH if ceasing else 0
+            par_name = PARAMS_NAME + ("c" if ceasing else "nc") + str(i)
+
+            vk = self.mcTest.generate_params(par_name, keyrot=True)
+
+            # generate sidechain
+            cmdInput = {
+                "version": 2,
+                "withdrawalEpochLength": ep_len,
+                "toaddress": "dada",
+                "amount": Decimal("15"),
+                "wCertVk": vk,
+                "constant": constant,
+            }
+
+            ret = self.nodes[0].sc_create(cmdInput)
+            creating_tx = ret['txid']
+            scid = ret['scid']
+            mark_logs("Node 0 created the SC {} via tx {}.".format(scid, creating_tx), self.nodes, DEBUG_MODE)
+            newsc["id"] = scid
+            newsc["params"] = par_name
+            newsc["constant"] = constant
+            newsc["epoch_len"] = ep_len
+            self.sc.append(newsc)
+
+    def setup_test(self):
+        if USE_SNAPSHOT:
+            txs   = self.load_from_storage(self.options.tmpdir + "/txs/")
+            ctxs  = self.load_from_storage(self.options.tmpdir + "/ctxs/")
+            certs = []
+            for s in range(NUM_CEASING_SIDECHAINS):
+                certs.append(self.load_from_storage(self.options.tmpdir + f"/certs/{s}/"))
+            for s in range(NUM_NONCEASING_SIDECHAINS):
+                certs.append(self.load_from_storage(self.options.tmpdir + f"/certs/{s+NUM_CEASING_SIDECHAINS}/"))
+        else:
+            mark_logs("Node 0 generates {} block".format(ForkHeights['NON_CEASING_SC']), self.nodes, DEBUG_MODE)
+            self.nodes[0].generate(ForkHeights['NON_CEASING_SC'] + 800) # TODO: reduce to match the actual number of utxos needed
+            self.sync_all()
+
+            # forward transfer amounts
+            creation_amount = Decimal("50")
+            small_amount    = Decimal("0.000001")
+
+            ## Create sidechains
+            self.mcTest = CertTestUtils(self.options.tmpdir, self.options.srcdir)
+            self.sc = []
+            self.create_sidechains(NUM_CEASING_SIDECHAINS, True)
+            self.create_sidechains(NUM_NONCEASING_SIDECHAINS, False)
+            self.nodes[0].generate(EPOCH_LENGTH) # goto end epoch
+            self.sync_all()
+
+            self.utxos = self.nodes[0].listunspent(0)
+            txs = self.create_big_transactions(NODE0_LIMIT_B + EPSILON)
+            ctxs = self.create_chained_transactions()
+
+            certs = []
+            for i in range(NUM_CEASING_SIDECHAINS):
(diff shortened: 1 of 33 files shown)


src/gtest/test_sidechain_to_mempool.cpp
```diff
@@ -241,58 +242,58 @@ TEST_F(SidechainsInMempoolTestSuite, hasSidechainCreationTxTest) {
...
     CTxMemPoolEntry scPoolEntry(scTx, /*fee*/CAmount(1), /*time*/ 1000, /*priority*/1.0, /*height*/1987);
-    aMempool.addUnchecked(scPoolEntry.GetTx().GetHash(), scPoolEntry);
-    res = aMempool.hasSidechainCreationTx(scIdOk);
+    aMempool->addUnchecked(scPoolEntry.GetTx().GetHash(), scPoolEntry);
+    res = aMempool->hasSidechainCreationTx(scIdOk);
     EXPECT_TRUE(res);
 }
 
 TEST_F(SidechainsInMempoolTestSuite, ScAndFwdsAndBtrInMempool_ScNonRecursiveRemoval) {
-    // Associated scenario: Sidechain creation and some fwds and btr are in mempool.
```
ptagl: I guess this is one too many replacements here! 😄

src/gtest/test_sidechain_to_mempool.cpp
```diff
@@ -387,80 +388,80 @@ TEST_F(SidechainsInMempoolTestSuite, ScAndFwdsAndBtrInMempool_ScRecursiveRemoval
 }
 
 TEST_F(SidechainsInMempoolTestSuite, FwdsAndBtrOnlyInMempool_ScRecursiveRemoval) {
-    // Associated scenario: upon block disconnections fwds and btr have entered into mempool.
-    // While unmounting block containing scCreation, scCreation cannot make to mempool. fwds and btr must me purged
+    // Associated scenario: upon block disconnections fwds and btr have entered into mempool->
+    // While unmounting block containing scCreation, scCreation cannot make to mempool-> fwds and btr must me purged
```
ptagl: Also here we should probably keep "."!

src/gtest/test_sidechain_to_mempool.cpp
```diff
@@ -387,80 +388,80 @@ TEST_F(SidechainsInMempoolTestSuite, ScAndFwdsAndBtrInMempool_ScRecursiveRemoval
...
+    aMempool->remove(scTx, removedTxs, removedCerts, /*fRecursive*/true);
 
     EXPECT_TRUE(std::count(removedTxs.begin(), removedTxs.end(), fwdTx1));
     EXPECT_TRUE(std::count(removedTxs.begin(), removedTxs.end(), fwdTx2));
     EXPECT_TRUE(std::count(removedTxs.begin(), removedTxs.end(), btrTx));
 }
 
 TEST_F(SidechainsInMempoolTestSuite, ScAndFwdsAndBtrInMempool_FwdRecursiveRemoval) {
-    // Associated scenario: upon block disconnections a fwd cannot make to mempool.
+    // Associated scenario: upon block disconnections a fwd cannot make to mempool->
```
ptagl:
```suggestion
    // Associated scenario: upon block disconnections a fwd cannot make to mempool.
```

src/gtest/test_sidechain_to_mempool.cpp
```diff
@@ -387,80 +388,80 @@ TEST_F(SidechainsInMempoolTestSuite, ScAndFwdsAndBtrInMempool_ScRecursiveRemoval
...
+    aMempool->remove(fwdTx2, removedTxs, removedCerts, /*fRecursive*/true);
 
     EXPECT_FALSE(std::count(removedTxs.begin(), removedTxs.end(), fwdTx1));
     EXPECT_TRUE(std::count(removedTxs.begin(), removedTxs.end(), fwdTx2));
     EXPECT_FALSE(std::count(removedTxs.begin(), removedTxs.end(), btrTx));
 }
 
 TEST_F(SidechainsInMempoolTestSuite, ScAndFwdsAndBtrInMempool_BtrRecursiveRemoval) {
-    // Associated scenario: upon block disconnections a btr cannot make to mempool.
+    // Associated scenario: upon block disconnections a btr cannot make to mempool->
```
ptagl:
```suggestion
    // Associated scenario: upon block disconnections a btr cannot make to mempool.
```

src/gtest/test_sidechain_to_mempool.cpp
```diff
@@ -831,12 +832,12 @@ TEST_F(SidechainsInMempoolTestSuite, UnconfirmedCsw_LargerThanSidechainBalanceAr
     CTransaction cswTx = GenerateCSWTx(cswInput);
 
     CTxMemPoolEntry cswEntry(cswTx, /*fee*/CAmount(5), /*time*/ 1000, /*priority*/1.0, /*height*/1987);
-    EXPECT_TRUE(mempool.addUnchecked(cswTx.GetHash(), cswEntry));
+    EXPECT_TRUE(mempool->addUnchecked(cswTx.GetHash(), cswEntry));
 
-    // Sidechain State is Ceased and there is no Sidechain balance conflicts in the mempool. No removed Txs and Certs expected.
+    // Sidechain State is Ceased and there is no Sidechain balance conflicts in the mempool-> No removed Txs and Certs expected.
```
ptagl:
```suggestion
    // Sidechain State is Ceased and there is no Sidechain balance conflicts in the mempool. No removed Txs and Certs expected.
```

src/gtest/test_sidechain_to_mempool.cpp
```diff
@@ -881,14 +882,14 @@ TEST_F(SidechainsInMempoolTestSuite, UnconfirmedCswForAliveSidechainsAreRemovedF
...
-    EXPECT_TRUE(mempool.addUnchecked(cswTx.GetHash(), cswEntry));
+    EXPECT_TRUE(mempool->addUnchecked(cswTx.GetHash(), cswEntry));
 
     std::list<CTransaction> removedTxs;
     std::list<CScCertificate> removedCerts;
     int dummyHeight{1789};
-    // Sidechain State is Ceased and there is no Sidechain balance conflicts in the mempool. No removed Txs and Certs expected.
-    mempool.removeStaleTransactions(&sidechainsView, removedTxs, removedCerts);
-    mempool.removeOutOfScBalanceCsw(&sidechainsView, removedTxs, removedCerts);
+    // Sidechain State is Ceased and there is no Sidechain balance conflicts in the mempool-> No removed Txs and Certs expected.
```
ptagl:
```suggestion
    // Sidechain State is Ceased and there is no Sidechain balance conflicts in the mempool. No removed Txs and Certs expected.
```

src/init.cpp
```diff
@@ -1966,7 +1974,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
 #endif
 
     // Start the thread that notifies listeners of transactions that have been
-    // recently added to the mempool.
+    // recently added to the mempool->
```
ptagl:
```suggestion
    // recently added to the mempool.
```

src/rpc/blockchain.cpp
```diff
@@ -2675,15 +2675,15 @@ UniValue getcertmaturityinfo(const UniValue& params, bool fHelp)
 }
 
 /**
- * @brief Removes any transaction from the mempool.
+ * @brief Removes any transaction from the mempool->
```
ptagl:
```suggestion
 * @brief Removes any transaction from the mempool.
```

src/rpc/blockchain.cpp
```diff
@@ -2675,15 +2675,15 @@ UniValue getcertmaturityinfo(const UniValue& params, bool fHelp)
...
  */
 UniValue clearmempool(const UniValue& params, bool fHelp)
 {
     if (fHelp || params.size() != 0)
     {
         throw runtime_error(
             "clearmempool\n"
-            "\nRemoves any transaction and certificate from the mempool. Wallets are NOT synchronized.\n"
+            "\nRemoves any transaction and certificate from the mempool-> Wallets are NOT synchronized.\n"
```
ptagl:
```suggestion
            "\nRemoves any transaction and certificate from the mempool. Wallets are NOT synchronized.\n"
```

src/main.cpp
```diff
@@ -1622,7 +1625,7 @@ MempoolReturnValue AcceptTxToMemoryPool(CTxMemPool& pool, CValidationState &stat
             }
 
             // We pass pcoinsTip to IsScTxApplicableToState because we want to validate the fees against the last certificate
-            // in the blockchain, and not against certificates in the mempool.
+            // in the blockchain, and not against certificates in the mempool->
```
ptagl:
```suggestion
            // in the blockchain, and not against certificates in the mempool.
```

src/main.cpp
```diff
@@ -1795,24 +1798,26 @@ MempoolReturnValue AcceptTxToMemoryPool(CTxMemPool& pool, CValidationState &stat
             }
         }
 
-        pool.addUnchecked(hash, entry, !IsInitialBlockDownload());
+        if (!pool->addUnchecked(hash, entry, !IsInitialBlockDownload())) {
+            return MempoolReturnValue::MEMPOOL_FULL;
+        }
```
ptagl: I've noted that in case of failures of the `addUnchecked()` function inside `AcceptCertificateToMemoryPool()` we print a debug line, would it be useful also here for standard transactions?

src/main.cpp
```diff
@@ -4343,16 +4348,16 @@ bool static ConnectTip(CValidationState &state, CBlockIndex *pindexNew, CBlock *
     int64_t nTime5 = GetTimeMicros(); nTimeChainState += nTime5 - nTime4;
     LogPrint("bench", "  - Writing chainstate: %.2fms [%.2fs]\n", (nTime5 - nTime4) * 0.001, nTimeChainState * 0.000001);
 
-    // Remove conflicting transactions from the mempool.
+    // Remove conflicting transactions from the mempool->
```
ptagl:
```suggestion
    // Remove conflicting transactions from the mempool.
```

src/main.cpp
```diff
@@ -6694,33 +6699,33 @@ void static ProcessGetData(CNode* pfrom)
         // dependencies of relevant unconfirmed transactions. SPV clients want to
         // do that because they want to know about (and store and rebroadcast and
         // risk analyze) the dependencies of transactions relevant to them, without
-        // having to download the entire memory pool.
+        // having to download the entire memory pool->
```
ptagl:
```suggestion
        // having to download the entire memory pool.
```

qa/rpc-tests/mempool_size_limit.py
```diff
@@ -0,0 +1,593 @@
...
+# Copyright (c) 2018 The Zencash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.test_framework import ForkHeights
+from test_framework.authproxy import JSONRPCException
+from test_framework.util import assert_equal, initialize_chain_clean, \
+    start_nodes, sync_blocks, sync_mempools, disconnect_nodes, connect_nodes_bi, mark_logs,\
+    get_epoch_data, wait_bitcoinds, stop_nodes, \
+    swap_bytes, wait_and_assert_operationid_status
```
a-petrini: There are a few unused imports: `disconnect_nodes`, `wait_bitcoinds`, `stop_nodes`, `wait_and_assert_operationid_status`

src/main.cpp
```diff
@@ -1469,23 +1469,26 @@ MempoolReturnValue AcceptCertificateToMemoryPool(CTxMemPool& pool, CValidationSt
         }
 
         // Store transaction in memory
-        pool.addUnchecked(certHash, entry, !IsInitialBlockDownload());
+        if (!pool->addUnchecked(certHash, entry, !IsInitialBlockDownload())) {
```
ptagl: I wonder if it's possible to make this check (also) earlier in this function, just to avoid making expensive checks when the mempool is full and unable to include it.

src/main.cpp
```diff
@@ -1795,7 +1798,9 @@ MempoolReturnValue AcceptTxToMemoryPool(CTxMemPool& pool, CValidationState &stat
             }
         }
 
-        pool.addUnchecked(hash, entry, !IsInitialBlockDownload());
+        if (!pool.addUnchecked(hash, entry, !IsInitialBlockDownload())) {
+            return MempoolReturnValue::MEMPOOL_FULL;
```
a-petrini, line 1823: Just for symmetry with the `addUnchecked()` for certs, it would be nice to add an informative `LogPrint()`

src/txmempool.cpp
```diff
@@ -2009,6 +2063,80 @@ size_t CTxMemPool::DynamicMemoryUsage() const {
...
+        to_be_removed.push_back(t);
+        size_to_be_removed -= remove_candidate->first.GetBytes();
+    }
+
+    // Actually remove things from mempool
+    unsigned int total_removed = 0;
+    for (const CMemPoolEntry* r: to_be_removed) {
+        std::list<CTransaction> removed_txs;
+        std::list<CScCertificate> removed_certs;
+        if (certificate) {
```
drgora: This condition is wrong, this check should be on `r`. Will fix this with the next push, sorry about that.

src/wallet/wallet.h
```diff
@@ -1117,7 +1117,7 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
     bool AddAccountingEntry(const CAccountingEntry&, CWalletDB & pwalletdb);
 
     static CFeeRate minTxFee;
-    static CAmount GetMinimumFee(unsigned int nTxBytes, unsigned int nConfirmTarget, const CTxMemPool& pool);
+    static CAmount GetMinimumFee(unsigned int nTxBytes, unsigned int nConfirmTarget, const std::shared_ptr<CTxMemPool> pool);
```
a-petrini: To preserve the `const`-ness of `pool` variable which was present in the original function signature, I suggest 
`std::shared_ptr<const CTxMemPool> pool`

src/main.cpp
```diff
@@ -6699,33 +6699,33 @@ void static ProcessGetData(CNode* pfrom)
...
         // do that because they want to know about (and store and rebroadcast and
         // risk analyze) the dependencies of transactions relevant to them, without
-        // having to download the entire memory pool.
+        // having to download the entire memory pool->
         pfrom->PushMessage("notfound", vNotFound);
     }
 }
 
-void ProcessMempoolMsg(const CTxMemPool& pool, CNode* pfrom)
+void ProcessMempoolMsg(const std::shared_ptr<CTxMemPool> pool, CNode* pfrom)
```
a-petrini: Similarly to a following comment, to preserve the `const`-ness, I would suggest `std::shared_ptr<const CTxMemPool> pool`

a-petrini (changes requested):

src/txmempool.cpp
```diff
@@ -2009,6 +2063,80 @@ size_t CTxMemPool::DynamicMemoryUsage() const {
...
+    LOCK(cs);
+    size_t new_entry_usage = entry ? entry->GetSize() : 0;
+    assert(new_entry_usage < max_size);
+    size_t current_usage = totalTxSize + totalCertificateSize;
+    if (current_usage + new_entry_usage <= max_size) return true;
+
+    const CCertificateMemPoolEntry* certificate = dynamic_cast<const CCertificateMemPoolEntry*>(entry);
+    if (certificate && totalTxSize > m_max_size / 2) {
+        LogPrint("mempool", "%s():%d - Trying to remove transactions to make room for certificate (tot_tx: %zu, max: %zu)\n", __func__, __LINE__, totalTxSize, m_max_size / 2);
+        trimToSize(nullptr, current_usage - new_entry_usage);
```
ptagl: Would it be possible to "optimize" the mempool usage by calling `trimToSize(nullptr, max_size - new_entry_usage);`?

ptagl (changes requested):

src/wallet/wallet.cpp
```diff
@@ -3663,7 +3663,7 @@ bool CWallet::AddAccountingEntry(const CAccountingEntry& acentry, CWalletDB & pw
     return true;
 }
 
-CAmount CWallet::GetMinimumFee(unsigned int nTxBytes, unsigned int nConfirmTarget, const CTxMemPool& pool)
+CAmount CWallet::GetMinimumFee(unsigned int nTxBytes, unsigned int nConfirmTarget, const std::shared_ptr<CTxMemPool> pool)
```
titusen: Unnecessary copy constructor call. Please, pass by reference.

src/wallet/wallet.h
titusen: Unnecessary copy constructor call. Please, pass by reference.

src/txmempool.cpp
```diff
@@ -605,9 +658,11 @@ void CTxMemPool::remove(const CTransactionBase& origTx, std::list<CTransaction>&
 
     for(const uint256& hash : objToRemove)
     {
-        if (mapTx.count(hash))
+        std::map<uint256, CTxMemPoolEntry>::const_iterator entry_it;
```
titusen: Please, use shorter syntax:
`       const auto& entry_it = mapTx.find(hash);
        if (entry_it != mapTx.end())`

src/gtest/test_sidechain_to_mempool.cpp
drgora: Oh, right, a minor flaw in the otherwise brilliant `replace_all` idea :)

src/gtest/test_sidechain_to_mempool.cpp
drgora: Done

src/gtest/test_sidechain_to_mempool.cpp
drgora: Done

src/gtest/test_sidechain_to_mempool.cpp
drgora: Done

src/gtest/test_sidechain_to_mempool.cpp
drgora: Done

src/gtest/test_sidechain_to_mempool.cpp
drgora: Done

src/init.cpp
drgora: Done

src/rpc/blockchain.cpp
drgora: Done

src/rpc/blockchain.cpp
drgora: Done

src/main.cpp
drgora: Done

src/main.cpp
drgora: Done

src/main.cpp
drgora: Done

qa/rpc-tests/mempool_size_limit.py
```diff
@@ -0,0 +1,593 @@
...
+            tx_hex = tx['hex']
+            tx_input = tx['fee']
+            assert(tx_input > mininput)
+            txid = self.nodes[0].decoderawtransaction(tx_hex)['txid']
+            self.nodes[0].sendrawtransaction(tx_hex, True)
+            print("Ok")
+        except JSONRPCException as e:
+            assert(False)
+
+        prev_mpool = mpool
```
JackPiri, line 449: What about checking the evicted tx is actually the one expected?
After keeping track of tx with minfeerate in previous filling loop, with something like
```
if (feerate < minfeerate):
  minfeerate = feerate
  minfeeratetxid = txid
```
you could then check
```suggestion
prev_mpool = mpool
assert(minfeeratetxid in mpool)
mpool = self.nodes[0].getrawmempool()
assert(minfeeratetxid not in mpool)
```

qa/rpc-tests/mempool_size_limit.py
```diff
@@ -0,0 +1,593 @@
...
+            newtx = newtx + rawtx[script_position + script_length:]
+
+            signresult = self.nodes[0].signrawtransaction(newtx, None, None, "NONE")
+            tsize = len(signresult['hex']) // 2
+            tot_size += tsize
+            txFile = open(self.options.tmpdir + "/txs/tx" + str(i), "wb")
+            txFile.write(bytes(str(t['amount'] / tsize) + '\n', encoding='utf8'))
+            txFile.write(bytes(signresult["hex"], encoding='utf8'))
+            txFile.close()
+            txs.append({'fee': t['amount'] / tsize, 'hex': signresult["hex"]})
```
JackPiri: I found a way to exactly compute fee rate during data setup
```suggestion
txs.append({'fee': (t['amount'] - 128 * Decimal(0.00005816)) / tsize, 'hex': signresult["hex"]}) # 128 is outputs quantity, 5816 (xb816 reversed) is output value
```
Probably it would be good also to refactor the key name, moving from `'fee'` to `'feerate'`, also to avoid confusion with the key `'fee'` of `create_sc_certificates`, which actually refers to fee itself (not fee rate)
With above computation, I think you can also make simpler the `run_test` function, avoiding to distinguish between `tx_input` and `feerate` (and their corresponding minimum), keeping only fee rate.

src/main.cpp
drgora: Makes sense, done

src/txmempool.cpp
drgora: Makes sense, done

qa/rpc-tests/mempool_size_limit.py
JackPiri: Or possibly just extracting minimum before the assertions:
`min(feerates, key=feerates.get)`

src/init.cpp
```diff
@@ -1090,8 +1091,30 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
...
+        CAmount n = 0;
+        if (ParseMoney(mapArgs["-minrelaytxfee"], n) && n > 0)
+            ::minRelayTxFee = CFeeRate(n);
+        else
+            return InitError(strprintf(_("Invalid amount for -minrelaytxfee=<amount>: '%s'"), mapArgs["-minrelaytxfee"]));
+    }
+
+    // -mempoollimit limits
+    int64_t nMempoolSizeLimit = GetArg("-maxmempool", DEFAULT_MAX_MEMPOOL_SIZE_MB) * 1000000;
+    if (nMempoolSizeLimit < 0) {
```
JackPiri: I guess the check here should be consistent with the error message, so:
```suggestion
    if (nMempoolSizeLimit < 4 * 1000000) {
```

src/main.cpp
a-petrini: Thanks!

src/main.h
```diff
@@ -135,6 +135,8 @@ static const bool DEFAULT_ADDRESSINDEX = false;
 static const bool DEFAULT_TIMESTAMPINDEX = false;
 static const bool DEFAULT_SPENTINDEX = false;
 
+static constexpr unsigned int DEFAULT_MAX_MEMPOOL_SIZE_MB = 400;
```
JackPiri, line 138: What about moving this constant to file `txmempool.h`? It seems to me to be more scoped in that way.
I've checked, there should be no building conflict doing this (in case you agree, please double check if you can get rid of now-became-unused includes introduced by this PR, like for example for file `src/test/policyestimator_tests.cpp`).

src/init.cpp
drgora: Good catch! Done

src/wallet/wallet.h
a-petrini: Switching to `std::unique_ptr` and reverting to pass a const reference made my first comment obsolete

src/main.cpp
a-petrini: As before, latest changes made this comment obsolete

src/main.h
drgora: I agree with you it would be a great idea. But, unfortunately, right now we have all similar variables (such as `DEFAULT_MAX_ORPHAN_TRANSACTIONS`) defined in `main.h`.
I believe we really should refactor things at some point, and move these variables where they actually belong, in an organized way. But for the time being I'd rather be consistent, and leave it here.

src/txmempool.cpp
drgora: Why not, done

src/txmempool.cpp
```diff
@@ -2009,6 +2071,72 @@ size_t CTxMemPool::DynamicMemoryUsage() const {
...
+    }
+
+    // we must either reject incoming tx, or remove something
+    int64_t size_to_be_removed = current_usage + new_entry_usage - max_size;
+    std::multimap<CRawFeeRate, uint256> raw_fee_rates;
+    std::map<uint256, CRawFeeRate> rates_cache;
+    if (certificate) {
+        LogPrint("mempool", "%s():%d - Trying to remove something to make room for certificate\n", __func__, __LINE__);
+        for (const auto& cert: mapCertificate) {
+            CRawFeeRate feerate = avgFeeRateWithDeps(cert.second.GetCertificate(), true, rates_cache);
```
JackPiri: As a follow-up activity, it can be considered to cache the result even further (in addition to what is already available through `rates_cache`); I'm referring to storing the result in a member field and update it as soon as a tx/cert is added/removed.
Clearly this is valuable if a significant performance improvement can be achieved, otherwise just ignore this comment.

src/gtest/test_mempool.cpp
```diff
@@ -203,7 +203,7 @@ TEST_F(MempoolTest, TxInputLimit) {
 TEST_F(MempoolTest, OverwinterNotActiveYet) {
     // UpdateNetworkUpgradeParameters(Consensus::UPGRADE_OVERWINTER, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
 
-    CTxMemPool pool(::minRelayTxFee);
+    std::unique_ptr<CTxMemPool> pool(::minRelayTxFee, DEFAULT_MAX_MEMPOOL_SIZE_MB * 1000000);
```
JackPiri, line 206: Ridicolously minor/pedantic comment (line 216 below):
```
 EXPECT_FALSE(AcceptTxToMemoryPool(*pool, state1, tx1, LimitFreeFlag::OFF, &missingInputs, RejectAbsurdFeeFlag::OFF, DisconnectingFlag::OFF,
                                      MempoolProofVerificationFlag::SYNC));
```
(btw, it would not compile anyway :stuck_out_tongue_winking_eye:).

src/rpc/blockchain.cpp
```diff
@@ -1411,12 +1411,12 @@ UniValue getchaintips(const UniValue& params, bool fHelp)
 UniValue mempoolInfoToJSON()
```
JackPiri, line 1411: As a follow-up activity, what about providing additional info related to bytes-for-txs and bytes-for-certs?
I know currently in `mempool_size_limit.py` we keep cumulatively track of bytes/sizes, but it could be useful to have those info being required on demand.

qa/rpc-tests/mempool_size_limit.py
drgora: Good point, done

qa/rpc-tests/mempool_size_limit.py
```diff
@@ -0,0 +1,605 @@
...
+        mark_logs("Sending low fee non ceasing certificates, expecting success", self.nodes, DEBUG_MODE)
+        cert_fees = []
+        nc_certs = [self.nodes[0].sendrawtransaction(certs[2][0]['hex']),
+                    self.nodes[0].sendrawtransaction(certs[3][0]['hex'])]
+        cert_sent_size = (len(certs[2][0]['hex']) + len(certs[3][0]['hex'])) // 2
+        cert_sent = 2
+        cert_fees.append(certs[2][0]['fee'] / (len(certs[2][0]['hex']) // 2))
+        cert_fees.append(certs[3][0]['fee'] / (len(certs[3][0]['hex']) // 2))
+
+        mark_logs("Filling mpool with low fee certificates, expecting success and transaction eviction", self.nodes, DEBUG_MODE)
```
JackPiri, line 488: Similarly to previous comment, it could be worth checking that at least the lowest fee transaction is evicted (I know possibly the eviction affects multiple transactions, but the lowest fee for sure).

qa/rpc-tests/mempool_size_limit.py
```diff
@@ -0,0 +1,605 @@
...
+        mark_logs("Sending one more low fee certificate, expecting failure", self.nodes, DEBUG_MODE)
+        try:
+            c = certs[0].pop(0)
+            self.nodes[0].sendrawtransaction(c['hex'])
+            assert(False)
+        except JSONRPCException as e:
+            assert_equal(e.error['code'], -7)
+            print("Ok")
+
+        mark_logs("Sending high fee certificates, expecting success and sc0 low quality certificate eviction", self.nodes, DEBUG_MODE)
```
JackPiri: Similarly to previous comment, it could be worth checking that at least the lowest fee certificate is evicted (I know possibly the eviction affects multiple certificate, but the lowest fee for sure).

JackPiri (approved):
Wonderful job, this PR is really valuable and clean.
I left some non-blocking comments, it would be nice to have the non-big-effort ones to get addressed in this PR, the others can be considered in a follow-up PR.
Thanks,

src/txmempool.cpp
drgora: I see your point, and that was in fact my very first idea. The thing is that any modification in the mempool can potentially have an impact on the entire feerates graph.
In other words, maintaining the graph means doing a complete exploration on each modification. I'm not clear if there is a way to reduce this complexity significantly enough for this idea to bring a performance improvement.
But anyways, we should keep this idea in the radar for the future.

src/gtest/test_mempool.cpp
drgora: Done

qa/rpc-tests/mempool_size_limit.py
drgora: Good point, but I believe that would require a new data set, or using the current data set in a different way. Currently we might have multiple transactions with the same minimum fee rate, and their ordering is not defined

src/rpc/blockchain.cpp
drgora: I agree, might be a valuable addition in furure.

src/main.cpp
drgora: I believe this might be an interesting follow up improvement, but requires some design plan and modifications to the current code (for example to have a dry-run `trimToSize` which does not actually remove things from mempool.

src/txmempool.cpp
```diff
@@ -2009,6 +2071,72 @@ size_t CTxMemPool::DynamicMemoryUsage() const {
...
+    const CCertificateMemPoolEntry* certificate = dynamic_cast<const CCertificateMemPoolEntry*>(entry);
+    if (certificate && totalTxSize > m_max_size / 2) {
+        LogPrint("mempool", "%s():%d - Trying to remove transactions to make room for certificate (tot_tx: %zu, max: %zu)\n", __func__, __LINE__, totalTxSize, m_max_size / 2);
+        trimToSize(nullptr, max_size - new_entry_usage);
+        current_usage = totalTxSize + totalCertificateSize;
+        if (current_usage + new_entry_usage <= m_max_size) return true;
+    }
+
+    // we must either reject incoming tx, or remove something
+    int64_t size_to_be_removed = current_usage + new_entry_usage - max_size;
```
titusen, line 2089: Please, change these to std::unordered_map. It has better operational complexity.

src/main.h
JackPiri: Agreed to handle this in a follow-up PR.

qa/rpc-tests/mempool_size_limit.py
JackPiri: Agreed to handle this in a separate PR.

qa/rpc-tests/mempool_size_limit.py
JackPiri: Agreed to handle this in a separate PR.

src/txmempool.cpp
JackPiri: Agreed to handle this in a separate PR.

qa/rpc-tests/mempool_size_limit.py
JackPiri: Agreed to handle this in a separate PR.

qa/rpc-tests/mempool_size_limit.py
JackPiri: Agreed to handle this in a separate PR.

src/rpc/blockchain.cpp
JackPiri: Agreed to handle this in a separate PR.

qa/rpc-tests/mempool_size_limit.py
a-petrini: Nice, thanks!

a-petrini (approved):

titusen (approved):
Besides follow-up LGTM. Thanks.

src/txmempool.cpp
drgora: I agree, done with the latest set of updates.

src/wallet/wallet.cpp
drgora: This modifications has been reverted

ptagl (approved):
Looks good to me, thanks!

PR #575, opened 2023-07-12
ZEND-607 resource utilization improvement

titusen:
Message serialization is moved below preliminary validation in order to avoid unnecessary resources utilization

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -6909,6 +6909,21 @@ void ProcessTxBaseMsg(const CTransactionBase& txBase, CNode* pfrom)
     }
 }
 
+bool static ValidateStreamSize(CNode* pfrom, CDataStream& vRecv, const char* msgTypeName)
+{
+    auto vRecvStreamSz = ReadCompactSize(vRecv);
+    if (vRecvStreamSz > MAX_INV_SZ)
+    {
+        Misbehaving(pfrom->GetId(), 20);
+        return error("message %s size() = %u", msgTypeName, vRecvStreamSz);
+    }
+    else if(vRecvStreamSz * sizeof(CInv)!= vRecv.in_avail()) {
+        Misbehaving(pfrom->GetId(), 20);
+        return error("invalid message payload for %u", vRecvStreamSz);
+    }    
+    return vRecv.Rewind(GetSizeOfCompactSize(vRecvStreamSz));
+}
+
 bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, int64_t nTimeReceived)
 {
     const CChainParams& chainparams = Params();
@@ -7163,13 +7178,13 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, 
 
     else if (strCommand == "inv")
     {
+        if (!ValidateStreamSize(pfrom, vRecv, "inv"))
+        {
+            return false;
+        }
+
         vector<CInv> vInv;
         vRecv >> vInv;
-        if (vInv.size() > MAX_INV_SZ)
-        {
-            Misbehaving(pfrom->GetId(), 20);
-            return error("message inv size() = %u", vInv.size());
-        }
 
         LOCK(cs_main);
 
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -7163,13 +7163,20 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
...
+        if (vRecvStreamSz > MAX_INV_SZ)
         {
             Misbehaving(pfrom->GetId(), 20);
-            return error("message inv size() = %u", vInv.size());
+            return error("message inv size() = %u", vRecvStreamSz);
+        }
+        else if(vRecvStreamSz * sizeof(CInv)!= vRecv.in_avail()) {
+            Misbehaving(pfrom->GetId(), 20);
+            return error("invalid message payload for %u", vRecvStreamSz);
         }
```
ptagl: If I understand correctly, the meaning of this check is to increase the ban score in case the declared size is different from the real size.

At first glance, it makes sense to me, even though it is a different behavior compared to the past.

That said, do you think it's possible for a message sent by an honest peer to reach this branch of code and get banned (for instance because of corrupted data at the network level)?

src/main.cpp
```diff
@@ -7254,13 +7261,20 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
...
+            return error("message getdata size() = %u", vRecvStreamSz);
+        }
+        else if(vRecvStreamSz * sizeof(CInv)!= vRecv.in_avail()) {
+            Misbehaving(pfrom->GetId(), 20);
+            return error("invalid message payload for %u", vRecvStreamSz);
         }
+        vRecv.Rewind(GetSizeOfCompactSize(vRecvStreamSz));
+
+        vector<CInv> vInv;
+        vRecv >> vInv;
```
ptagl: I like these checks before deserializing the payload.

I see the new code is currently being used in two places (`inv` and `getdata`), and it could be useful somewhere else, would it be possible to define a generic template function to reduce duplication?

In my mind, such a function should:
- Get the stream
- Check size
- Deserialize the stream and return the related data structure (or error)

src/main.cpp
titusen: IMO very little. There would have to be a network corruption on tcp level.

src/main.cpp
titusen: done

src/main.cpp
```diff
@@ -6909,6 +6909,22 @@ void ProcessTxBaseMsg(const CTransactionBase& txBase, CNode* pfrom)
     }
 }
 
+bool static ValidateStream(CNode* pfrom, CDataStream& vRecv)
+{
+    auto vRecvStreamSz = ReadCompactSize(vRecv);
+    if (vRecvStreamSz > MAX_INV_SZ)
+    {
+        Misbehaving(pfrom->GetId(), 20);
+        return error("message inv size() = %u", vRecvStreamSz);
```
a-petrini: This is very very minor and absolutely non-blocking: I would pass `strMessage` to the `ValidateStream()` function, and replace the `inv` string in this error message. This would preserve the original behavior, logging different errors for `inv` or `getdata` messages.

a-petrini (approved):
Nice PR, thanks! I left a minor non-blocking comment, feel free to ignore

src/main.cpp
titusen: done

JackPiri (approved):
Very clear and concise PR.
Thanks

src/main.cpp
```diff
@@ -6909,6 +6909,22 @@ void ProcessTxBaseMsg(const CTransactionBase& txBase, CNode* pfrom)
     }
 }
 
+bool static ValidateStream(CNode* pfrom, CDataStream& vRecv, const char* msgTypeName)
```
drgora: Given that this function is `CInv` specific, I believe it would be clearer in future to name it something like `ValidateInvStream()`, or `ValidateInvStreamSize()`.

src/main.cpp
```diff
@@ -6909,6 +6909,22 @@ void ProcessTxBaseMsg(const CTransactionBase& txBase, CNode* pfrom)
...
+    {
+        Misbehaving(pfrom->GetId(), 20);
+        return error("message %s size() = %u", msgTypeName, vRecvStreamSz);
+    }
+    else if(vRecvStreamSz * sizeof(CInv)!= vRecv.in_avail()) {
+        Misbehaving(pfrom->GetId(), 20);
+        return error("invalid message payload for %u", vRecvStreamSz);
+    }
+    vRecv.Rewind(GetSizeOfCompactSize(vRecvStreamSz));
+    return true;
```
drgora: How about this option, which is more compact and uses the return value of `Rewind()`?
```suggestion
    return vRecv.Rewind(GetSizeOfCompactSize(vRecvStreamSz));
```

drgora (approved):
Thank you for the PR. I left couple minor comments, definitely non-blocking. LGTM.

src/main.cpp
titusen: Done

src/main.cpp
titusen: Done

drgora: Thanks!

ptagl (approved):

PR #576, opened 2023-07-19
Gp/coins selection algorithm

JackPiri:
This pull request brings in improvements in the input (coins/notes) selection phase when wallet needs to create tx/cert, in particular:
- validation of transaction size constraint (previously absent); this allows to exclude inputs sets too big that would result in a discarded tx/cert (due to oversized limit); furthermore, for z-transactions, in case of impossibility to generate the transaction, an error is immediately signalled during the inputs selection phase instead of waiting for several joinsplits computation (as was previously done)
- new coins/notes selection criteria, aiming to maximize size of selection set*; at the same time change is kept at zero or at low level, through the use of increasing change bands
- improved automatic fee selection, granting that the minimum fee is paid (previously the operation was looped with iterative estimation, resulting in higher fee rates)

For reviewing this PR, the author suggests to check the changes with the following order:
1. `transaction.h`/`wallet.h`/`wallet.cpp` (focusing on transparent part):
- the new approach of estimating the non-input parts of tx/cert is shown, followed by the actual selection of inputs;
- the fundamentals structure for tx/cert estimation are introduced and major modifications targets methods `SelectCoins` and `SelectCoinsMinConf`, new method `SelectNotes` is provided (but it is better to check later)
2. `coinsselectionalgorithm.h`/`coinsselectionalgorithm.cpp`:
- the new solvers for coins selection algorithm are provided: fast and not optimal (`SlidingWindow`), slow and optimal (`B&B`); the new algorithm for notes selection is provided: `ForNotes` (but it is better to check later)
- check the integration of those new algorithms within `SelectCoinsMinConf` method
3. `sidechainrpc.h`/`sidechainrpc.cpp`:
- check the integration of the new algorithms within `addInputs` method
4. `asyncrpcoperation_sendmany.h`/`asyncrpcoperation_sendmany.cpp`/`wallet.h`/`wallet.cpp`/`coinsselectionalgorithm.h`/`coinsselectionalgorithm.cpp`:
- major modifications targets `main_impl` method
- check the integration of new method `SelectNotes` and new algorithm `ForNotes` within `main_impl` method
5. `oversized_tx_cert`:
- the creation of oversized tx/cert is tested; the majority of positive tests were failing with previous implementation (but it was a bad failure, because in many cases a solution could be found)
6. all the rest...(very minor sparse refactoring are included in this PR)

*this aspect is crucial in terms of community goal, since the growth of the global set of UTXOs is a critical issue for any UTXO-based model (in contrast with account-based model); hence the reduction of this set should be always encouraged and welcome.

_Initial discussion available on PR #495._

### DIFF
--- a/qa/pull-tester/rpc-tests.sh
+++ b/qa/pull-tester/rpc-tests.sh
@@ -163,6 +163,7 @@ testScripts=(
   'sc_big_commitment_tree_getblockmerkleroot.py',11,25
   'p2p_ignore_spent_tx.py',215,455
   'shieldedpooldeprecation_rpc.py',558,1794
+  'oversized_tx_cert.py',300,800
 );
 
 testScriptsExt=(
--- a/qa/rpc-tests/oversized_tx_cert.py
+++ b/qa/rpc-tests/oversized_tx_cert.py
@@ -1,0 +1,373 @@
+#!/usr/bin/env python3
+import codecs
+from decimal import Decimal
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.authproxy import JSONRPCException
+from test_framework.util import assert_equal, assert_greater_or_equal_than, initialize_chain_clean,\
+    get_epoch_data, start_nodes, connect_nodes_bi, stop_node, wait_and_assert_operationid_status, swap_bytes
+import os
+import zipfile
+import time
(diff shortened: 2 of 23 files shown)


JackPiri:

PR #581, opened 2023-07-26
Refactoring for sc_cert_getblocktemplate.py

JackPiri:
- Python script performs optional check based on overall execution time
- Pyhton script optimization

### DIFF
--- a/qa/rpc-tests/sc_cert_getblocktemplate.py
+++ b/qa/rpc-tests/sc_cert_getblocktemplate.py
@@ -5,13 +5,14 @@
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 from test_framework.test_framework import BitcoinTestFramework
 from test_framework.authproxy import JSONRPCException
-from test_framework.util import assert_equal, assert_greater_than, initialize_chain_clean, \
+from test_framework.util import assert_equal, initialize_chain_clean, \
     start_nodes, sync_blocks, sync_mempools, connect_nodes_bi, mark_logs,\
     get_epoch_data, assert_false, swap_bytes
 from test_framework.test_framework import ForkHeights
 from test_framework.mc_test.mc_test import CertTestUtils, generate_random_field_element_hex
-import time
 from decimal import Decimal
+
+import time
 
 DEBUG_MODE = 1
 NUMB_OF_NODES = 3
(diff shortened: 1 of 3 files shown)


titusen (approved):
LGTM, thanks.

qa/rpc-tests/sc_cert_getblocktemplate.py
```diff
@@ -138,25 +149,30 @@ def run_test_with_scversion(self, scversion, ceasable = True):
 
         ### 1: Creating a transaction
         node2_address = self.nodes[2].getnewaddress()
-        mark_logs("\nCall GetBlockTemplate on each node to create a cached (empty) version; check it is empty", self.nodes, DEBUG_MODE)
+        mark_logs("\nCall GetBlockTemplate on each node to create a new cached version; check it is empty", self.nodes, DEBUG_MODE)
         for i in range(0, NUMB_OF_NODES):
             cacheTime[i] = time.time()
             blocktemplate = self.nodes[i].getblocktemplate()
             assert_equal(len(blocktemplate['certificates']), 0)
             assert_equal(len(blocktemplate['transactions']), 0)
```
a-petrini: If we really want to save a bit more ms, we could collect `cacheTime[i]` and `blocktemplate[i]` in this loop, and then do the assertions and print in another, separate loop.

However, my proposal is to keep the code as it is and just remove the final assertion on the number of unexpected_too_long_op. 
The reason behind that is that we cannot control how much time `getblocktemplate()` takes to provide a result on every execution environment this test is executed in, so we can expect that we would not be able to check the inclusion of txs and certs may be delayed.
I would just print a message at the end of the execution, warning that  at least one check took too long, in case this happens.

a-petrini (approved):
I left a non-blocking comment, but I would like to hear your opinion about!
No other remarks for the rest. Thanks for this PR.

src/rpc/mining.cpp
```diff
@@ -648,6 +647,8 @@ UniValue getblocktemplate(const UniValue& params, bool fHelp)
 
         // Need to update only after we know CreateNewBlockWithKey succeeded
         pindexPrev = pindexPrevNew;
+
+        nStart = GetTime(); // update last execution time as final step
```
drgora: Maybe we can rename this variable now that it's even more distant from its current name?
I'd suggest something like `nLastExecution`, or `nLastUpdated`.

qa/rpc-tests/sc_cert_getblocktemplate.py
JackPiri: Regarding the assertions I agree with you but probably the saving would be very very minimal and not worth the effort.

I have the same feeling the final assertion on `unexpected_too_long_op` should be removed, adding a warning message is a good way to at least keep track of the behaviour in case an analysis is needed. I'll go with your suggestions.

a-petrini: Totally agree! Thank you

titusen (approved):
LGTM, thanks

drgora (approved):
Looks good to me, thanks! I Only left one totally negligible comment, feel free to ignore it.

src/rpc/mining.cpp
JackPiri: Yes, it seems to me `nLastExecution` can be a better name, modified accordingly.

ptagl (approved):

PR #582, opened 2023-07-27
[ZEND-651] Improve certificates sorting in getblocktemplate

titusen:
Certificates are now sorted by fee and then by quality

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -645,6 +645,10 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn,  unsigned int nBlo
         TxPriorityCompare comparer(fSortedByFee);
         std::make_heap(vecPriority.begin(), vecPriority.end(), comparer);
 
+        // Map [SidechainID, TopQuality]
+        // Used to keep track of the top quality seen so far for any Sidechain ID
+        std::unordered_map<uint256, int64_t> topQualityCertMap;
+
         // considering certs having a higher priority than any possible tx.
         // An algorithm for managing tx/cert priorities could be devised
         while (!vecPriority.empty())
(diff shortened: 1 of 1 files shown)


src/miner.cpp
```diff
@@ -71,10 +71,10 @@ bool TxPriorityCompare::operator()(const TxPriority& a, const TxPriority& b)
             const CScCertificate& aCert = dynamic_cast<const CScCertificate&>(*a.get<2>());
```
JackPiri: Maybe the comment at lines 64-66 can be updated to reflect new logic; something like:

```
// If we are comparing two certs, we must be sure they are ordered by
// fee rate and, in case of ties, by quality.
// This criterion is a consensus rule and overrides the others two
```

src/miner.cpp
```diff
@@ -71,10 +71,10 @@ bool TxPriorityCompare::operator()(const TxPriority& a, const TxPriority& b)
...
             const CScCertificate& bCert = dynamic_cast<const CScCertificate&>(*b.get<2>());
 
-            if (aCert.GetScId() == bCert.GetScId() )
-            {
-                return aCert.quality > bCert.quality;
+            if (a.get<1>() == b.get<1>()) { 
+                return aCert.quality < bCert.quality;
```
JackPiri: Can be a third-level sorting based on priority worth? I am referring to the case where both a fee tie and a quality tie occur.

src/miner.cpp
```diff
@@ -71,10 +71,10 @@ bool TxPriorityCompare::operator()(const TxPriority& a, const TxPriority& b)
...
             }
+            return a.get<1>() < b.get<1>();
         } catch (...) {
             LogPrintf("%s():%d - ERROR: cast error\n", __func__, __LINE__ );
             assert("could not cast txbase obj" == 0);
```
JackPiri: I know it was already that way before this PR, but the equality check on double type at line 93 (`if (a.get<0>() == b.get<0>())`) is quite bad, would you consider refactoring it?

JackPiri (approved):
The PR is fine to me; please consider updating descriptive comment (lines 64-66) for aligning docs and code, the other two suggestions by me are optional and can be skipped.
Thanks

ptagl (approved):

a-petrini (approved):

drgora (approved):

PR #586, opened 2023-08-01
Ap/join network sync improvement zend 616

a-petrini:
This PR changes the logic behind the connection of two zend nodes in the test framework. In particular, we introduced a check that explicitly waits for the two nodes to exchange their respective `verack` messages before continuing with the test. This prevents racing conditions such as requesting data to a node before the connection takes place.

For this purpose, we enriched the `getpeerinfo()` RPC command with two new fields: `bytessent_per_msg` and `bytesrecv_per_msg`, two dictionaries which collect the amount of sent and received data, divided by message category.

This PR also modifies a few tests that were sporadically failing because of that racing condition.

Finally, we fixed `mempool_coinbase_spends` test which lacked proper synchronization between the two nodes.

### DIFF
--- a/qa/rpc-tests/headers_10.py
+++ b/qa/rpc-tests/headers_10.py
@@ -79,10 +79,7 @@ class headers(BitcoinTestFramework):
 #                       
 # Node(2): [0]->[1]->[2m]
 
-#        raw_input("press enter to go on..")
-
         print("\n\nJoin network")
-#        raw_input("press enter to join the netorks..")
         self.mark_logs("Joining network")
         self.join_network()
 
@@ -145,9 +142,6 @@ class headers(BitcoinTestFramework):
 # Node(2): [0]->[1]->[2m]  **Active**
 #                 \     
 #                  +->[2h]    
-
-#        raw_input("press enter to go on..")
-        time.sleep(10)
 
 if __name__ == '__main__':
     headers().main()
--- a/qa/rpc-tests/headers_10.py
+++ b/qa/rpc-tests/headers_10.py
@@ -4,19 +4,9 @@
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 from test_framework.test_framework import BitcoinTestFramework
-from test_framework.authproxy import JSONRPCException
-from test_framework.util import assert_equal, initialize_chain_clean, \
-    start_nodes, start_node, connect_nodes, stop_node, stop_nodes, \
-    sync_blocks, sync_mempools, connect_nodes_bi, wait_bitcoinds, p2p_port, check_json_precision, \
-    disconnect_nodes
-import traceback
-import os,sys
-import shutil
-from random import randint
-from decimal import Decimal
-import logging
-
+from test_framework.util import initialize_chain_clean, start_nodes, sync_blocks
 import time
+
 class headers(BitcoinTestFramework):
 
     def setup_chain(self, split=False):
@@ -94,9 +84,8 @@ class headers(BitcoinTestFramework):
         print("\n\nJoin network")
 #        raw_input("press enter to join the netorks..")
         self.mark_logs("Joining network")
-        self.join_network_sync()
+        self.join_network()
 
-#        time.sleep(10)
         sync_blocks(self.nodes, 1, True, 15)
 
         for i in range(0, 3):
--- a/qa/rpc-tests/nodehandling.py
+++ b/qa/rpc-tests/nodehandling.py
@@ -8,7 +8,7 @@
 #
 
 from test_framework.test_framework import BitcoinTestFramework
-from test_framework.util import assert_equal, connect_nodes_bi, p2p_port,connect_nodes
+from test_framework.util import assert_equal, p2p_port,connect_nodes
 
 import time
 
@@ -55,7 +55,7 @@ class NodeHandlingTest (BitcoinTestFramework):
         for node in self.nodes[0].getpeerinfo():
             assert(node['addr'] != url.hostname+":"+str(p2p_port(1)))
 
-        connect_nodes_bi(self.nodes,0,1) #reconnect the node
+        connect_nodes(self.nodes, 0, 1) #reconnect the node
         found = False
         for node in self.nodes[0].getpeerinfo():
             if node['addr'] == url.hostname+":"+str(p2p_port(1)):
@@ -88,4 +88,4 @@ class NodeHandlingTest (BitcoinTestFramework):
 
 
 if __name__ == '__main__':
-    NodeHandlingTest ().main ()
+    NodeHandlingTest().main()
--- a/qa/rpc-tests/nodehandling.py
+++ b/qa/rpc-tests/nodehandling.py
@@ -76,7 +76,7 @@ class NodeHandlingTest (BitcoinTestFramework):
         start_time = time.time()
         connections_count = 0
         while time.time() - start_time < total_duration:
-            connect_nodes(self.nodes[0], 0)
+            connect_nodes(self.nodes, 0, 0)
             connections_count += 1
             time.sleep(0.1)
             for node in self.nodes[0].getpeerinfo():
--- a/qa/rpc-tests/test_framework/test_framework.py
+++ b/qa/rpc-tests/test_framework/test_framework.py
@@ -12,6 +12,7 @@ import sys
 import shutil
 import tempfile
 import traceback
+from enum import Enum, auto
 
 from test_framework.authproxy import JSONRPCException
 from test_framework.wsproxy import JSONWSException
@@ -30,6 +31,10 @@ ForkHeights = {
     'SHIELDED_POOL_DEPRECATION': 990
 }
 
+class syncType(Enum):
+    SYNC_BLOCKS = auto()
+    SYNC_MEMPOOLS = auto()
+    SYNC_ALL = auto()
 
 class BitcoinTestFramework(object):
 
@@ -82,16 +87,15 @@ class BitcoinTestFramework(object):
             sync_blocks(self.nodes)
             sync_mempools(self.nodes)
 
-    def join_network(self, id = 1):
+    def join_network(self, id = 1, sync_type = syncType.SYNC_BLOCKS):
         # Join the (previously split) adjanced nodes in linear network together: 0-1 >-< 2-3
         assert self.is_network_split
         connect_nodes_bi(self.nodes, id, id + 1)
+        if sync_type == syncType.SYNC_BLOCKS or sync_type == syncType.SYNC_ALL:
+            sync_blocks([self.nodes[id], self.nodes[id + 1]])
+        if sync_type == syncType.SYNC_MEMPOOLS or sync_type == syncType.SYNC_ALL:
+            sync_mempools([self.nodes[id], self.nodes[id + 1]])
         self.is_network_split = False
-
-    def join_network_sync(self, id = 1):
-        # Join the (previously split) adjanced nodes in linear network together: 0-1 >-< 2-3 and syncs both parts
-        self.join_network(id)
-        self.sync_all()
 
     def main(self):
         import optparse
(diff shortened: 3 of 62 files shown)


src/net.h
```diff
@@ -501,7 +505,8 @@ class CNode
         try
         {
             BeginMessage(pszCommand);
-            EndMessage();
+            size_t bytesSent = EndMessage();
+            AccountForSentBytes(pszCommand, bytesSent);
```
ptagl: I like the idea here, it made me think of a possible different approach:
- keep `void EndMessage()` (instead of returning a `size_t`) to keep it similar to `BeginMessage()`
- Call `AccountForSentBytes()` directly inside `EndMessage()` so that we remove some "duplication"

What do you think?

src/net.h
```diff
@@ -697,6 +711,22 @@ class CNode
...
+    {
+        mapSendBytesPerMsgType[msg_type] += sent_bytes;
+    }
+
+    void AccountForRecvBytes(const std::string& msg_type, size_t recv_bytes)
+        EXCLUSIVE_LOCKS_REQUIRED(cs_vRecvMsg)
+    {
+        auto msgTypeInMap = mapRecvBytesPerMsgType.find(msg_type);
+        if (msgTypeInMap == mapRecvBytesPerMsgType.end())
+            msgTypeInMap = mapRecvBytesPerMsgType.find("*other*");
```
ptagl: Can we use the message type instead of the string value itself?

```suggestion
            msgTypeInMap = mapRecvBytesPerMsgType.find(NetMsgType::OTHER);
```

qa/rpc-tests/test_framework/util.py
```diff
@@ -318,27 +318,51 @@ def set_node_times(nodes, t):
...
+
 def wait_bitcoinds():
     # Wait for all bitcoinds to cleanly exit
     for bitcoind in bitcoind_processes.values():
         bitcoind.wait()
     bitcoind_processes.clear()
 
-def connect_nodes(from_connection, node_num):
-    ip_port = "127.0.0.1:"+str(p2p_port(node_num))
+def connect_nodes(nodes, a, b):
```
ptagl, line 343: What about giving the same signature to `disconnect_nodes()` to keep consistency?

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -86,13 +86,9 @@ def join_network(self, id = 1):
         # Join the (previously split) adjanced nodes in linear network together: 0-1 >-< 2-3
         assert self.is_network_split
         connect_nodes_bi(self.nodes, id, id + 1)
+        sync_blocks([self.nodes[id], self.nodes[id + 1]])
```
ptagl: Given that most of the times we rely on a full sync, can we give the possibility to choose how to perform the join? For instance:
- input parameter for `join_network()` to sync blocks only, mempool only, sync all
- by default, sync_all

titusen: IMO, if a subset of these is needed, they can be called directly.

qa/rpc-tests/headers_10.py
```diff
@@ -94,9 +84,8 @@ def run_test(self):
         print("\n\nJoin network")
 #        raw_input("press enter to join the netorks..")
```
titusen: Please, remove the comment

titusen (approved):
Nice cleanup. LGTM, thanks.

src/net.h
a-petrini: I like the second proposal more. I modified the PR accordingly, however `EndMessage()` signature needed to be modified as well, since `AccountForSentBytes()` needs the message type.

src/net.h
a-petrini: Yes, that's a better option, thanks!

qa/rpc-tests/test_framework/util.py
a-petrini: Ok, I also took the chance to remove all the occurrences of `import disconnect_nodes` whenever it is never used in the test itself

qa/rpc-tests/headers_10.py
a-petrini: Removed, thanks!

qa/rpc-tests/test_framework/util.py
```diff
@@ -318,27 +318,51 @@ def set_node_times(nodes, t):
...
-        zero_version_peers = [peer for peer in peer_info if peer['version'] == 0]
-        if len(zero_version_peers) == 0:
-            break
+    # poll until version handshake completes to avoid race conditions with transaction relaying
+    # Handshake completes after each peer receives a verack from the other one
+    if a != b:  # connecting to self causes immediate disconnection, so avoid doing the following checks
+        wait_until(lambda: sum(peer['version'] != 0 for peer in from_connection.getpeerinfo()) == from_num_peers)
+        wait_until(lambda: sum(peer['version'] != 0 for peer in to_connection.getpeerinfo()) == to_num_peers)
+        wait_until(lambda: sum(peer['bytesrecv_per_msg'].pop('verack', 0) == 24 for peer in from_connection.getpeerinfo()) == from_num_peers)
+        wait_until(lambda: sum(peer['bytesrecv_per_msg'].pop('verack', 0) == 24 for peer in to_connection.getpeerinfo()) == to_num_peers)
```
drgora: I believe it would be easier to maintain this in future, if we replace `24` with a named variable and a brief description of how these 24 bytes are composed.

src/main.cpp
```diff
@@ -6601,7 +6601,7 @@ void static ProcessGetData(CNode* pfrom)
                     if (inv.type == MSG_BLOCK)
                     {
                         LogPrint("forks", "%s():%d - Pushing block [%s]\n", __func__, __LINE__, block.GetHash().ToString() );
-                        pfrom->PushMessage("block", block);
+                        pfrom->PushMessage(NetMsgType::BLOCK, block);
                     }
                     else // MSG_FILTERED_BLOCK)
                     if (inv.type == MSG_FILTERED_BLOCK)
```
JackPiri, line 6606: Probably we want to modify also line 6613 (`pfrom->PushMessage("merkleblock", merkleBlock);`) leveraging `NetMsgType` (also need to add the `"merkleblock"` command in `protocol.h`/`protocol.cpp`).

src/net.h
```diff
@@ -362,6 +364,8 @@ class CNode
     ~CNode();
 
 private:
+    std::map<std::string, uint64_t> mapSendBytesPerMsgType;
```
JackPiri: I think it can be even more useful/simple to keep track of messages count (per message type); in case we could have the value of the map being an `std::pair<uint64_t,uint64_t>` in order to provide both the information (bytes count and messages count).
In case you agree, please consider similar comment for `CNodeStats`.

qa/rpc-tests/nodehandling.py
```diff
@@ -55,7 +55,7 @@ def run_test(self):
         for node in self.nodes[0].getpeerinfo():
             assert(node['addr'] != url.hostname+":"+str(p2p_port(1)))
 
-        connect_nodes_bi(self.nodes,0,1) #reconnect the node
```
JackPiri, line 58: Just a double check for making sure everything is fine: we moved from `connect_nodes_bi` to `connect_nodes`, is it ok?

src/rpc/net.cpp
```diff
@@ -101,6 +101,19 @@ UniValue getpeerinfo(const UniValue& params, bool fHelp)
             "    \"lastrecv\": ttt,                      (numeric) the time in seconds since epoch (Jan 1 1970 GMT) of the last receive\n"
             "    \"bytessent\": n,                       (numeric) the total bytes sent\n"
             "    \"bytesrecv\": n,                       (numeric) the total bytes received\n"
+            "    \"bytessent_per_msg\":  {\n"
+            "        msg,                                (string) the total bytes sent aggregated by message type.\n"
```
JackPiri: Not sure here the fields `\"bytessent_per_msg\"` and `\"bytesrecv_per_msg\"` are properly documented, in my opinion some more details on the fact it's a dictionary of keys as strings and values as numerics should be provided.

qa/rpc-tests/test_framework/util.py
```diff
@@ -318,27 +318,51 @@ def set_node_times(nodes, t):
...
+
+    while attempt < attempts and elapsed < timeout:
+        if predicate():
+            return True
+        attempt += 1
+        elapsed += 0.1
+        time.sleep(0.1)
+
+    return False
+
```
JackPiri: Given this utility function can be reused in the future and the `predicate()` evaluation can potentially take some time to execute, why not refactoring the time control (`timeout` and `elapsed`) so that it actually checks for executing time?
Something like:

```suggestion
def wait_until(predicate, attempts=float('inf'), timeout=float('inf')):
    attempt = 0
    elapsed = time.time()

    while attempt < attempts and (time.time() - elapsed) < timeout:
        if predicate():
            return True
        attempt += 1
        time.sleep(0.1)

    return False
```

JackPiri (approved):
The PR is fine to me, it should definitely solve our sync errors during test and that's great.
I left few minor comments; probably only the one related to `PushMessage` in `main.cpp` really to be addressed, the others can be skipped.

src/main.cpp
a-petrini: Yes, nice catch!

src/net.h
a-petrini: I like the idea, even though there is no practical use for that. I implemented this change in the last commit.

qa/rpc-tests/nodehandling.py
a-petrini: Yes, the functionality and limitation of `connect_nodes_bi` changed slightly with this PR

src/rpc/net.cpp
a-petrini: Done, I modified the help message also for accommodating the changes of the maps themselves

qa/rpc-tests/test_framework/util.py
a-petrini: Yes, it sounds reasonable and I like this improvement

src/net.cpp
```diff
@@ -2356,6 +2359,13 @@ CNode::CNode(SOCKET hSocketIn, const CAddress& addrIn, const std::string& addrNa
     else
         LogPrint("net", "Added connection peer=%d\n", id);
 
+    for (size_t i = 0; i < allNetMessageTypesSize; i++) {
+        std::pair<uint64_t, uint64_t>& messageTypeSend = mapSendBytesPerMsgType[allNetMessageTypes[i]];
+        messageTypeSend.first = messageTypeSend.second = 0;
+        std::pair<uint64_t, uint64_t>& messageTypeRecv = mapRecvBytesPerMsgType[allNetMessageTypes[i]];
+        messageTypeRecv.first = messageTypeRecv.second = 0;
```
drgora: Tiny thing here, but how about something like the following?
```suggestion
        mapSendBytesPerMsgType.insert(allNetMessageTypes[i], {0, 0});
        mapRecvBytesPerMsgType.insert(allNetMessageTypes[i], {0, 0});
```

qa/rpc-tests/test_framework/test_framework.py
drgora: I'll just add my point of view here, but definitely not a blocking thing for me.
I like @ptagl's suggestion, it adds more flexibility for test developers and it allows writing more uniform and clean tests, with minimum impact on the code.

drgora (approved):
I left a couple of minuscule comments, feel free to ignore. Looks good to me, thanks!

qa/rpc-tests/test_framework/util.py
a-petrini: Good suggestion, modified accordingly

src/net.cpp
a-petrini: Way cleaner, I like that.

qa/rpc-tests/test_framework/test_framework.py
a-petrini: I modified the function to improve its flexibility. Now synchronization of blocks and mempools are optional, with the former enabled by default, and the latter disabled.

ptagl (approved):

PR #588, opened 2023-08-03
Overflow/underflow check for -maxsendbuffer and -maxreceivebuffer

JackPiri:
Handling cases like `-maxsendbuffer=-1`, `-maxsendbuffer=0`, `-maxsendbuffer=4294967296`, etc...

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -378,8 +378,8 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-listen", _("Accept connections from outside (default: 1 if no -proxy or -connect)"));
     strUsage += HelpMessageOpt("-listenonion", strprintf(_("Automatically create Tor hidden service (default: %d)"), DEFAULT_LISTEN_ONION));
     strUsage += HelpMessageOpt("-maxconnections=<n>", strprintf(_("Maintain at most <n> connections to peers (default: %u)"), DEFAULT_MAX_PEER_CONNECTIONS));
-    strUsage += HelpMessageOpt("-maxreceivebuffer=<n>", strprintf(_("Maximum per-connection receive buffer, <n>*1000 bytes (default: %u)"), 5000));
-    strUsage += HelpMessageOpt("-maxsendbuffer=<n>", strprintf(_("Maximum per-connection send buffer, <n>*1000 bytes (default: %u)"), 1000));
+    strUsage += HelpMessageOpt("-maxreceivebuffer=<n>", strprintf(_("Maximum per-connection receive buffer, <n>*1000 bytes (default: %u)"), DEFAULT_MAX_RECEIVE_BUFFER));
+    strUsage += HelpMessageOpt("-maxsendbuffer=<n>", strprintf(_("Maximum per-connection send buffer, <n>*1000 bytes (default: %u)"), DEFAULT_MAX_SEND_BUFFER));
     strUsage += HelpMessageOpt("-onion=<ip:port>", strprintf(_("Use separate SOCKS5 proxy to reach peers via Tor hidden services (default: %s)"), "-proxy"));
     strUsage += HelpMessageOpt("-onlynet=<net>", _("Only connect to nodes in network <net> (ipv4, ipv6 or onion)"));
     strUsage += HelpMessageOpt("-permitbaremultisig", strprintf(_("Relay non-P2SH multisig (default: %u)"), 1));
@@ -859,8 +859,6 @@ bool AppInitServers()
 namespace {
     uint64_t nLocalServices = NODE_NETWORK;
     int nMaxConnections = DEFAULT_MAX_PEER_CONNECTIONS;
-    unsigned int nSendBufferMaxSize;
-    unsigned int nReceiveFloodSize;
 } // namespace
 
 /** Initialize bitcoin.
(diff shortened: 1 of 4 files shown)


a-petrini (approved):
Looks good, thanks!

ptagl (changes requested):
Thanks for the effort in improving checks on input parameters!

To me, the changes seem in line with the typical approach used in that file, I was wondering if we could start thinking of a new approach to separate parameters validation from usage.

In particular, I wonder if we could introduce a new function like `ValidateStartupArguments()` that takes care of performing all the sanity checks on the inputs while keeping `AppInit2()` lite and more readable.

Given that we may be using similar validations for several arguments in the future, we may also think of including a second util function like `ValidateRanges(argKey, [optional] min, [option] max)` to reduce code duplication.

In such a way, we could probably also easily increase code coverage with some unit tests.

What do you think?

src/init.cpp
```diff
@@ -1976,9 +1974,17 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
 
     connOptions.nLocalServices      = nLocalServices;
     connOptions.nMaxConnections     = nMaxConnections;
-    connOptions.nSendBufferMaxSize  = 1000*GetArg("-maxsendbuffer", 1*1000);
-    connOptions.nReceiveFloodSize   = 1000*GetArg("-maxreceivebuffer", 5*1000);
-
+    int64_t msb_unbound = GetArg("-maxsendbuffer", 1*1000);
+    if (msb_unbound > 0 && msb_unbound <= std::numeric_limits<unsigned int>::max())
+        connOptions.nSendBufferMaxSize = 1000 * static_cast<unsigned int>(msb_unbound); // from bytes to kilobytes
```
drgora: Two things here (and also for line `1984`):
- given that we multiply it by `1000`, I believe we should check that `msg_unbound <= std::numeric_limits<unsigned int>::max() / 1000`, otherwise it would overflow
- I believe the comment is misleading. Shouldn't it be the other way round, `kilobytes to bytes`?

drgora (changes requested):
Thank you for the PR! I left a small comment that I believe should be addressed before merging.
I also second @ptagl on his comment on modularization.

src/init.cpp
JackPiri: Yes, agree on both. Modified accordingly.

drgora: Thank you!

titusen (approved):
I agree with Paolos comments. The rest LGTM, thanks.

src/init.cpp
```diff
@@ -1976,9 +1974,18 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
 
     connOptions.nLocalServices      = nLocalServices;
     connOptions.nMaxConnections     = nMaxConnections;
-    connOptions.nSendBufferMaxSize  = 1000*GetArg("-maxsendbuffer", 1*1000);
-    connOptions.nReceiveFloodSize   = 1000*GetArg("-maxreceivebuffer", 5*1000);
-
+    unsigned int fromKBtoB = 1000; // from kilobytes to bytes
+    int64_t msb_unbound = GetArg("-maxsendbuffer", DEFAULT_MAX_SEND_BUFFER);
+    if (msb_unbound > 0 && msb_unbound <= std::numeric_limits<unsigned int>::max() / fromKBtoB)
```
AndreaLanfranchi: I would personally enforce a more strict range of values. While this check solves the underflow issue we must consider this is a limit per connection and the way it is checked it allows up to 2Gigabytes unprocessed messages per socket.
I understand this is a condition unlikely to happen but ... as always, better safe than sorry.
Maybe a hard limit of, say, 50MeBi is appropriate ?

ptagl: Thanks for the feedback! The idea totally makes sense, and we evaluated such possibility, but in the end discarded it for a few reasons:
- passing a negative number is definitely a mistake, but passing a huge value may not (it depends on the context)
- it's hard to find a proper max limit and it seems a little subjective (why 50 MB? why not 100 or higher?)
- a max limit that makes sense today may be too low in the next years and would require continuous adjustments to the code

So, given the points above, and that users who change that value are supposed to be experts and should know what they are doing, we prefer to reject only values that are clearly wrong (like negative ones).

AndreaLanfranchi: Acknowledged. Let me add some comments here:

> and that users who change that value are supposed to be experts

this IMHO should not be assumed: on the contrary we should be providing a software solution that is "stupid proof" to help spread the node instances even for "curious". If instead the aim is to address mainly professional node operators then I cannot disagree but it seems it goes agains a decentralization ethos

> it's hard to find a proper max limit and it seems a little subjective (why 50 MB? why not 100 or higher?)

Given the number of max connections allowed an overall allotment of "unprocessed messages" is IMHO advisable: it helps identify why we may get in such an unfortunate solution

> a max limit that makes sense today may be too low in the next years and would require continuous adjustments to the code

I would expect the code to be continuosly adjusted in the next years: we should not be afraid of the workload required to continuosly improve and adjust the UX.

Last but not least : actual implementation might allow the insertion of a value as low as 1 which results in a 1'000 byte "spamming threshold". In such a condition a node sending a full block of 2MB would be deemed as spammer and further messages rejected. At the very least I would enforce a value which is GE than max message size.

src/util.cpp
```diff
@@ -400,6 +400,23 @@ int64_t GetArg(const std::string& strArg, int64_t nDefault)
     return nDefault;
 }
 
+int64_t GetArgWithinLimits(const std::string& strArg, int64_t nDefault, std::vector<std::pair<int64_t, int64_t>> limits)
```
drgora: I think it might be worth to pass `limits` as a const reference here.

src/init.cpp
```diff
@@ -1979,9 +1977,12 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
 
     connOptions.nLocalServices      = nLocalServices;
     connOptions.nMaxConnections     = nMaxConnections;
-    connOptions.nSendBufferMaxSize  = 1000*GetArg("-maxsendbuffer", 1*1000);
-    connOptions.nReceiveFloodSize   = 1000*GetArg("-maxreceivebuffer", 5*1000);
 
+    unsigned int fromKBtoBfactor = 1000; // from kilobytes to bytes
+    std::vector<std::pair<int64_t, int64_t>> limits = {{1, std::numeric_limits<unsigned int>::max() / fromKBtoBfactor}};
```
drgora: Very small thing here, but I would suggest being more explicit on the usage/meaning of these variables, and make them `static const`.

drgora (approved):
This PR looks good to me. I left a couple small comments that I believe might be good addressing before merge, but not blocking. Thank you.

src/util.cpp
JackPiri: Done

src/init.cpp
JackPiri: Made them `const` (I think `static` is not needed being `AppInit2` called at startup only) and renamed more properly.

ptagl (approved):

PR #591, opened 2023-08-07
Fix for message header validity

JackPiri:
Fixed validation logic for command (rejected if null), moved header/message validation from `ProcessMessage` to `ReceiveMsgBytes`.
Partially inspired by https://github.com/bitcoin/bitcoin/pull/19107.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -7922,41 +7922,12 @@ bool ProcessMessages(CNode* pfrom, const std::atomic<bool>& interruptMsgProc)
         // at this point, any failure means we can delete the current message
         it++;
 
-        // Scan for message start
-        if (memcmp(msg.hdr.pchMessageStart, Params().MessageStart(), MESSAGE_START_SIZE) != 0) {
-            LogPrintf("PROCESSMESSAGE: INVALID MESSAGESTART %s peer=%d\n", SanitizeString(msg.hdr.GetCommand()), pfrom->id);
-            fOk = false;
-            break;
-        }
-
-        // Read header
-        CMessageHeader& hdr = msg.hdr;
-        if (!hdr.IsValid(Params().MessageStart()))
-        {
-            LogPrintf("PROCESSMESSAGE: ERRORS IN HEADER %s peer=%d\n", SanitizeString(hdr.GetCommand()), pfrom->id);
-            continue;
-        }
-        string strCommand = hdr.GetCommand();
-
-        // Message size
-        unsigned int nMessageSize = hdr.nMessageSize;
-
-        // Checksum
-        CDataStream& vRecv = msg.vRecv;
-        uint256 hash = Hash(vRecv.begin(), vRecv.begin() + nMessageSize);
-        unsigned int nChecksum = ReadLE32((unsigned char*)&hash);
-        if (nChecksum != hdr.nChecksum)
-        {
-            LogPrintf("%s(%s, %u bytes): CHECKSUM ERROR nChecksum=%08x hdr.nChecksum=%08x\n", __func__,
-               SanitizeString(strCommand), nMessageSize, nChecksum, hdr.nChecksum);
-            continue;
-        }
-
         // Process message
+        string strCommand = msg.hdr.GetCommand();
         bool fRet = false;
         try
         {
-            fRet = ProcessMessage(pfrom, strCommand, vRecv, msg.nTime, interruptMsgProc);
+            fRet = ProcessMessage(pfrom, strCommand, msg.vRecv, msg.nTime, interruptMsgProc);
             if (interruptMsgProc)
                 return true;
         }
@@ -7966,12 +7937,12 @@ bool ProcessMessages(CNode* pfrom, const std::atomic<bool>& interruptMsgProc)
             if (strstr(e.what(), "end of data"))
             {
                 // Allow exceptions from under-length message on vRecv
-                LogPrintf("%s(%s, %u bytes): Exception '%s' caught, normally caused by a message being shorter than its stated length\n", __func__, SanitizeString(strCommand), nMessageSize, e.what());
+                LogPrintf("%s(%s, %u bytes): Exception '%s' caught, normally caused by a message being shorter than its stated length\n", __func__, SanitizeString(strCommand), msg.hdr.nMessageSize, e.what());
             }
             else if (strstr(e.what(), "size too large"))
             {
                 // Allow exceptions from over-long size
-                LogPrintf("%s(%s, %u bytes): Exception '%s' caught\n", __func__, SanitizeString(strCommand), nMessageSize, e.what());
+                LogPrintf("%s(%s, %u bytes): Exception '%s' caught\n", __func__, SanitizeString(strCommand), msg.hdr.nMessageSize, e.what());
             }
             else
             {
@@ -7985,7 +7956,7 @@ bool ProcessMessages(CNode* pfrom, const std::atomic<bool>& interruptMsgProc)
         }
 
         if (!fRet)
-            LogPrintf("%s(%s, %u bytes) FAILED peer=%d\n", __func__, SanitizeString(strCommand), nMessageSize, pfrom->id);
+            LogPrintf("%s(%s, %u bytes) FAILED peer=%d\n", __func__, SanitizeString(strCommand), msg.hdr.nMessageSize, pfrom->id);
 
         break;
     }
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -682,10 +682,8 @@ bool CNode::ReceiveMsgBytes(const char *pch, unsigned int nBytes)
             handled = msg.readData(pch, nBytes);
 
         if (handled < 0)
-                return false;
-
-        if (msg.in_data && msg.hdr.nMessageSize > MAX_PROTOCOL_MESSAGE_LENGTH) {
-            LogPrint("net", "Oversized message from peer=%i, disconnecting\n", GetId());
+        {
+            LogPrint("net", "Invalid %s from peer=%i, disconnecting\n", (!msg.in_data ? "header" : "message"), GetId());
             return false;
         }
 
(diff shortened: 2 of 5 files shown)


src/net.cpp
```diff
@@ -754,6 +766,18 @@ int CNetMessage::readData(const char *pch, unsigned int nBytes)
...
+        // reject messages with invalid checksum
+        uint256 hash = Hash(vRecv.begin(), vRecv.begin() + hdr.nMessageSize);
+        unsigned int nChecksum = ReadLE32((unsigned char*)&hash);
+        if (nChecksum != hdr.nChecksum)
+        {
+            LogPrint("net" "Message error: invalid checksum nChecksum=%08x hdr.nChecksum=%08x\n", __func__, nChecksum, hdr.nChecksum);
+            return -1;
+        }
+    }
+
```
a-petrini: I have a bit of concern about moving the checksum evaluation from `ProcessMessages` to `CNetMessage::readData` function, as it implicitly moves the execution context of this check from `threadMessageHandler` to `threadSocketHandler`.
As now it does not make much difference, as the presence of the `cs_vRecvMsg` lock guarantees that only one of the two threads is running, but in the future we plan to disjoint the two threads. This will imply that zend will read the socket and process the messages (almost) asynchronously; for this reason, I suppose that `threadMessageHandler` should be as quick as possible in reading the socket, leveraging all the heavy checks to the message processing thread.
Unfortunately, it's very hard to forecast how leaving this check here would affect performance, but I am open to discussion whether leave it as it is for the moment, and make a change only after the final refactoring.

a-petrini (commented):
Thanks for the PR, I like how the checks have been rationally grouped and moved!
However, I left a comment that could be worth discussing about

src/net.cpp
```diff
@@ -725,12 +723,29 @@ int CNetMessage::readHeader(const char *pch, unsigned int nBytes)
...
+    // reject messages with invalid command
+    if (!hdr.HasValidCommand())
+    {        
+        LogPrint("net", "Header error: invalid command (%s)\n", SanitizeString(hdr.GetCommand()));
+        return -1;
+    }
+
+    // reject messages larger than MAX_SERIALIZED_COMPACT_SIZE or MAX_PROTOCOL_MESSAGE_LENGTH
+    if (hdr.nMessageSize > MAX_SERIALIZED_COMPACT_SIZE || hdr.nMessageSize > MAX_PROTOCOL_MESSAGE_LENGTH)
```
AndreaLanfranchi: Just nitpicking: considering MAX_PROTOCOL_MESSAGE_LENGTH is waaay lower than MAX_SERIALIZED_COMPACT_SIZE maybe reversing the order of checks would make sense.

titusen (approved):
LGTM, thanks

src/net.cpp
```diff
@@ -725,12 +723,29 @@ int CNetMessage::readHeader(const char *pch, unsigned int nBytes)
         hdrbuf >> hdr;
     }
     catch (const std::exception&) {
+        LogPrint("net", "Header error: failed to deserialize");
```
drgora: It looks like we're missing the newline here.

src/net.cpp
```diff
@@ -725,12 +723,29 @@ int CNetMessage::readHeader(const char *pch, unsigned int nBytes)
...
         return -1;
     }
 
-    // reject messages larger than MAX_SERIALIZED_COMPACT_SIZE
-    if (hdr.nMessageSize > MAX_SERIALIZED_COMPACT_SIZE)
-            return -1;
+    // reject messages with invalid start
+    if (memcmp(hdr.pchMessageStart, Params().MessageStart(), MESSAGE_START_SIZE) != 0) {
+        LogPrint("net", "Header error: invalid start (%s)", SanitizeString(std::string(hdr.pchMessageStart, hdr.pchMessageStart + MESSAGE_START_SIZE)));
```
drgora: Also here, missing newline at the end of the string.

src/net.cpp
```diff
@@ -751,6 +766,18 @@ int CNetMessage::readData(const char *pch, unsigned int nBytes)
...
     nDataPos += nCopy;
 
+    if (complete())
+    {
+        // reject messages with invalid checksum
+        uint256 hash = Hash(vRecv.begin(), vRecv.begin() + hdr.nMessageSize);
+        unsigned int nChecksum = ReadLE32((unsigned char*)&hash);
+        if (nChecksum != hdr.nChecksum)
+        {
+            LogPrint("net" "Message error: invalid checksum nChecksum=%08x hdr.nChecksum=%08x\n", __func__, nChecksum, hdr.nChecksum);
```
drgora: I believe that format specifiers do not correspond to the arguments here, I guess you meant to remove `__func__`?

src/net.cpp
drgora: I strongly agree, this point is very interesting. Once we disentangle the two threads, we will have more room to balance processing tasks on different threads, for our workloads.
But for the time being, personally I'm ok with moving it to the new location.

drgora (approved):
Looks good to me, thanks. I left few comments that I believe should be applied before merging, but nothing worth further discussion.

src/net.cpp
JackPiri: Done

src/net.cpp
JackPiri: Yes you are right, updated accordingly.

src/net.cpp
JackPiri: Added newline.

src/net.cpp
JackPiri: Added newline.

src/net.cpp
```diff
@@ -725,12 +723,29 @@ int CNetMessage::readHeader(const char *pch, unsigned int nBytes)
...
+        LogPrint("net", "Header error: invalid command (%s)\n", SanitizeString(hdr.GetCommand()));
+        return -1;
+    }
+
+    // reject messages larger than MAX_PROTOCOL_MESSAGE_LENGTH or MAX_SERIALIZED_COMPACT_SIZE
+    if (hdr.nMessageSize > MAX_PROTOCOL_MESSAGE_LENGTH || hdr.nMessageSize > MAX_SERIALIZED_COMPACT_SIZE)
+    {
+        LogPrint("net", "Header error: size too large (%u bytes)\n", hdr.nMessageSize);
+        return -1;
+    }
```
ptagl: With the new approach, we achieved the goal of **detecting invalid messages earlier** (before passing them to the higher-level thread for processing), but we are also defining a new handling of errors.

Let's consider the following checks:
- **Valid message start**
- **Header validity**
- **Checksum**

**Before** this PR, they were performed by the `ProcessMessages()` function, and in case of failure, the **message was** just **discarded**. **Now**, this PR moves these checks to the `threadSocketHandler()` level, and in case of failure, the **peer** is **disconnected**.

As emerged in other discussions recently, that could be a problem because a failed check (e.g. invalid checksum) isn't necessarily caused by a malicious actor. It's very unlikely, but some network errors can silently "bypass" the TCP checksum and reach the application layer, **eventually causing the ban of honest peers**.

I like moving checks earlier in the application flow, but given the reasons explained above, I would rather keep the original behavior in terms of rejected messages, without banning nodes that may be honest.

JackPiri: I get your point. Technically we are not banning in case of failure of those checks, only triggering a disconnection, but clearly is a much more punishing behaviour with respect to just discarding the message.
Since I prefer the new position the checks have been moved, I've prepared some modifications in order to provide the same behavior we had before in case of failing check.
Just summarizing:

1. undeserializable header -> discard (don't disconnect)
2. wrong header start -> discard (don't disconnect)
3. invalid command -> discard (don't disconnect)
4. out-threshold payload size (in header) -> discard (don't disconnect)
5. unmatching checksum -> discard (don't disconnect)

Question is @ptagl : **for the first and the fourth check, do we want to be more assertive (disconnect, as before) or be permissive (discard, don't disconnect)?** In my opinion both cases can fall inside the case of unintentional wrong message (like for network errors), hence I would opt for being permissive (discard, don't disconnect).

titusen (approved):
LGTM, thanks.

src/net.cpp
ptagl: @JackPiri after reviewing further the PR I would propose the following behavior:

1. Inside `readHeader()`:
- In case of exception, return -1 and then disconnect (already done)
- In case of wrong message start, return -1 and then disconnect (to be adjusted)
- In case of size exceeded, return -1 and then disconnect (already done)
- No call to `hdr.HasValidCommand()` (see next sections)

2. Restore the previous behavior of `readData()`:
- Remove checksum computation

3. If message is `complete()`:
- Validate the checksum (maybe the related code can be encapsulated into the `CNetMessage` class)?
- Call `hdr.HasValidCommand()`
- If any of the previous two checks fail, discard the message and count it in the statistics as "unknown" message (using `NetMsgType::OTHER`)
- If the two checks pass, then continue as usual.


I think that this should give us the same behavior as before and hopefully simplify a bit the readability of this section of code. What do you think?

src/main.cpp
```diff
@@ -7919,55 +7919,25 @@ bool ProcessMessages(CNode* pfrom)
...
-               SanitizeString(strCommand), nMessageSize, nChecksum, hdr.nChecksum);
-            continue;
-        }
-
         // Process message
         bool fRet = false;
         try
         {
-            fRet = ProcessMessage(pfrom, strCommand, vRecv, msg.nTime);
+            fRet = ProcessMessage(pfrom, msg.hdr.GetCommand(), msg.vRecv, msg.nTime);
```
a-petrini: Just a minor non-blocking comment, as it may simplify the changes: I suggest to declare and define a string reference to `msg.hdr.GetCommand()`to avoid calling `GetCommand()` each time.

src/net.cpp
JackPiri: This should be addressed with commit afcb3a6cf7cf2200eedf25b020aae0f034a75e03.

src/main.cpp
JackPiri: Done.

ptagl (approved):

PR #593, opened 2023-08-22
[ZEND-668] unittests create new block

titusen:
Unit tests for CreateNewBlock. Note that last test is disabled due to it trying to find correct nonce.


src/primitives/transaction.h
```diff
@@ -1056,9 +1056,9 @@ struct CMutableTransactionBase
     int32_t nVersion;
     std::vector<CTxIn> vin;
 
-protected:
-    std::vector<CTxOut> vout;
+// protected:
 public:
+    std::vector<CTxOut> vout;
```
JackPiri: Please restore `vout` protected accessibility and use public functions `addOut` and `getOut` within test `test_createnewblock.cpp`.

JackPiri (changes requested):

src/gtest/tx_creation_utils.cpp
```diff
@@ -668,15 +668,15 @@ MempoolReturnValue BlockchainTestManager::TestAcceptTxToMemoryPool(CValidationSt
 {
     CCoinsViewCache* saved_pcoinsTip = pcoinsTip;
 
-    std::unique_ptr<CTxMemPool> pool(new CTxMemPool(::minRelayTxFee, DEFAULT_MAX_MEMPOOL_SIZE_MB * 1000000));
```
JackPiri: While I totally agree with modification (from global `mempool` to local `pool`) contained in `main.cpp`, I think this function should be left unchanged; better to use locally created `pool` instead of resetting each time global `mempool`.

What I am proposing is supported by:

- other tests in this module using the same approach (`TxInputLimit`, `SproutV3TxFailsAsExpected`, `SproutV3TxWhenGrothNotActive`, `SproutNegativeVersionTx`)
- just tried running all gtests on my machine and confirm what is written above


Even if it can be worth to have a deeper understanding of the failure you encountered, with my proposal there is no need to modify tests like `src/wallet/gtest/test_wallet_zkeys.cpp` and `src/wallet/gtest/test_deadlock.cpp`.

JackPiri (changes requested):

src/Makefile.gtest.include
```diff
@@ -10,6 +10,7 @@ zen_gtest_SOURCES = \
 	gtest/tx_creation_utils.cpp \
 	gtest/test_checktransaction.cpp \
 	gtest/test_consensus.cpp \
+	gtest/test_createnewblock.cpp \
```
JackPiri: Please move this test to the bottom as reported below:

```
...
	gtest/test_asyncproofverifier.cpp \
	gtest/test_blockdownload.cpp \
	gtest/test_createnewblock.cpp
...
```

In this way we can simplify the PR so that the following modifications can be reverted (not needed anymore):
- `src/gtest/test_mempool.cpp`
- `src/util.cpp/src/util.h`
- `src/wallet/gtest/test_deadlock.cpp`
- `src/wallet/gtest/test_wallet_zkeys.cpp`

I have tested multiple times on my machine running full test suite as well as single tests and no error shows; also the CI is fine.

As a final comment, for sure we can consider making the test suite more stable and reliable in a separate PR; #1539 (https://github.com/zcash/zcash/issues/1539) from ZCash repo for reference.

JackPiri (changes requested):
Please address comment related to test order.
Thanks

src/gtest/tx_creation_utils.cpp
titusen: Done

src/Makefile.gtest.include
titusen: Done

drgora: I second @JackPiri's suggestion, but I cannot see it implemented in the code. Could you please double check?

src/gtest/tx_creation_utils.cpp
drgora: I second @JackPiri's suggestion, but I cannot see it implemented in the code. Could you please double check?

src/gtest/test_createnewblock.cpp
```diff
@@ -0,0 +1,678 @@
+// Copyright (c) 2011-2014 The Bitcoin Core developers
```
JackPiri, line 1: Please consider modifying this file based on following comments:

- `CreateNewBlock_1tx`: can be removed since it is based on the submission of a single transaction, furthermore the transaction is invalid (it is rejected in `CreateNewBlock` after failing check `HaveInputs`)
- `CreateNewBlock_1cert`: similar to above.
- `CreateNewBlock_validity`: can be removed since it is already available (commented) in `miner_tests.cpp` and it is actually skipped.
- `CreateNewBlock_sorting_certs`: currently all the certificates are being rejected (in `CreateNewBlock` after failing check `HaveInputs` or free-fee check), hence there is no sorting at all inside the provided block; this test should be modified so that all the three certificates are accepted, then a check on the returned block should be performed in order to verify correct sorting of certificates.

JackPiri (changes requested):
Please modify `test_createnewblock.cpp` as per comment.
Thanks

JackPiri:

PR #594, opened 2023-08-24
[ZEND-552] Remove stale transactions when a hard fork activates 

ptagl:
The `RemoveStaleTransaction()` function is currently called every time a block is connected/disconnected to eventually remove transactions that became invalid.

To keep the flow as efficient as possible, such function doesn't run all the checks available inside `AcceptTxToMemoryPool()` but only the ones that could really fail and depend on the _context_ (like the chain height, or the state of a sidechain).

This PR includes a small fix to apply additional checks in case a hard fork activates, because the change in the set of validation rules may affect pending transactions (for instance, a shielding transaction will not be valid anymore after the activation of hard fork #11).

### DIFF
--- a/qa/rpc-tests/mempool_hard_fork_cleaning.py
+++ b/qa/rpc-tests/mempool_hard_fork_cleaning.py
@@ -4,19 +4,22 @@
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
 
+from test_framework.blockchainhelper import EXPECT_SUCCESS, BlockchainHelper, SidechainParameters
 from test_framework.test_framework import BitcoinTestFramework, ForkHeights
-from test_framework.util import assert_equal, assert_greater_than, initialize_chain_clean, \
+from test_framework.util import assert_equal, assert_greater_than, initialize_chain_clean, mark_logs, \
                                 start_nodes, sync_blocks, connect_nodes_bi, wait_and_assert_operationid_status
 
+DEBUG_MODE = 1
+NUMB_OF_NODES = 2
 Z_FEE = 0.0001
 
 class Test (BitcoinTestFramework):
 
     def setup_chain(self):
         print("Initializing test directory "+self.options.tmpdir)
-        initialize_chain_clean(self.options.tmpdir, 2)
+        initialize_chain_clean(self.options.tmpdir, NUMB_OF_NODES)
 
-    def setup_network(self, split=False):
+    def setup_network(self):
         self.nodes = start_nodes(2, self.options.tmpdir, [['-experimentalfeatures', '-zmergetoaddress']] * 2)
         connect_nodes_bi(self.nodes, 0, 1)
         self.is_network_split=False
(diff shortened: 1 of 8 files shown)


qa/rpc-tests/mempool_hard_fork_cleaning.py
```diff
@@ -0,0 +1,128 @@
...
+        mark_logs("Split the network", self.nodes, DEBUG_MODE)
+        self.split_network(0)
+
+        mark_logs("Node 0 creates a shielding transaction", self.nodes, DEBUG_MODE)
+        total_funds_before_shielding = sum(x['amount'] for x in self.nodes[0].listunspent(0))
+        opid = self.nodes[0].z_mergetoaddress(["ANY_TADDR"], self.nodes[0].z_getnewaddress(), Z_FEE, 2, 2)["opid"]
+        wait_and_assert_operationid_status(self.nodes[0], opid)
+
+        # Give time to the wallet to be notified and updated the balance (listunspent would otherwise return an outdated value)
+        wait_until(self.nodes[0].getmempoolinfo()['fullyNotified'] == True, 20)
```
a-petrini: `wait_until()` takes a callable argument, but the current one - `self.nodes[0].getmempoolinfo()['fullyNotified'] == True` - is not, hence it generates a few warnings like these:
`Warning: exception "'bool' object is not callable" occured in test_framework.util`

I suggest to change that to:
`wait_until(lambda: self.nodes[0].getmempoolinfo()['fullyNotified'] == True, 20)`

a-petrini (approved):
Looks good to me, thanks for this PR! I only left a minor non-blocking comment regarding the Python test.

qa/rpc-tests/mempool_hard_fork_cleaning.py
```diff
@@ -0,0 +1,128 @@
...
+NUMB_OF_NODES = 2
+Z_FEE = 0.0001
+
+class Test (BitcoinTestFramework):
+
+    def setup_chain(self):
+        print("Initializing test directory "+self.options.tmpdir)
+        initialize_chain_clean(self.options.tmpdir, NUMB_OF_NODES)
+
+    def setup_network(self):
```
titusen: You can remove setup_network method and just add
def setup_nodes(self):
         self.nodes = start_nodes(NUMB_OF_NODES, self.options.tmpdir, [['-experimentalfeatures', '-zmergetoaddress', '-limitdebuglogsize=false']] * NUMB_OF_NODES)

qa/rpc-tests/mempool_hard_fork_cleaning.py
ptagl: Nice catch, thanks! My original version wasn't even working as expected, it was in practice like a `time.sleep(2)`, now it's much better!

qa/rpc-tests/mempool_hard_fork_cleaning.py
ptagl: I like it, included in the latest push. Thanks!

titusen (approved):
LGTM, thanks.

PR #599, opened 2023-08-31
codacy issues fixes

titusen:


JackPiri (approved):
The PR is fine to me.
Thanks

src/zcash/util.cpp
```diff
@@ -15,13 +15,12 @@ std::vector<unsigned char> convertIntToVectorLE(const uint64_t val_int) {
 // Convert bytes into boolean vector. (MSB to LSB)
 std::vector<bool> convertBytesVectorToVector(const std::vector<unsigned char>& bytes) {
     std::vector<bool> ret;
-    ret.resize(bytes.size() * 8);
+    ret.resize(bytes.size() * __CHAR_BIT__);
```
drgora: It seems that `__CHAR_BIT__` is not part of the c++ standard.
I would suggest using `CHAR_BIT`, which is guaranteed by the standard to be defined in `climits`.

titusen: done

drgora (approved):
Looks good to me, thanks

src/zcash/util.cpp
```diff
@@ -15,13 +16,12 @@ std::vector<unsigned char> convertIntToVectorLE(const uint64_t val_int) {
...
     std::vector<bool> ret;
-    ret.resize(bytes.size() * 8);
+    ret.resize(bytes.size() * CHAR_BIT);
 
-    unsigned char c;
     for (size_t i = 0; i < bytes.size(); i++) {
-        c = bytes.at(i);
-        for (size_t j = 0; j < 8; j++) {
-            ret.at((i*8)+j) = (c >> (7-j)) & 1;
+        unsigned char c = bytes.at(i);
```
a-petrini, line 22: Why moving the declaration of the `c` variable inside the loop?

a-petrini (approved):
Looks good, I only left a minor question

src/zcash/util.cpp
titusen: Cause it's not used outside the loop.

a-petrini: Yes, true. I am more concerned by the optimization matter, as we are re-declaring the variable at every iteration. I think that the compiler would optimize it, however.

a-petrini: After further investigation, I see there is no impact on the performance, even without compiler optimization.

PR #600, opened 2023-09-01
Gp/release scripts fix

JackPiri:
Minor fixes for making release scripts more robust and flexible.

### DIFF
--- a/contrib/devtools/release_management/release_management_utils.py
+++ b/contrib/devtools/release_management/release_management_utils.py
@@ -60,8 +60,12 @@ def git_create_branch(cwd: str, branch_name: str):
 def git_commit(cwd: str, commit_title: str, directories_for_add: list[str] = []):
     for directory_for_add in directories_for_add:
         subprocess.run(["git", "add", directory_for_add], cwd=cwd) # Add changes to the index
-    result = subprocess.run(["git", "commit", "-a", "-S", "-m", commit_title], capture_output=True, text=True, cwd=cwd) # Commit changes with the given message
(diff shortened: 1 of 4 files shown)


contrib/devtools/release_management/release_management_utils.py
```diff
@@ -179,10 +179,17 @@ def insert_line_into_file(file_path: str, line_number: int, line_content: str):
     with open(file_path, "w") as file:
         file.writelines(lines)
 
-def ask_for_step_skip(script_step: str, config: dict):
-    config[k_script_steps][script_step] = {}
-    config[k_script_steps][script_step][k_stop] = False
-    if (input("Do you want to skip this step? (Y/N)").upper() == "Y"):
+def ask_for_step_skip(script_step: str, config: dict, step_details: str = ""):
```
a-petrini: The usage of the `config` variable is a bit confusing: it is declared and initialized as global in the other two scripts, hence explicitly passing that shouldn't be mandatory. 
Before this change, this function was called without explicitly passing the dict.

I would suggest to declare and define this variable in the global space of this file, and import that in the two other scripts. This would be enough for removing `config` from this function signature + each invocation, without affecting the visibility on the other scripts.

a-petrini (approved):
Looks good, there's a minimal change that I suggest and it's worth investigating, at least for a bit.

contrib/devtools/release_management/release_management_utils.py
a-petrini: I saw the new change, looks good to me

ptagl (approved):

PR #601, opened 2023-09-04
Update mempool_size_limit.py to fix issue with feerate

drgora:
This PR fixes a minor inconsistency in the fee vs. feerate values stored in snapshots created for testing mempool size limitations.
The snapshot is updated accordingly.

Also, few helper functions are moved out of the local class, so that they can be seamlessly used by other tests.


a-petrini (approved):
Thanks for the update, looks good to me.
I'll use the refactored function for cleaning up the tests on my side.

JackPiri (approved):
The PR is fine to me (explicitly checked fee rates and everything is ok).
Thanks

PR #602, opened 2023-09-05
Ap/accept tx to mempool call consistency

a-petrini:


drgora (approved):
Looks good to me, thanks

JackPiri (approved):
The PR is fine to me.
Just a question: is it ok that in `test_asyncproofverifier.cpp` and `test_reindex.cpp` we reset both `mempool` and `connman`, while in `test_sidechain_blocks.cpp`, `test_wallet.cpp` and `test_wallet_cert.cpp` we reset `mempool` only?
Thanks

ptagl (approved):

PR #611, opened 2023-09-26
Update mempool size limit code and test

drgora:
This PR fixes minor issues which caused the size limit to be exceeded temporarily under specific circumstances.

In addition, the blockchain snapshot used for tests `mempool_size_limit.py` and `mempool_size_limit_more.py` has been moved to backblaze and removed from this repository.

### DIFF
--- a/qa/rpc-tests/mempool_size_limit.py
+++ b/qa/rpc-tests/mempool_size_limit.py
@@ -9,10 +9,6 @@ from test_framework.authproxy import JSONRPCException
 from test_framework.util import assert_equal, initialize_chain_clean, \
     start_nodes, sync_blocks, sync_mempools, connect_nodes_bi, mark_logs,\
     get_epoch_data, hex_str_to_bytes, swap_bytes, download_snapshot
-#from test_framework.script import CScript, CScriptNum, OP_CHECKBLOCKATHEIGHT, OP_RETURN, OP_PUSHDATA1, OP_PUSHDATA2, OP_1
-#from test_framework.mininode import CTransaction, ToHex, ser_string
-#from io import BytesIO
-#from binascii import unhexlify, hexlify
 from test_framework.mc_test.mc_test import *
 import os
 import zipfile
@@ -83,17 +79,20 @@ def create_tx_script(script_size, amount):
     for i in range(bytes_per_out):
         script_pubkey += "01"
 
-    # Push fake a hash and height on the stack, followed by checkblockatheight
+    # Push a fake hash and height on the stack, followed by checkblockatheight
     script_pubkey += "4c20"  # PUSH1 32bytes
     script_pubkey += "0000000000000000000000000000000000000000000000000000000000000000" # 0
     script_pubkey += "51"  # PUSH 1
     script_pubkey += "b4" # OP_CHECKBLOCKATHEIGHT
     ## concatenate NUM_OUT txouts of above script_pubkey which we'll insert before the txout for change
     txouts = b2x(varlenEncode(NUM_OUT))
-    amount_str = b2x((amount_to_satoshi(amount) // NUM_OUT).to_bytes(8, 'little'))
+    satoshis_amount = amount_to_satoshi(amount) // NUM_OUT
+    amount_str = b2x((satoshis_amount).to_bytes(8, 'little'))
+    satoshis_amount_remainder = satoshis_amount + (amount_to_satoshi(amount) % NUM_OUT)
+    amount_str_remainder = b2x((satoshis_amount_remainder).to_bytes(8, 'little'))
     for k in range(NUM_OUT):
         # add txout value
-        txouts += amount_str
+        txouts += amount_str if (k > 0) else amount_str_remainder
         # add length of script_pubkey
         txouts += b2x(varlenEncode(len(script_pubkey)//2))
         # add script_pubkey
@@ -103,6 +102,10 @@ def create_tx_script(script_size, amount):
 
 '''
 TODO: replace the default ones with the _noraw variants of the functions below
+#from test_framework.script import CScript, CScriptNum, OP_CHECKBLOCKATHEIGHT, OP_RETURN, OP_PUSHDATA1, OP_PUSHDATA2, OP_1
+#from test_framework.mininode import CTransaction, ToHex, ser_string
+#from io import BytesIO
+#from binascii import unhexlify, hexlify
 def create_tx_script_noraw(script_size, amount):
     bytes_per_out = script_size // NUM_OUT
 
@@ -163,7 +166,7 @@ def create_big_transaction(node, tmpdir, input_t, script_size, fee, i):
     if (tmpdir is not None):
         write_to_file(tmpdir + "/tx" + str(i), feerate, signresult["hex"])
     tx = {'feerate': feerate, 'hex': signresult["hex"]}
-    return tx
+    return tx, tsize
 
 
 def create_big_transactions(node, tmpdir, utxos, size):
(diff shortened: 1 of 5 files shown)


qa/rpc-tests/mempool_size_limit_more.py
```diff
@@ -6,9 +6,9 @@
 from test_framework.test_framework import BitcoinTestFramework
 from test_framework.authproxy import JSONRPCException
 from test_framework.util import assert_equal, initialize_chain_clean, \
-    start_nodes, sync_blocks,connect_nodes_bi, mark_logs, \
+    start_nodes, sync_blocks,connect_nodes_bi, mark_logs, download_snapshot, \
     colorize as cc
-from mempool_size_limit import load_from_storage, satoshi_round, create_tx_script
+from mempool_size_limit import load_from_storage, satoshi_round, create_big_transaction
```
a-petrini: `connect_nodes_bi` and `satoshi_round` are no longer required (this could not be caused by this refactoring, and I am pretty sure they could have been removed even before that)

src/txmempool.cpp
```diff
@@ -609,15 +610,14 @@ std::vector<uint256> CTxMemPool::mempoolDependenciesOf(const CTransactionBase& o
     return res;
 }
 
-CRawFeeRate CTxMemPool::avgFeeRateWithDeps(const uint256& rootTx, const bool certificatesAllowed, std::unordered_map<uint256, CRawFeeRate>& cache) const
+std::pair<CRawFeeRate, std::vector<uint256>> CTxMemPool::avgFeeRateWithDeps(const uint256& rootTx, const std::vector<CTxIn>& vin, const bool certificatesAllowed, std::unordered_map<uint256, CRawFeeRate>& cache) const
```
JackPiri, line 613: Can be worth removing all the null-check on `entry` and leaving just one at function start? Something like:
```
if (!entry)
  return true;
```

qa/rpc-tests/mempool_size_limit.py
```diff
@@ -8,12 +8,17 @@
 from test_framework.authproxy import JSONRPCException
 from test_framework.util import assert_equal, initialize_chain_clean, \
     start_nodes, sync_blocks, sync_mempools, connect_nodes_bi, mark_logs,\
-    get_epoch_data, swap_bytes
+    get_epoch_data, hex_str_to_bytes, swap_bytes, download_snapshot
+#from test_framework.script import CScript, CScriptNum, OP_CHECKBLOCKATHEIGHT, OP_RETURN, OP_PUSHDATA1, OP_PUSHDATA2, OP_1
```
JackPiri: Can we get rid of those unused `import`s?

qa/rpc-tests/mempool_size_limit.py
```diff
@@ -66,60 +73,128 @@ def load_from_storage(path):
...
-        script_length = int(rawtx[script_position - 2 : script_position], 16) * 2
-
-        newtx = rawtx[0:out_num_pos + 8]
-        newtx = newtx + txscript
-        newtx = newtx + rawtx[script_position + script_length:]
-
-        signresult = node.signrawtransaction(newtx, None, None, "NONE")
-        tsize = len(signresult['hex']) // 2
+        tx = create_big_transaction(node, tmpdir + "/txs", utxos.pop(), 512 * NUM_OUT, fee, i)
         tot_size += tsize
```
JackPiri, line 181: Here `tsize` should be replaced with `len(tx["hex"]) // 2`.

qa/rpc-tests/mempool_size_limit.py
```diff
@@ -66,60 +73,128 @@ def load_from_storage(path):
...
+    script_pubkey += "0000000000000000000000000000000000000000000000000000000000000000" # 0
+    script_pubkey += "51"  # PUSH 1
+    script_pubkey += "b4" # OP_CHECKBLOCKATHEIGHT
+    ## concatenate NUM_OUT txouts of above script_pubkey which we'll insert before the txout for change
+    txouts = b2x(varlenEncode(NUM_OUT))
+    amount_str = b2x((amount_to_satoshi(amount) // NUM_OUT).to_bytes(8, 'little'))
+    for k in range(NUM_OUT):
         # add txout value
-        txouts += "b816000000000000"
+        txouts += amount_str
```
JackPiri: A bit pedantic, so please double check if it's actually worth: should the remainder of the division `amount` by `NUM_OUT` be taken in consideration and being added to the last entry? If this doesn't conflict with fee/feerate just skip this comment.

JackPiri (approved):
The PR is fine to me, please just address comment https://github.com/HorizenOfficial/zen/pull/611/files#r1340382257, the others can be ignored if you prefer.
Thanks

qa/rpc-tests/mempool_size_limit.py
```diff
@@ -194,28 +270,33 @@ def create_chained_transactions(node, tmpdir, utxos):
     signresult = node.signrawtransaction(rawtx, None, None, "NONE")
     csize = len(signresult['hex']) // 2
     feerate = low_fee / csize
+    assert(abs(csize - EXP_SIZE) <= 1)
 
-    write_to_file(tmpdir + "/ctxs/ctx0", feerate, signresult["hex"])
-    txs.append({'feerate': high_fee / csize, 'hex': signresult["hex"]})
+    write_to_file(tmpdir + "/tx0", feerate, signresult["hex"])
+    txs.append({'feerate': feerate, 'hex': signresult["hex"]})
```
a-petrini, line 279: Just a check, not a real thing to point out. The fact the we used `high_fee` before the update was less correct than what happens now?

qa/rpc-tests/mempool_size_limit.py
a-petrini, line 181: ...or let `create_big_transaction()` return `tsize` as well, as it is on `create_big_transaction_noraw()`

qa/rpc-tests/mempool_size_limit.py
```diff
@@ -66,60 +73,128 @@ def load_from_storage(path):
...
+    return tx
+
+
+def create_big_transactions(node, tmpdir, utxos, size):
     txs = []
     tot_size = 0
     i = 0
     os.makedirs(tmpdir + "/txs")
+    minfeerate, maxfeerate = MAX_FEE, Decimal("0")
+    fee = Decimal(0.5) + Decimal(0.01) * i
```
a-petrini: Just a minor thing, as this might not impact the test at all, but to consider for future tests.
From `Decimal` module documentation:
`Decimal instances can be constructed from integers, strings, floats, or tuples. Construction from an integer or a float performs an exact conversion of the value of that integer or float.`
It implies that `Decimal(0.01)` is actually converted to `Decimal('0.01000000000000000020816681711721685132943093776702880859375')`, which in the long run may lead to wrong results in case of comparisons.
If we want to avoid the conversion, I suggest to use the constructor that takes a string as input parameter, i.e. `Decimal('0.01')`

a-petrini (approved):
thanks for improving this feature! I left a few non-blocking comments and approved the PR

src/txmempool.cpp
JackPiri: Please ignore this, already resolved.

qa/rpc-tests/mempool_size_limit.py
drgora: Indeed, the previous version was not consistent, as `tx0` was actually using `low_fee`, and not `high_fee`. It is correct to use `low_fee` for the first transaction (which by itself becomes a good eviction candidate), and `high_fee` for the second one (which increases the feerate of the dependencies chain of `tx0`).

qa/rpc-tests/mempool_size_limit.py
drgora: Yes, good point, I moved them next to the `create_tx_script_noraw()` implementation (where they would be needed), which is commented out and left for future reference.

qa/rpc-tests/mempool_size_limit_more.py
drgora: Good point, done.

qa/rpc-tests/mempool_size_limit.py
drgora: Yes, this was the intended implementation, thank you

drgora: Good catch, indeed I forgot to update this. I implemented @a-petrini's suggestion here, which was the intended implementation.

qa/rpc-tests/mempool_size_limit.py
drgora: Great point! Applied your suggestion to all the occurrences that I found in the tests touched by this PR.

qa/rpc-tests/mempool_size_limit.py
a-petrini: Clear! Thanks for explaining!

qa/rpc-tests/mempool_size_limit.py
drgora: Done in the last update

qa/rpc-tests/mempool_size_limit.py
a-petrini: thanks!

ptagl (approved):

PR #612, opened 2023-09-27
[ZEND-569] Remove boost threads from networking system

a-petrini:
This PR removes all the instances of `boost::thread` from the networking subsystem, in favor of STL `<thread>` library.
One key difference between the libraries is that `boost::thread`s are interruptible; this feature has been replicated by the newly introduced class `CThreadInterrupt`.
The PR also removes the `boost::condition_variable` as well, preferring the STL equivalent.
Finally, `boost::thread_group` has been removed as well, but it has not been replaced with anything equivalent.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -6535,7 +6535,7 @@ bool static AlreadyHave(const CInv& inv) EXCLUSIVE_LOCKS_REQUIRED(cs_main)
     return true;
 }
 
-void static ProcessGetData(CNode* pfrom)
+void static ProcessGetData(CNode* pfrom, const std::atomic<bool>& interruptMsgProc)
 {
     std::deque<CInv>::iterator it = pfrom->vRecvGetData.begin();
 
@@ -6550,7 +6550,8 @@ void static ProcessGetData(CNode* pfrom)
 
         const CInv &inv = *it;
         {
-            boost::this_thread::interruption_point();
+            if (interruptMsgProc)
+                return;
             it++;
 
             if (inv.type == MSG_BLOCK || inv.type == MSG_FILTERED_BLOCK)
@@ -6957,7 +6958,7 @@ bool static ValidateStreamSize(CNode* pfrom, CDataStream& vRecv, const char* msg
     return vRecv.Rewind(GetSizeOfCompactSize(vRecvStreamSz));
 }
 
-bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, int64_t nTimeReceived)
+bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, int64_t nTimeReceived, const std::atomic<bool>& interruptMsgProc)
 {
     const CChainParams& chainparams = Params();
     LogPrint("net", "%s() - received: %s (%u bytes) peer=%d\n", __func__, SanitizeString(strCommand), vRecv.size(), pfrom->id);
@@ -7139,7 +7140,8 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, 
 
         BOOST_FOREACH(CAddress& addr, vAddr)
         {
-            boost::this_thread::interruption_point();
+            if (interruptMsgProc)
+                return true;
 
             static constexpr bool rate_limited = true;
              // Apply rate limiting.
@@ -7227,7 +7229,8 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, 
         {
             const CInv &inv = vInv[nInv];
 
-            boost::this_thread::interruption_point();
+            if (interruptMsgProc)
+                return true;
             pfrom->AddInventoryKnown(inv);
 
             bool fAlreadyHave = AlreadyHave(inv);
@@ -7319,7 +7322,7 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, 
         }
 
         pfrom->vRecvGetData.insert(pfrom->vRecvGetData.end(), vInv.begin(), vInv.end());
-        ProcessGetData(pfrom);
+        ProcessGetData(pfrom, interruptMsgProc);
     }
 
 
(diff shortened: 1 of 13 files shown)


src/main.cpp
```diff
@@ -6535,7 +6535,7 @@ bool static AlreadyHave(const CInv& inv) EXCLUSIVE_LOCKS_REQUIRED(cs_main)
     return true;
 }
 
-void static ProcessGetData(CNode* pfrom)
+void static ProcessGetData(CNode* pfrom, std::atomic<bool>& interruptMsgProc)
```
drgora: Just a small thing here, I would suggest making `interruptMsgProc` a const ref, to make it clear that it is read only from this function.
Same thing for all the other functions that take a reference to `interruptMsgProc` in `main.cpp`.

drgora (approved):
Thank you for the PR. Great improvement with getting rid of all the `Millisleep`s around, and of unneeded boost threads!
Looks good to me!

src/main.cpp
a-petrini: Thanks! I agree with your comment. Code changes have been updated as suggested.

src/net.cpp
```diff
@@ -24,9 +24,6 @@
 #include <fcntl.h>
 #endif
 
-#include <boost/filesystem.hpp>
-#include <boost/thread.hpp>
-
 #include <openssl/conf.h>
 #include <openssl/ssl.h>
 #include <openssl/err.h>
```
JackPiri, line 30: It is something that was already present, but it seems to me here (line 30) we should include `<zen/tlsmanager.h>` (instead of `<zen/tlsmanager.cpp>`) and move `extern std::unique_ptr<CConnman> connman;` from `<zen/tlsmanager.cpp>` file to `<zen/tlsmanager.h>` file.

src/net.cpp
```diff
@@ -773,15 +765,27 @@ CConnman::CConnman() {
...
+        threadNonTLSPoolsCleaner.join();
     NetCleanup();
 };
 
 CConnman::~CConnman()
 {
     try
     {
-        StopNode();
+        if (!alreadyStopped)
```
JackPiri: I don't understand why we call `StopNode` both in `Shutdown` (`init.cpp`) and in `~CConnman`. Since singleton `connman` is declared as global variable in `main.cpp`, we always call in first place `ShutDown` and later destroy `connman`, so we can remove call to `StopNode` from the destructor. This would also allow to remove the variable `alreadyStopped`.

JackPiri (changes requested):
Please consider addressing the two comments I left, they could make things a little bit simpler to read and maintain.
Thanks

src/net.cpp
```diff
@@ -773,15 +765,27 @@ CConnman::CConnman() {
...
+    if (threadOpenConnections.joinable())
+        threadOpenConnections.join();
+    if (threadOpenAddedConnections.joinable())
+        threadOpenAddedConnections.join();
+    if (threadDNSAddressSeed.joinable())
+        threadDNSAddressSeed.join();
+    if (threadSocketHandler.joinable())
+        threadSocketHandler.join();
+    if (threadNonTLSPoolsCleaner.joinable())
+        threadNonTLSPoolsCleaner.join();
```
ptagl, line 779: What about including all the thread references into an array and creating a loop to iterate and call `joinable()` and `join()` (and save some lines of code)?

src/net.cpp
```diff
@@ -1403,17 +1416,18 @@ void ThreadSocketHandler()
             }
             FD_ZERO(&fdsetSend);
             FD_ZERO(&fdsetError);
-            MilliSleep(timeout.tv_usec/1000);
+            if (!interruptNet.sleep_for(std::chrono::milliseconds(timeout.tv_usec/1000)))
+                return;
```
ptagl: Does it make sense here to use `std::chrono::microseconds` and avoid the division?

src/threadinterrupt.h
```diff
@@ -0,0 +1,34 @@
...
+*/
+class CThreadInterrupt
+{
+public:
+    explicit operator bool() const;
+    void operator()();
+    void reset();
+    bool sleep_for(std::chrono::milliseconds rel_time);
+    bool sleep_for(std::chrono::seconds rel_time);
+    bool sleep_for(std::chrono::minutes rel_time);
```
ptagl: Can we get rid of the duplication here and just declare `bool sleep_for(std::chrono::nanoseconds rel_time);`?

The conversion from other duration types would be automatic then, and we should be safe from overflows as long as we stay below 292 years as the maximum duration, but please, double check in case you decide to proceed this way.

ptagl (changes requested):

src/net.cpp
a-petrini: Nice catch, I updated the include directives as suggested.

src/net.cpp
a-petrini: I agree: the original intention was to follow Bitcoin PR, and avoiding calling `StopNode()` twice, but it make sense just to call it once from `init.cpp`'s `Shutdown()`.

src/threadinterrupt.h
a-petrini: Even better, now `sleep_for` has been templetized, and accepts any `std::chrono::duration` type, leveraging `wait_for()` function for the conversion across different duration types.

src/net.cpp
a-petrini: Done, converted to `std::chrono::microseconds`

src/net.cpp
a-petrini: I personally prefer to keep it like that (but I am open to discuss!) as I generally resort to vectors of threads only when I need to launch multiple copies of the same thread, i.e. for splitting a specific task over a pool of threads, each one working on a subset of data.
I will post a commit which implements the approach you suggest, so we can discuss if it's better to keep a vector of references to thread objects, anyway.

titusen (approved):
The "CThreadInterrupt" class could also throw an exception just like boost::interrupt does. It works better when interrupt is nested in function calls. Still if a thread was interrupted it can be achieved by if (interrupted) throw error but requires more syntax. So for our needs it's enough.

LGTM, thanks.

ptagl (approved):

JackPiri (approved):
The PR is fine to.
Thanks

PR #616, opened 2023-10-26
Update libzendoo to 0.5.0

drgora:
This PR updates libzendoo dependency to version 0.5.0.

This new version of libzendoo introduces a new logging system, which takes options from a specific configuration file. A default configuration file for the new system is added to the repository, with the name `mc_crypto_log_config.yaml`.
On the first run, zend copies the default config file to the configured datadir, and modifies it so that the new logs are stored in the same datadir (where also `debug.log` is located).

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -198,6 +198,11 @@ boost::filesystem::path GetDebugLogPath()
     return GetDataDir() / "debug.log";
 }
 
+boost::filesystem::path GetMCCryptoLogPath()
+{
+    return GetDataDir(false) / "mc_crypto.log";
+}
+
 void OpenDebugLog()
 {
     boost::call_once(&DebugPrintInit, debugPrintInitFlag);
@@ -627,7 +632,7 @@ boost::filesystem::path GetConfigFile()
 
 boost::filesystem::path GetMcCryptoConfigFile()
 {
-    boost::filesystem::path pathConfigFile(GetArg("-mc_crypto_conf", "sample_log_config.yaml"));
+    boost::filesystem::path pathConfigFile(GetArg("-mc_crypto_conf", "mc_crypto_log_config.yaml"));
     if (!pathConfigFile.is_absolute())
         pathConfigFile = GetDataDir(false) / pathConfigFile;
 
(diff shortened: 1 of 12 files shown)


src/util.cpp
```diff
@@ -625,6 +630,15 @@ boost::filesystem::path GetConfigFile()
     return pathConfigFile;
 }
 
+boost::filesystem::path GetMcCryptoConfigFile()
+{
+    boost::filesystem::path pathConfigFile(GetArg("-mc_crypto_conf", "mc_crypto_log_config.yaml"));
```
JackPiri, line 635: Can be worth adding description of new param `"-mc_crypto_conf"` (and maybe also `"-conf"`) in docs at function `HelpMessage` of file `init.cpp`?

src/util.cpp
```diff
@@ -198,6 +198,11 @@ boost::filesystem::path GetDebugLogPath()
     return GetDataDir() / "debug.log";
 }
 
+boost::filesystem::path GetMCCryptoLogPath()
+{
+    return GetDataDir(false) / "mc_crypto.log";
```
JackPiri, line 203: Why is the parameter `fNetSpecific` set to `false`? I see an inconstistency with `GetDebugLogPath` (everything would be written as mainnet).

a-petrini (approved):
thanks for this PR, looks good to me!

src/util.cpp
drgora: I was also in doubt on this, but in the end I preferred this solution for simplicity. Having network specific paths would involve having 3 different configuration files, which must be copied (and possibly configured) on each first boot on given networks.
Given the amount of data logged (and its consistent format), I thought that it was not worth splitting the logs in network specific log files, and preferred logging in the main datadir only.

JackPiri: Ok, your explanation makes sense.
Maybe we could put a very brief comment just to make clear this was a decision we are aware of and not an oversight. But it's just up to you, I'm fine with your reply.

JackPiri (approved):
The PR is fine to me, thanks.

src/util.cpp
drgora: Yeah, makes sense. Done. Notice that `-conf` was already documented there.

ptagl (approved):

PR #618, opened 2023-10-30
Fixed a sporadic error in sc_cert_nonceasing.py

ptagl:
The error was related to a wrong `minconf` parameter that was sometimes causing a transaction to be created with an input that could eventually "disappear" due to the disconnection of some blocks.


drgora (approved):
Great catch, thanks!

JackPiri (approved):
The PR is fine to me, thanks

JackPiri (approved):

PR #619, opened 2023-11-08
Fix startup warning

drgora:
This PR fixes the startup warning message that is shown when default configuration files are missing. The message now shows the actual path that is being used, and not the default hardcoded one.

Also, if copying default config files fail, the execution fails with a proper error message, instead of creating empty config file in the destination directory.


JackPiri (approved):
Tried running this version, I confirm both issues have been solved.
Thanks

a-petrini (approved):
Looks good to me, thanks for this fix

ptagl (approved):

PR #620, opened 2023-11-10
Fix mc_crypto log config file path for Win

a-petrini:
This PR fixes a bug affecting Windows builds, caused by type mismatch when passing mc-crypto logger configuration file path to the dependency.


drgora (approved):
Looks good to me, thanks

ptagl (approved):

JackPiri (approved):
The PR is fine to me, thanks.

PR #621, opened 2023-11-14
OS -independent mc-crypto log configuration file

a-petrini:
This PR simplify the management of the configuration file for the newly introduced `mc-crypto` logger, adapting the configuration depending by the underlying OS `zend` is running on


src/util.cpp
```diff
@@ -630,13 +631,66 @@ boost::filesystem::path GetConfigFile()
...
-        pathConfigFile = GetDataDir(false) / pathConfigFile;
+    if (GetBoolArg("-enable_mc_crypto_logger", false)) 
+        return GetDataDir(false) / "mc_crypto_log_config.yaml";
+    else
+        return boost::filesystem::path("");
+}
 
-    return pathConfigFile;
+void createMcCryptoLogConfigFile() {
+    LogPrintf("mc-crypto logger configuration file not found! Creating a new on in current dataDir\n");
```
JackPiri: Just a small typo, this should be `a new **one** in %s`.

src/util.cpp
```diff
@@ -18,6 +18,7 @@
 
 #include <stdarg.h>
 #include <stdio.h>
+#include <regex>
```
JackPiri, line 21: Proposing again this comment (https://github.com/HorizenOfficial/zen/pull/616#discussion_r1374339863), since now the handling of the config file seems more easier and straightforward.
Please if considering this comment modify occurrences of `GetDataDir(false)` at lines 204, 640, 672.

cronicc: Regarding the originally linked comment, there are users (me for example) that run a testnet and mainnet node from the same data dir at the same time. The expectation is that mainnet logs to ~/.zen/*.log and testnet to ~/.zen/testnet3/*.log. Using the same config file for testnet and mainnet would be acceptable, but the log files should be in their respective directories to rule out multiple processes trying to write to the same file.

src/util.cpp
```diff
@@ -630,13 +631,66 @@ boost::filesystem::path GetConfigFile()
...
+    path: FILENAME_PLACEHOLDER\n\
+    encoder:\n\
+      pattern: \"{d(%Y-%m-%d %H:%M:%S)} | {({l}):5.5} | {f}:{L} - {m}{n}\"\n\
+    policy:\n\
+      trigger:\n\
+        kind: size\n\
+        limit: 10 kb\n\
+      roller:\n\
+        kind: fixed_window\n\
+        pattern: FILENAME_PLACEHOLDER{}.gz\n\
```
JackPiri, line 657: I'd suggest to add one more folder level as default; even if probably it's a matter of taste, I'd prefer default not spreading those files directly in datadir.

```suggestion
        pattern: FILENAME_ARCHIVE_PLACEHOLDER{}.gz\n\
        count: 5\n\
        base: 1\n\
\n\
root:\n\
  level: info\n\
  appenders:\n\
    - file\n";

    // Replace FILENAME_PLACEHOLDER/FILENAME_ARCHIVE_PLACEHOLDER with the proper paths (depending on the OS!)
    boost::filesystem::path logFilename = GetDataDir(false) / "mc_crypto.log";
    boost::filesystem::path logFilenameArchive = GetDataDir(false) / "mc_crypto_logs" / "mc_crypto.log";

    std::string updatedContent = std::regex_replace(logConfigContents, std::regex("FILENAME_PLACEHOLDER"), logFilename.c_str());
    updatedContent = std::regex_replace(updatedContent, std::regex("FILENAME_ARCHIVE_PLACEHOLDER"), logFilenameArchive.c_str());
```

JackPiri (approved):
Left few minor comments, probably only the one related to networks directories really worth being addressed. Apart from that the PR is fine to me.
Thanks

src/util.cpp
a-petrini: Thanks for pointing that out, it makes sense to save logs in different directories.

src/util.cpp
a-petrini: Thanks! Corrected

src/util.cpp
a-petrini: Thanks for the comment. I would leave the directories structure as it is, as with the default configuration just an handful of files gets created, and leave the user the task of customizing these aspect by altering the configuration files.

drgora (approved):

ptagl (approved):

PR #624, opened 2023-11-27
Gp/unshielding to script fork

JackPiri:
This PR introduces a new hard fork related to unshielding transactions; after **testnet** block 1396500 **(to be later defined for mainnet)** every unshielding transaction (moving funds from the private pool back to the public pool) must provide transparent output with locking script of type P2SH.
A minor reorganization and refactoring test suite related to fork-manager and shielded-pool-deprecation fork is included in this PR.

### DIFF
--- a/qa/rpc-tests/unshieldingtoscriptonly.py
+++ b/qa/rpc-tests/unshieldingtoscriptonly.py
@@ -14,7 +14,7 @@ from test_framework.util import assert_equal, assert_greater_than, \
 import sys
 
 RPC_VERIFY_REJECTED = -26
-RPC_HARD_FORK_DEPRECATION = -40 # TODO: update
+RPC_HARD_FORK_DEPRECATION = -40
 Z_FEE = 0.0001
 
 class ShieldedPoolDeprecationTest (BitcoinTestFramework):
--- a/qa/rpc-tests/unshieldingtoscriptonly.py
+++ b/qa/rpc-tests/unshieldingtoscriptonly.py
@@ -1,0 +1,207 @@
+#!/usr/bin/env python3
+# Copyright (c) 2016 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+
+from test_framework.test_framework import BitcoinTestFramework, ForkHeights
+from test_framework.authproxy import JSONRPCException
+from test_framework.util import assert_equal, assert_greater_than, \
+    assert_greater_or_equal_than, initialize_chain_clean, start_nodes, \
+    connect_nodes, wait_and_assert_operationid_status, \
+    send_shielding_raw, send_unshielding_raw
+
+import sys
+
+RPC_VERIFY_REJECTED = -26
+RPC_HARD_FORK_DEPRECATION = -40 # TODO: update
+Z_FEE = 0.0001
+
+class ShieldedPoolDeprecationTest (BitcoinTestFramework):
+
+    def setup_chain(self):
+        print("Initializing test directory "+self.options.tmpdir)
+        initialize_chain_clean(self.options.tmpdir, 2)
+
+    def setup_network(self, split=False):
+        self.nodes = start_nodes(2, self.options.tmpdir, [['-debug=zrpcunsafe', '-experimentalfeatures', '-zmergetoaddress']] * 2)
+        connect_nodes(self.nodes, 0, 1)
+        self.is_network_split=False
+        self.sync_all()
+
+    def make_all_mature(self):
+        self.sync_all()
+        self.nodes[0].generate(1)
+        self.sync_all()
+
+    def run_test (self):
+
+        ForkHeight_SPD = ForkHeights['SHIELDED_POOL_DEPRECATION']
+        ForkHeight = ForkHeights['UNSHIELDING_TO_SCRIPT_ONLY']
+
+        print("Mining blocks...")
+
+        self.nodes[0].generate(ForkHeight_SPD - 10)
+        self.sync_all()
+
+        # addresses
+        script_address = "zr3EMnGsEtpqTVbBdo4RztB3RQfNMUXt63j" # (0x 2092 0123456789012345678901234567890123456789 94a14d8a)
+        node0_zaddr0 = self.nodes[0].z_getnewaddress()
+
+        # generating some shielded notes
+        notes_to_create = 15
+        for repeat in range(notes_to_create):
+            opid = self.nodes[0].z_shieldcoinbase("*", node0_zaddr0, Z_FEE, 2)["opid"]
+            wait_and_assert_operationid_status(self.nodes[0], opid)
+            self.sync_all()
+        # generating some shielded notes through raw interface
+        notes_to_create_raw = 3
+        minimum_amount = 1.0
+        receive_result_and_key = [None] * notes_to_create_raw
+        for i in range(notes_to_create_raw):
+            zckeypair = self.nodes[0].zcrawkeygen()
+            zcsecretkey = zckeypair["zcsecretkey"]
+            zcaddress = zckeypair["zcaddress"]
+            tx_hash, encrypted_note = send_shielding_raw(self.nodes[0], minimum_amount, zcaddress)
+            receive_result_and_key[i] = [self.nodes[0].zcrawreceive(zcsecretkey, encrypted_note), zcsecretkey]
+            self.sync_all()
+        receive_result_and_key_index = 0
+
+        self.nodes[0].generate(1)
+        self.sync_all()
+        
+        # first round pre-fork, second round post-fork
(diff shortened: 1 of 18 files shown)


src/zen/forks/fork12_unshieldingtoscriptonlyfork.h
```diff
@@ -0,0 +1,20 @@
...
+
+class UnshieldingToScriptOnlyFork : public ShieldedPoolDeprecationFork
+{
+public:
+    UnshieldingToScriptOnlyFork();
+
+    /**
+     * @brief returns true if the unshielding (z->t) transactions must be performed towards script addresses
+     */
+    inline virtual bool mustUnshieldToScript() const { return true; };
```
drgora, line 16: Just a minor thing here, feel free to ignore: methods defined in class definitions are implicitly `inline` already, I'd suggest to remove the explicit keyword

src/wallet/rpcwallet.cpp
```diff
@@ -5257,6 +5296,9 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
             "z_sendmany \"fromaddress\" [{\"address\":... ,\"amount\":...},...] ( minconf ) ( fee ) (sendChangeToSource)\n"
             + ShieldingRPCMethodsDeprecationWarning(false) + "\n"
             + "Details: sending transparent funds to shielded addresses " + (AreShieldingRPCMethodsDeprecated() ? "has been " : "is going to be ") + "deprecated.\n"
+            + UnsieldingRPCMethodsDeprecationWarning() + "\n"
+            + "Details: sending shielded funds to script addresses " + (AreShieldingRPCMethodsDeprecated() ? "has been " : "is going to be ") + "deprecated.\n"
```
drgora:
```suggestion
            + "Details: sending shielded funds to non-script addresses " + (AreUnshieldingRPCMethodsDeprecated() ? "has been " : "is going to be ") + "deprecated.\n"
```

src/wallet/rpcwallet.cpp
```diff
@@ -5385,6 +5427,18 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
         throw JSONRPCError(RPC_HARD_FORK_DEPRECATION, shieldedPoolDeprecationErrorMessage);
     }
 
+    // we want to forbid unshielding to non-script addresses
+    if (chainActive.Height() + 1 >= GetUnshieldingToScriptOnlyForkHeight() &&
+        !fromTaddr && taddrRecipients.size() > 0)
+    {
+        for (const auto taddrRecipient: taddrRecipients)
```
drgora: I believe we can avoid a copy here:
```suggestion
        for (const auto& taddrRecipient: taddrRecipients)
```

src/wallet/rpcwallet.cpp
```diff
@@ -6061,8 +6115,10 @@ UniValue z_mergetoaddress(const UniValue& params, bool fHelp)
             + strDisabledMsg
             + ShieldingRPCMethodsDeprecationWarning(false) + "\n"
             + "Details: merging transparent funds to shielded address " + (AreShieldingRPCMethodsDeprecated() ? "has been " : "is going to be ") + "deprecated.\n"
-            "\nMerge multiple UTXOs and notes into a single UTXO or note." +
-            (AreShieldingRPCMethodsDeprecated() ? "" : "\nCoinbase UTXOs are ignored; use `z_shieldcoinbase` to combine those into a single note.") +
+            + UnsieldingRPCMethodsDeprecationWarning() + "\n"
+            + "Details: merging shielded funds to script address " + (AreShieldingRPCMethodsDeprecated() ? "has been " : "is going to be ") + "deprecated.\n"
```
drgora:
```suggestion
            + "Details: merging shielded funds to non-script address " + (AreUnshieldingRPCMethodsDeprecated() ? "has been " : "is going to be ") + "deprecated.\n"
```

a-petrini (approved):
Nothing to point out, looks good to me.
Thanks for this PR!

qa/rpc-tests/unshieldingtoscriptonly.py
```diff
@@ -0,0 +1,207 @@
...
+from test_framework.authproxy import JSONRPCException
+from test_framework.util import assert_equal, assert_greater_than, \
+    assert_greater_or_equal_than, initialize_chain_clean, start_nodes, \
+    connect_nodes, wait_and_assert_operationid_status, \
+    send_shielding_raw, send_unshielding_raw
+
+import sys
+
+RPC_VERIFY_REJECTED = -26
+RPC_HARD_FORK_DEPRECATION = -40 # TODO: update
```
drgora: Does this really need updating? It looks ok to me already

drgora (changes requested):
Thank you for the PR. It looks good to me, except a couple small things regarding RPC messages that I believe should be fixed.
I also left few other non-blocking comments, feel free to apply or reject any as you see fit.

src/wallet/rpcwallet.cpp
JackPiri: Great, three letters changing completely the meaning! :sweat_smile:
Modified accordingly.

src/wallet/rpcwallet.cpp
JackPiri: Agree with you.
Modified accordingly.

src/wallet/rpcwallet.cpp
JackPiri: Same as above.
Modified accordingly.

qa/rpc-tests/unshieldingtoscriptonly.py
JackPiri: Correct, it's a left over belonging to development phase.
Modified accordingly.

src/zen/forks/fork12_unshieldingtoscriptonlyfork.h
JackPiri: This was done replicating what is already present in other `forkXX_....h` files, let's keep this coherency for now.

drgora (approved):

src/wallet/rpcwallet.cpp
```diff
@@ -88,12 +89,50 @@ namespace{
...
+    bool AreUnshieldingRPCMethodsDeprecated()
+    {
+        return chainActive.Height() + 1 >= GetUnshieldingToScriptOnlyForkHeight();
+    }
+
+    //! Method for providing warning message associated to unshielding RPC methods deprecation
+    /*!
+      \return warning message
+    */
+    std::string UnsieldingRPCMethodsDeprecationWarning()
```
ptagl:
```suggestion
    std::string UnshieldingRPCMethodsDeprecationWarning()
```

src/wallet/rpcwallet.cpp
```diff
@@ -88,12 +89,50 @@ namespace{
...
+    }
+
+    //! Method for identifying if unshielding RPC methods are deprecated
+    /*!
+      \return shielding RPC methods deprecated
+    */
+    bool AreUnshieldingRPCMethodsDeprecated()
+    {
+        return chainActive.Height() + 1 >= GetUnshieldingToScriptOnlyForkHeight();
+    }
```
ptagl, line 120: Can we reuse this function in the code below when we need to perform the same check?

For instance inside:
- `std::string UnsieldingRPCMethodsDeprecationWarning()`
- `UniValue z_sendmany()`
- `UniValue z_mergetoaddress()`

src/primitives/transaction.cpp
```diff
@@ -1117,6 +1117,21 @@ bool CTransaction::ContextualCheck(CValidationState& state, int nHeight, int dos
                     }
                 }
             }
+            if (ForkManager::getInstance().mustUnshieldToScript(nHeight))
+            {
+                if (GetJoinSplitValueIn() > 0)
+                {
+                    for (auto const& out: vout)
+                    {
+                        if (out.scriptPubKey.GetType() != CScript::ScriptType::P2SH)
```
ptagl: Just a proposal, feel free to ignore: what about `!out.scriptPubKey.IsPayToScriptHash()`?

ptagl (approved):
It looks good to me, thanks, I've just left a few minor comments that are not mandatory to merge the PR.

src/wallet/rpcwallet.cpp
JackPiri: Agree, addressed.

src/primitives/transaction.cpp
JackPiri: Makes sense, addressed.

src/wallet/rpcwallet.cpp
JackPiri: Addressed all instances except for `std::string UnshieldingRPCMethodsDeprecationWarning()` (in order to avoid calling twice `GetUnshieldingToScriptOnlyForkHeight`).

a-petrini (approved):

PR #628, opened 2024-01-09
Shielded pool removal

a-petrini:
This PR removes the shielded pool from zend.
In particular, removes the possibility of submitting transactions involving shielded notes, hence t->z, z->z and z->t transactions will no longer be accepted and processed.
This PR also modifies all the RPC commands directly related to these transactions - `z_sendmany` and `z_mergetoaddress` - and adds deprecation warnings to all the remaining `z_` commands.


src/rpc/protocol.cpp
```diff
@@ -5,6 +5,7 @@
 
 #include "rpc/protocol.h"
 
+#include "main.h"
```
JackPiri: The `include` added here can be removed.

src/wallet/rpcwallet.cpp
```diff
@@ -10,7 +10,9 @@
 #include "main.h"
 #include "net.h"
 #include "netbase.h"
+#include "rpc/protocol.h"
```
JackPiri: Also the following `include`s can be removed:

- `#include "rpc/protocol.h"`
- `#include "zen/forks/fork11_shieldedpooldeprecationfork.h"`
- `#include "zen/forks/fork12_shieldedpoolremovalfork.h"`

src/zen/forks/fork0_originalfork.h
```diff
@@ -108,6 +108,11 @@ class OriginalFork : public Fork
      * @brief returns true if the shielding (t->z) transactions are forbidden
      */
     inline virtual bool isShieldingForbidden() const { return false; };
+
+    /**
+     * @brief returns true if the shielded pool has been removed (no more t->z, z->z, z->t)
+     */
+    virtual bool isShieldedPoolRemoved() const { return false; };
```
JackPiri: Probably this one here should be: `inline virtual bool isShieldedPoolRemoved() const { return false; };`.

JackPiri (approved):
The PR is fine to me.
Thanks

src/rpc/protocol.cpp
a-petrini: Agree, done!

src/wallet/rpcwallet.cpp
a-petrini: Removed

src/zen/forks/fork0_originalfork.h
a-petrini: Updated the signature, thanks!

src/wallet/rpcwallet.cpp
```diff
@@ -4106,6 +4062,8 @@ UniValue z_listunspent(const UniValue& params, bool fHelp)
     if (fHelp || params.size() > 4)
         throw runtime_error(
             "z_listunspent ( minconf maxconf includeWatchonly [\"zaddr\",...] )\n"
+            + ShieldedPoolRPCMethodsDeprecationWarning() + "\n"
+            + "Details: shielded pool transactions (t->z, z->z, z->t) " + (AreShieldedPoolRPCMethodsDeprecated() ? "have been " : "are going to be ") + "deprecated.\n"
```
ptagl: I wonder if we should use "disabled" instead of "deprecated", as "deprecation" is typically used with a slightly different meaning and may confuse users.

drgora (approved):
Looks good to me, thanks

PR #640, opened 2024-03-22
Adding test reproducing mempool conflicts

a-petrini:
This PR introduces a Python test reproducing a scenario where conflicting certificates and transactions should be correctly evicted from the mempool.

### DIFF
--- a/qa/rpc-tests/sc_mempool_conflict.py
+++ b/qa/rpc-tests/sc_mempool_conflict.py
@@ -7,12 +7,12 @@ from test_framework.test_framework import BitcoinTestFramework
 from test_framework.test_framework import ForkHeights
 from test_framework.authproxy import JSONRPCException
 from test_framework.util import initialize_chain_clean, get_epoch_data, swap_bytes, \
-    start_nodes, connect_nodes_bi, mark_logs
+    start_nodes, mark_logs
 from test_framework.mc_test.mc_test import *
 from decimal import Decimal
 
 DEBUG_MODE    = 1
-NUMB_OF_NODES = 2
+NUMB_OF_NODES = 1
 EPOCH_LENGTH  = 0
 
 FT_SC_FEE      = Decimal('0')
@@ -29,7 +29,8 @@ PARAMS_NAME = "sc"
 #   this transaction in a block
 # - Invalidate the block containing the certificate; the certificate gets pushed back to the
 #   mempool, creating a conflict with the transaction
-# - Under normal circumstances, the transaction should be purged from the mempool
+# - Under normal circumstances, the transaction should be purged from the mempool (see 
+# `CTxMemPool::checkTxImmatureExpenditures()` inside `txmempool.cpp`)
 
 class non_ceasing_sc_mempool_conflict(BitcoinTestFramework):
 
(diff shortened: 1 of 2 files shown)


qa/rpc-tests/sc_mempool_conflict.py
```diff
@@ -0,0 +1,190 @@
...
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.test_framework import ForkHeights
+from test_framework.authproxy import JSONRPCException
+from test_framework.util import initialize_chain_clean, get_epoch_data, swap_bytes, \
+    start_nodes, connect_nodes_bi, mark_logs
+from test_framework.mc_test.mc_test import *
+from decimal import Decimal
+
+DEBUG_MODE    = 1
+NUMB_OF_NODES = 2
```
drgora: Could we just use 1 node for this test? It seems only one is ever used.

qa/rpc-tests/sc_mempool_conflict.py
```diff
@@ -0,0 +1,190 @@
...
+        # Step three:
+        # - send certificate for epoch 1
+        # - mine one block
+        amounts = {"address": btw_address, "amount": bwt_amount_1}
+
+        epoch_number = 1
+        ref_quality = 2
+        ref_height = self.nodes[0].getblockcount() - 2
+        cert_1 = self.try_send_certificate(0, scid, epoch_number, ref_quality, ref_height, MBTR_SC_FEE, FT_SC_FEE, amounts, False)
+        self.sync_all()
```
drgora, line 130: I believe this shouldn't be necessary (even if we had more than 1 node), maybe we can remove it?

qa/rpc-tests/sc_mempool_conflict.py
```diff
@@ -0,0 +1,190 @@
...
+                assert(failure_reason in errorString)
+            return None
+
+    def run_test(self):
+
+        # General sc / cert variables
+        creation_amount = Decimal("50")
+        bwt_amount_1    = Decimal("3")
+        address         = "dada"
+        btw_address     = self.nodes[0].getnewaddress()
```
drgora: Definitely minuscule thing here: there are few `btw`s in the test, where I believe it should be `bwt`s.

drgora (approved):
I left few minor comments, but this definitely looks good to me, thanks!

qa/rpc-tests/sc_mempool_conflict.py
```diff
@@ -0,0 +1,190 @@
...
+# This script reproduces an issue introduced with non-ceasable sidechains, where conflicting
+# certificates and transactions may be submitted in the mempool, creating undefined behaviors.
+# Briefly, the test executes the following steps:
+# - create a non-ceasable sidechain
+# - submit a certificate and include that in a block
+# - submit a transaction explicitly spending one of the certificate output; do not include
+#   this transaction in a block
+# - Invalidate the block containing the certificate; the certificate gets pushed back to the
+#   mempool, creating a conflict with the transaction
+# - Under normal circumstances, the transaction should be purged from the mempool
```
ptagl: Do you think it could add value here to include a reference to the check that is performed by the mempool on Zend? 
```suggestion
# - Under normal circumstances, the transaction should be purged from the mempool (see `CTxMemPool::checkTxImmatureExpenditures()` inside `txmempool.cpp`)
```

qa/rpc-tests/sc_mempool_conflict.py
a-petrini: Yes, the second node is useless and I removed that altogether.

qa/rpc-tests/sc_mempool_conflict.py
a-petrini: I agree. I removed this part and adjusted the block heights / epochs for the next certificate sending

qa/rpc-tests/sc_mempool_conflict.py
a-petrini: Fixed, thanks!

qa/rpc-tests/sc_mempool_conflict.py
a-petrini: Thanks, I added the reference in the comment, as suggested

ptagl (approved):

PR #644, opened 2024-05-20
Update copyright headers

ptagl:
This pull request introduces the following changes:
- a new Python script `update-copyright-headers.py` to update copyright headers in the source files of the repository
- updated headers for "Zen Blockchain Foundation" for all the files modified between `2018-01-18` and `2023-09-27`
- updated headers for "The Horizen Foundation" for all the files modified between `2023-09-28` and `2024-05-19`

The previous script `fix-copyright-headers.py` has been removed and the `devtools` readme has been updated accordingly.


contrib/devtools/README.md
```diff
@@ -41,15 +41,19 @@ Configuring the github-merge tool for the bitcoin repository is done in the foll
 fix-copyright-headers.py
 ===========================
 
-Every year newly updated files need to have its copyright headers updated to reflect the current year.
-If you run this script from src/ it will automatically update the year on the copyright header for all
-.cpp and .h files if these have a git commit from the current year.
+This scripts takes care of updating the copyright headers in the source files of the repository.
```
ptagl:
```suggestion
This script takes care of updating the copyright headers in the source files of the repository.
```

contrib/devtools/README.md
```diff
@@ -41,15 +41,19 @@ Configuring the github-merge tool for the bitcoin repository is done in the foll
 fix-copyright-headers.py
```
ptagl:
```suggestion
update-copyright-headers.py
```

drgora (approved):
Looks good to me, thanks!

contrib/devtools/update-copyright-headers.py
```diff
@@ -0,0 +1,188 @@
...
+        print("Error: until_date must be higher than since_date.")
+        exit(1)
+
+def list_files(path, exclude_folders, extensions):
+    """
+    Recursively traverse a directory and return a list of files with specified extensions.
+    """
+    file_list = []
+    for root, dirs, files in os.walk(path):
+        # Exclude specified subfolders
```
JackPiri: From this code I understand the `args.exclude_dir` set is targeting only the first level of `args.directory` (not deeper subdirectories):

```
    # Check if the specified excluded subfolders exist
    for folder in args.exclude_dir:
        path = os.path.join(args.directory, folder)
        if not os.path.exists(path):
            print(f"Error: Folder '{path}' does not exist in the specified directory.")
            exit(1)
```

if my understanding is correct, probably we want to modify a little bit `list_files` function:

```
    file_list = []
    first_level = True
    for root, dirs, files in os.walk(path):
        if first_level:
            # Exclude specified subfolders
            dirs[:] = [d for d in dirs if d not in exclude_folders]
            first_level = False
        for file in files:
```

in the situation we have deeper subdirectories named the same way (not the case as of now).

JackPiri (approved):
The PR is fine to me, I've left just a minor comment for making the script a bit more coherent.
Thanks

contrib/devtools/update-copyright-headers.py
ptagl: Thanks for highlighting this, I improved the script by making the handling of paths more robust, combining relative and absolute paths.

This should now guarantee even more flexibility without the risk of multiple matches for the same folder.
In particular, supposing to have a folder tree like:
```
src/
├─ A/
├─ B/
├─ C/
│  ├─ A/
```
excluding the folder `A` wouldn't also cause `C/A` to be excluded, as it should be provided as a relative path to `src`.

ptagl:
@drgora I've added a basic mechanism to exclude commits that we shouldn't take into account when updating copyright headers (like the updates of the copyright headers themselves)! 😄 

The mechanism is based on a new `--exclude-commit` argument (list) with the following behavior:
- if the argument is **empty** (not provided by the user), the script tries to automatically exclude commits containing `update-copyright-headers.py` inside the message or `copyright` inside the summary (I didn't want to check the whole message here just to prevent accidental exclusion of relevant commits); after the automatic detection, user is asked to confirm that the list of commits is fine, with the possibility to continue, abort or ignore the list
- if the list is **provided** by the user, no automatic detection is performed

Let me know what you think about it.

Thanks!

drgora:
> @drgora I've added a basic mechanism to exclude commits that we shouldn't take into account when updating copyright headers (like the updates of the copyright headers themselves)! 😄
> 
> The mechanism is based on a new `--exclude-commit` argument (list) with the following behavior:
> 
>     * if the argument is **empty** (not provided by the user), the script tries to automatically exclude commits containing `update-copyright-headers.py` inside the message or `copyright` inside the summary (I didn't want to check the whole message here just to prevent accidental exclusion of relevant commits); after the automatic detection, user is asked to confirm that the list of commits is fine, with the possibility to continue, abort or ignore the list
> 
>     * if the list is **provided** by the user, no automatic detection is performed
> 
> 
> Let me know what you think about it.
> 
> Thanks!

Sounds great to me, thanks!

contrib/devtools/update-copyright-headers.py
JackPiri: It's fine to me now, thanks.

PR #658, opened 2025-04-17
HF for disabling SC creation and forward transfers

alsala:
To support the Horizen migration, a new hard fork has been implemented at a specified block height (to be determined) that disables the creation of new sidechains and the forwarding of funds to existing sidechains. Note that backward transfers and certificate processing will remain functional after the fork point.

See JIRA task at: https://horizenlabs.atlassian.net/browse/HZN-2258


qa/rpc-tests/sc_cr_and_fw_stop_fork.py
```diff
@@ -0,0 +1,260 @@
...
+        creating_tx = ret['txid']
+
+        mark_logs("Check creating tx={} is in node 0 mempool...".format(creating_tx), self.nodes, DEBUG_MODE)
+        assert_true(creating_tx in self.nodes[0].getrawmempool())
+
+        mark_logs("Node 0 generates 1 block", self.nodes, DEBUG_MODE)
+        bl = self.nodes[0].generate(1)[0]
+        pprint.pprint(self.nodes[0].getblock(bl))
+
+        mark_logs("Check we do have the new SC in netowrk part 0", self.nodes, DEBUG_MODE)
```
drgora: Small typo on network

qa/rpc-tests/sc_cr_and_fw_stop_fork.py
```diff
@@ -0,0 +1,260 @@
...
+                  DEBUG_MODE)
+        self.nodes[1].generate(2)
+
+        mark_logs("---------------------- Join the network -------------------", self.nodes, DEBUG_MODE)
+        self.join_network(0)
+        self.sync_all()
+
+        sc_info_post = self.nodes[0].getscinfo(ret['scid'])
+
+        mark_logs("Check we do not have the new SC in the joint netowrk", self.nodes, DEBUG_MODE)
```
drgora: Small typo on network

src/primitives/transaction.cpp
```diff
@@ -1084,6 +1084,14 @@ bool CTransaction::ContextualCheck(CValidationState& state, int nHeight, int dos
                                  CValidationState::Code::INVALID, "bad-txns-transparent-jsnotempty");
             }
 
+            // check that after this fork point we do not have any new sidechain creation or forward transfer to existing ones.
+            if (ForkManager::getInstance().isScCreationAndFwdtStopped(nHeight)) {
+                if (!vsc_ccout.empty() || !vft_ccout.empty()) {
+                    return state.DoS(dosLevel, error("ContextualCheck(): sc tx with sc creation output or fwdt output, but thei are illegat at this height"),
```
drgora: A couple of typos in this error msg

qa/rpc-tests/sc_cr_and_fw_stop_fork.py
```diff
@@ -0,0 +1,260 @@
...
+        - Join the network
+           . The network 0 SC creation block is reverted, the relevant SC creation tx gets back to the mempool but is
+             evicted since the fork point has been reached
+           . The SC info is not retrievable anymore and no SC creation tx can be found in the mempool
+        - Try creating a new SC and try to fwt to existing SC: it is not possible anymore after the fork point
+        """
+
+        mark_logs("Node 1 generates 1 block", self.nodes, DEBUG_MODE)
+        self.nodes[1].generate(1)
+        self.sync_all()
```
drgora, line 60: Is this actually needed?

qa/rpc-tests/sc_cr_and_fw_stop_fork.py
```diff
@@ -0,0 +1,260 @@
...
+        bl = self.nodes[0].generate(1)[0]
+        pprint.pprint(self.nodes[0].getblock(bl))
+
+        mark_logs("Check we do have the new SC in netowrk part 0", self.nodes, DEBUG_MODE)
+        sc_info_pre = self.nodes[0].getscinfo(ret['scid'])
+        pprint.pprint(sc_info_pre)
+        assert_true(1 == len(sc_info_pre['items']))
+
+        # ----------------------Network Part 1
+        mark_logs("Node 1 generates 2 nodes, crossing the fork point and reaching the longest chain", self.nodes,
```
drgora:
```suggestion
        mark_logs("Node 1 generates 2 blocks, crossing the fork point and reaching the longest chain", self.nodes,
```

qa/rpc-tests/sc_cr_and_fw_stop_fork.py
```diff
@@ -0,0 +1,260 @@
...
+from test_framework.authproxy import JSONRPCException
+from test_framework.mc_test.mc_test import *
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.test_framework import ForkHeights
+from test_framework.util import initialize_chain_clean, \
+    sync_blocks, sync_mempools, connect_nodes_bi, mark_logs, assert_true, assert_false, start_node, \
+    stop_nodes, wait_bitcoinds
+
+DEBUG_MODE = 1
+NUMB_OF_NODES = 3
```
drgora: It looks like we only ever use 2 nodes, so I'd suggest not to spawn the third

qa/rpc-tests/sc_cr_and_fw_stop_fork.py
```diff
@@ -0,0 +1,260 @@
...
+            error_string = e.error['message']
+            print(error_string)
+            expected_substring = "This method is disabled"
+            assert expected_substring in error_string, f"'{error_string}' does not contain '{expected_substring}'"
+        else:
+            raise RuntimeError("An exception was expected")
+
+        self.sync_all()
+
+        mark_logs("...stopping and restarting nodes", self.nodes, DEBUG_MODE)
```
drgora: Why do we restart nodes? It doesn't seem to be used afterwards

src/init.cpp
```diff
@@ -588,6 +588,7 @@ std::string HelpMessage(HelpMessageMode mode)
     }
 
     strUsage += HelpMessageOpt("-maxtipage=<n>", "regtest only - Set the maximum tip age used for the relay of chain blocks)");
+    strUsage += HelpMessageOpt("-skipscopforkcheck", strprintf("regtest only - do not check deprecation of sc_create / sc_send method based on fork point (default: %d))", false));
```
drgora: It looks like this parameter is not that useful after all. Even if we force the creation of transactions that are not allowed from the RPC, the mempool rejects them immediately.
I would suggest removing it altogether, unless I am missing something else?

drgora (commented):
Looks mostly good to me, but I'd say it would be great if we could remove (if possible) the `skipscopforkcheck` parameter so to simplify things a bit.

src/init.cpp
alsala: That is exactly the reason for the '`regtest`' flag: verifying via py script that an offending transaction is rejected from the mempool. We cannot rely on transactions being sent solely via the zend RPC command.

qa/rpc-tests/sc_cr_and_fw_stop_fork.py
alsala: This is for having some mature funds which node 0 will need afterwards for sending a tx

qa/rpc-tests/sc_cr_and_fw_stop_fork.py
alsala: fixed

qa/rpc-tests/sc_cr_and_fw_stop_fork.py
alsala: fixed

src/primitives/transaction.cpp
alsala: Fixed

qa/rpc-tests/sc_cr_and_fw_stop_fork.py
alsala: Done

qa/rpc-tests/sc_cr_and_fw_stop_fork.py
alsala: Ok, done

qa/rpc-tests/sc_cr_and_fw_stop_fork.py
alsala: Ok, code removed

src/init.cpp
alsala: Ok, we can remove the option, after all the consensus is already tested when a reorg is attempted, in that case the offending tx is removed from mempool.

alsala: Done. Added also the case of sending signed raw tx after the fork point.

drgora (approved):
Nice, lgtm! Thanks

src/init.cpp
drgora: I like it, thanks!

PR #661, opened 2025-05-02
Disabling CI/CD for MacOS

ysibirski:
Removing CI/CD for MacOS since it was discontinued by TravisCI since March 31st. 2025
https://docs.travis-ci.com/user/reference/osx/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -162,68 +162,68 @@ os: linux
       - DOCKER_IS_DEB=true
       - TEST_CMD="./qa/zen/test-fetch-params.sh"
 
-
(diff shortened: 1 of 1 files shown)


cronicc (approved):

drgora:
Interesting; does this mean that next releases won't have MacOS builds anymore?

cronicc:
> Interesting; does this mean that next releases won't have MacOS builds anymore?

Yes, we've decided that we can move forward without MacOS full node support, especially as full node mode in Sphere was only required for shielded transactions which have been removed.

PR #662, opened 2025-05-09
HF for zend migration and new optional flag for stop block processing at a set height

alsala:
Modified the HF in order to stop all TXs from flowing in the network but coinbase TXs.

Also, introduced an optional flag, configurable at zend node startup, that limits block processing beyond a specified height.

See JIRA tasks at:
https://horizenlabs.atlassian.net/browse/HZN-2258


src/primitives/transaction.cpp
```diff
@@ -1050,6 +1050,14 @@ bool CTransaction::ContextualCheck(CValidationState& state, int nHeight, int dos
...
         return false;
 
+    // after this fork point we do not have any transaction flowing but coinbase txes
+    if (!IsCoinBase() && ForkManager::getInstance().areTransactionsStopped(nHeight)) {
+        LogPrintf("%s():%d - rejecting (ver=%d) transaction %s at block height %d\n",
+            __func__, __LINE__, nVersion, GetHash().ToString(), nHeight);
+        return state.DoS(dosLevel, error("ContextualCheck(): tx are illegal at this height"),
+            CValidationState::Code::INVALID, "bad-txs-stopped");
+    }
+
```
drgora, line 1057: Minor thing here: I would suggest doing this check before the possibly expensive call to `CheckBlockAtHeight()`

drgora (approved):
LGTM, thanks!

src/primitives/transaction.cpp
alsala: Ok, done

la10736 (approved):
LGTM

src/primitives/transaction.cpp
```diff
@@ -1047,6 +1047,14 @@ bool CTransaction::IsVersionStandard(int nHeight) const {
 
 bool CTransaction::ContextualCheck(CValidationState& state, int nHeight, int dosLevel) const
 {
+    // after this fork point we do not have any transaction flowing but coinbase txes
+    if (!IsCoinBase() && ForkManager::getInstance().areTransactionsStopped(nHeight)) {
```
cronicc, line 1051: Is my understanding here correct that IsCoinBase covers both coinbase transactions and certificates, or only coinbase transactions?

If the intended behavior is to still allow certificates, I'd like to see this covered in the RPC test (certificates still work after the fork).

alsala: Actually it covers only coinbase txs, certificates are handled in another method. But it definitely makes sense testing that certificates are working after the forks. I will add that to the py test, thanks.

alsala: Test added

